PR #6, opened 2017-03-13
Added `resourceState` to the `ObjectMetadata` type

jwngr:
The `resourceState` property is in the payload, just not in the type information.

@katfang - Can you confirm that the type should be `string` (meaning it will always be there) and not `string?` (meaning it is optional)?

### DIFF
--- a/src/providers/storage.ts
+++ b/src/providers/storage.ts
@@ -79,6 +79,7 @@ export class ObjectBuilder {
 export interface ObjectMetadata {
   kind: string;
   id: string;
+  resourceState: string;
   selfLink?: string;
   name?: string;
   bucket: string;


katfang:
Huh, well, that's an interesting question.

Right now, we want `string`.

But the eventual goal is to have different event types, and drop `resourceState`. This'll happen when we move onto GCS Pubsub. How terrible would it be to change it to `string?` when we introduce the new event types? (Where down the line does this come? Well, GCS Pubsub is getting launched into beta at the end of March.)

jwngr:
> How terrible would it be to change it to `string?` when we introduce the new event types?

Since this is now a public API and people are using it, we can't just change its behavior until the next breaking change. We can introduce a new property which can be used instead and deprecate `resourceState`, but we will need to backfill `resourceState` until we cut a new major version, else may break people's existing Cloud Functions.

It sounds like the `string` typing is correct for this property. I've also already added this property to our public reference docs [here](https://firebase.google.com/docs/reference/functions/functions.storage.ObjectMetadata#resourceState).

Issue #8, opened 2017-03-16
How do i point to latest firebase-functions in github.com under package.json


jernkuan:
https://www.npmjs.com/package/firebase-functions is currently outdated. 
I am looking to use the latest github as resourceState has just been added.
```
{
  "name": "generate-thumbnails-functions",
  "description": "Generate thumbnails",
  "dependencies": {
    "@google-cloud/storage": "^0.4.0",
    "child-process-promise": "^2.2.0",
    "firebase-admin": "^4.1.3",
    "firebase-functions": "git+https://github.com/firebase/firebase-functions.git"
  }
}
```
On compiling with typescript, it's having an error
Cannot find module 'firebase-functions'
```
$ npm install firebase/firebase-functions --save
└── firebase-functions@0.5.2  (git://github.com/firebase/firebase-functions.git#e9efdd372e5240df68728149f1fa2625e0e20d4d)

$ ls node_modules/firebase-functions/
CONTRIBUTING.md LICENSE     README.md   changelog.txt   package.json
```
I am not seeing the lib directory in there

laurenzlong:
Hey @jernkuan , are you developing with TypeScript or JavaScript? If you are developing with JavaScript, then you can already use event.data.resourceState with v0.5.2. The field is already populated, we just didn't have export a typing for it yet.

Here's how to use the latest SDK directly from Github:
1.  Inside projectDirectory/functions, run: 
```
git clone https://github.com/firebase/firebase-functions.git
cd firebase-functions
npm run build
```
2. Modify the firebase-functions dependency inside functions/package.json to be:
`"firebase-functions": "file:./firebase-functions"`
3. Run `npm install` inside projectDirectory/functions

jernkuan:
Working with typescript and compiler throwing an error with the resourceState thus the need to use the latest from github. I think your suggestion would work. 

Thanxs and closing

laurenzlong:
Ok, no probs!

jwngr:
FYI that it's generally not the best practice to always point to the latest `master` instead of the latest release. It could cause unexpected issues if we happen to merge a breaking change into `master`. If you want to get the current release to work in TypeScript, you can just do `(object as any).resourceType` to get rid of the TypeScript warning.

PR #9, opened 2017-03-16
Add postinstall script to run build script

jernkuan:
Run script to build lib if lib directory is not found during postinstall

This solves https://github.com/firebase/firebase-functions/issues/8


jernkuan:
Solution does not work when deploying to firebase, since tsc is not available on container.

PR #13, opened 2017-03-19
Correct the argument types of the forEach function

Byeol:
According to the definition of `Array.prototype.forEach` function, `action` should return `void`.

### DIFF
--- a/spec/providers/database.spec.ts
+++ b/spec/providers/database.spec.ts
@@ -223,10 +223,10 @@ describe('DeltaSnapshot', () => {
       let count = 0;
       let counter = snap => count++;
 
-      subject.forEach(counter);
+      expect(subject.forEach(counter)).to.equal(false);
       populate(23, null);
 
-      subject.forEach(counter);
+      expect(subject.forEach(counter)).to.equal(false);
       expect(count).to.eq(0);
     });
 
@@ -243,13 +243,21 @@ describe('DeltaSnapshot', () => {
       expect(ret).to.equal(true);
     });
 
-    it('should not cancel further enumeration if callback does not return true', () => {
+    it('should not cancel further enumeration if callback returns a truthy value', () => {
       populate(null, { a: 'b', c: 'd', e: 'f', g: 'h' });
       let out = '';
       const ret = subject.forEach(snap => {
-        if (snap.val() === 'a') {
-          return true;
-        }
+        out += snap.val();
+        return 1;
+      });
+      expect(out).to.equal('bdfh');
+      expect(ret).to.equal(false);
+    });
+
+    it('should not cancel further enumeration if callback does not return', () => {
+      populate(null, { a: 'b', c: 'd', e: 'f', g: 'h' });
+      let out = '';
+      const ret = subject.forEach(snap => {
         out += snap.val();
       });
       expect(out).to.equal('bdfh');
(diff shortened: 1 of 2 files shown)


jwngr:
The type signature is correct, but the [implementation of `forEach()`](https://github.com/firebase/firebase-functions/blob/e9efdd372e5240df68728149f1fa2625e0e20d4d/src/providers/database.ts#L209-L216) actually returns the wrong thing in certain cases. As noted in the [reference documentation](https://firebase.google.com/docs/reference/functions/functions.database.DeltaSnapshot), `forEach()` should return a boolean which indicates if execution was cut short by `action()` returning `true` at some point. However, as `forEach()` is implemented, it always returns `true`. We need to update the method to check the return value of `action()`, and, if it is `true`, stop execution and return `true` ourselves.

I'm updating the title of this issue and assigning over to @laurenzlong for now.

cc/ @inlined - Since the relevant `TODO` at the top of this method's implementation has your name in it.

jwngr:
@Byeol - If you'd like to update this pull request given my explanation above, we would be happy to accept your contribution. We should probably include some tests [here](https://github.com/firebase/firebase-functions/blob/e9efdd372e5240df68728149f1fa2625e0e20d4d/spec/providers/database.spec.ts#L211) for different scenarios as well.

Byeol:
@jwngr You're right. I updated the implementation of `forEach()` to make it correspond to the reference documentation.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach), There is no way to stop or break a forEach() loop other than by throwing an exception. So I used [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method, which returns `true` if the callback function returns a truthy value for any array element; otherwise, false. `some()` immediately returns true if the callback function returns a truthy value for any array element.

src/providers/database.ts
```diff
@@ -206,11 +206,10 @@ export class DeltaSnapshot implements firebase.database.DataSnapshot {
     return valAt(this._delta, this._childPath) !== undefined;
   }
 
-  // TODO(inlined) what is this boolean for?
   forEach(action: (a: DeltaSnapshot) => boolean): boolean {
     let val = this.val();
     if (_.isPlainObject(val)) {
-      _.keys(val).forEach(key => action(this.child(key)));
+      return _.keys(val).some(key => action(this.child(key)));
```
jwngr: Lodash has a helper [`_.some()`](https://lodash.com/docs/4.17.4#some) which we can use directly, instead of having to do `_.keys(val).some()`.

jwngr: Also, I think we want to make sure this behavior only happens when `true` is returned, not when a truthy value (such as `"a"` or `1`) is returned. So, this should probably end with `action(this.child(key) === true`. Can you please add a test for the case where `action()` returns `1` to ensure this behavior is correct?

spec/providers/database.spec.ts
```diff
@@ -229,6 +229,32 @@ describe('DeltaSnapshot', () => {
       subject.forEach(counter);
       expect(count).to.eq(0);
     });
```
jwngr: Can you add some checks here that the return values of both `forEach()` calls in this test are `false`?

Byeol:
@jwngr Thanks for the review! I updated the implementation to cancel further enumeration only if the callback returns explicit `true`, and simplified the code with a Lodash helper `_.some()`.

jwngr:
Looks great to me! Thanks for iterating on this with me. I'm going to assign this over to @rjhuijsman and @laurenzlong so they can merge this in when they are ready for it.

PR #31, opened 2017-04-10
Change release function to only decrement counter after waiting 2 min

laurenzlong:
### Description
Fixing a bug where database connections were being closed even if a function was invoked recently. I changed the model for ref counting, to only decrement the counter after waiting 2 minutes. Whereas the old model decremented the counter right away, and then waited 2 minutes to destroy the app. With the new changes, an app will only be destroyed if there's 2 minutes of no invocations.

### Code sample

N/A


laurenzlong:
Great test cases! Just added them.

Issue #36, opened 2017-04-14
functions.database.DeltaSnapshot#key outputs incorrect value


cevanish12:
### Version info

**firebase-functions:** ^0.5

**firebase-admin:** ^4.1.4

### Test case

This is similar to what I have:

```
const functions = require('firebase-functions');

exports.myFunction = functions.database.ref('/Accounts/{uId}/{accountId}/')
    .onWrite(event => {
	 event.data.current.forEach(function(item) {
	     console.log(item.key);
	 });
    });
```

### Were you able to successfully deploy your functions?

Yup. I'd classify this as more of a semantic error.

### Expected behavior

If my path was /Accounts/1234/5678, the key should give me 5678.

### Actual behavior

The key property gives me 12345678 (combining the second and third level children).

### Temporary fix

I put this on top of my index.js file, and it seems to fix the problem, though it might not cover all cases.

```
const functions = require('firebase-functions');

functions.database.DeltaSnapshot.prototype._fullPath = function () {
    var out = (this._path || '') + '/' + (this._childPath || '');
    if (out === '') {
        out = '/';
    }
    return out;
}
```

laurenzlong:
Thanks for reporting! In the meanwhile, you can work around this by using event.params.accountId

davidkassa:
This also affects `item.ref`. Overriding the _fullPath prototype successfully works around this issue.

laurenzlong:
Fix has been merged into master.

PR #40, opened 2017-04-25
Fix bug that caused database key and child key to be concatenated

laurenzlong:
### Description

Fixing bug https://github.com/firebase/firebase-functions/issues/36

### Code sample
N/A


laurenzlong:
Added! Verified that the test fails with the old implementation and succeeds with the new one.

PR #46, opened 2017-05-02
Fix database event.data.val() bug

laurenzlong:
### Description

Fix bug where event.data.val() was returning array of nulls when database payload had a length property.

### Code sample

If new database node looks like:
{ length: "something", foo: "bar"}

Previously, calling event.data.val() would return array of nulls (since lodash's _.forEach method thought that the data was actually an array due to the presence of the length property, and tried to iterate over teh nonexistent array elements). Now, event.data.val() will return the expected value of { length: "something", foo: "bar"}

### DIFF
--- a/spec/providers/database.spec.ts
+++ b/spec/providers/database.spec.ts
@@ -137,6 +137,12 @@ describe('DeltaSnapshot', () => {
       populate({ myKey: 'foo', myOtherKey: 'bar' }, { myKey: null });
       expect(subject.val()).to.deep.equal({ myOtherKey: 'bar' });
     });
+
+    // Regression test: .val() was returning array of nulls when there's a property called length
+    it('should return correct values when data has "length" property', () => {
+      populate(null, { length: 3,  foo: 'bar' });
+      expect(subject.val()).to.deep.equal({ length: 3, foo: 'bar'});
+    });
   });
 
   describe('#child(): DeltaSnapshot', () => {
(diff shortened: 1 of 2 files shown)


spec/providers/database.spec.ts
```diff
@@ -137,6 +137,12 @@ describe('DeltaSnapshot', () => {
       populate({ myKey: 'foo', myOtherKey: 'bar' }, { myKey: null });
       expect(subject.val()).to.deep.equal({ myOtherKey: 'bar' });
     });
+
+    // Regression test: .val() was returning array of nulls when there's a property called length
```
rjhuijsman: Can you add "BUG(37683995)" in this comment, for our future reference if we want to remember more details?

src/providers/database.ts
```diff
@@ -248,16 +248,19 @@ export class DeltaSnapshot implements firebase.database.DataSnapshot {
...
-    _.forEach(node, (childNode, key) => {
-      obj[key] = this._checkAndConvertToArray(childNode);
-      numKeys++;
-      const integerRegExp = /^(0|[1-9]\d*)$/;
-      if (allIntegerKeys && integerRegExp.test(key)) {
-        maxKey = Math.max(maxKey, Number(key));
-      } else {
-        allIntegerKeys = false;
+    for (let key in node) {
+      if (node.hasOwnProperty(key)) {
```
rjhuijsman: You can remove a level of indentation for the code below, by using:
`if (!node.hasOwnProperty(key)) { continue; }`

laurenzlong:
@rjhuijsman done! Thanks for the tip!

PR #51, opened 2017-05-09
Use process.env.FIREBASE_PROJECT for Firebase config values if available

laurenzlong:
### Description

Use FIREBASE_PROJECT and CLOUD_RUNTIME_CONFIG during population of function.config(), in anticipating of these env variables being available in the runtime in the future.

### DIFF
--- a/spec/config.spec.ts
+++ b/spec/config.spec.ts
@@ -30,17 +30,12 @@ describe('config()', () => {
   afterEach(() => {
     mockRequire.stopAll();
     unsetSingleton();
+    delete process.env.FIREBASE_PROJECT;
+    delete process.env.CLOUD_RUNTIME_CONFIG;
   });
 
   it('loads config values from .runtimeconfig.json', () => {
     mockRequire('../../../.runtimeconfig.json', { foo: 'bar', firebase: {} });
-    let loaded = config();
-    expect(loaded).to.have.property('firebase');
-    expect(loaded).to.have.property('foo','bar');
-  });
-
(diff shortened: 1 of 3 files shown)


laurenzlong:
Must use with this branch of the CLI: https://github.com/firebase/firebase-tools/pull/328

src/config.ts
```diff
@@ -42,23 +42,30 @@ export namespace config {
...
+  }
+
   try {
-    loaded = require('../../../.runtimeconfig.json');
+    let path = process.env.CLOUD_RUNTIME_CONFIG || '../../../.runtimeconfig.json';
+    loadedFromFile = require(path);
   } catch (e) {
     try {
-      loaded = require('../../../config.json');
+      loadedFromFile = require('../../../config.json');
```
inlined: This caused confusion with @samtstern's functions and was meant to be alpha behavior. Should we cut this or give a deprecation notice?

spec/config.spec.ts
```diff
@@ -58,6 +58,6 @@ describe('config()', () => {
 
   it('throws an error if Firebase configs not present', () => {
     mockRequire('../../../.runtimeconfig.json', {});
-    expect(config).to.throw('Firebase config variables are missing.');
+    expect(config).to.throw('Firebase config variables are not available.');
```
inlined: We added more complex behaviors. Could we test them? Hint: I'm pretty sure I see a bug =p

src/config.ts
mbleigh: Cut it.

PR #57, opened 2017-05-26
Add support for create/update/delete event types for database functions.

laurenzlong:
### Description

Add support for create/update/delete event types for database functions.

### Code sample

```
 
exports.createEvent = functions.database.ref('/path').onCreate(event => {
        console.log('Data created.');
});
 
exports.updateEvent = functions.database.ref('/path').onUpdate(event => {
        console.log('Data updated.');
});
 
exports.deleteEvent = functions.database.ref('/path').onDelete(event => {
        console.log('Data at path is deleted or its only child is deleted.');
});
 
// onWrite currently exists and will be kept:
exports.writeEvent = functions.database.ref('/path').onWrite(event => {
        console.log('Data created, updated, or deleted');
});
```

### DIFF
--- a/spec/providers/database.spec.ts
+++ b/spec/providers/database.spec.ts
@@ -62,23 +62,6 @@ describe('DatabaseBuilder', () => {
         resource: 'projects/_/instances/subdomains/refs/users',
       } as any);
     });
-
-    it('should interpolate params until the server does it', () => {
-      let handler = database.ref('/users/{id}').onWrite(event => {
-        expect(event.resource).to.equal('projects/_/instances/subdomain/refs/users/aUserId');
-      });
-
-      return handler({
-        data: {
-          data: null,
-          delta: 'hello',
-        },
-        resource: 'projects/_/instances/subdomain/refs/users/{id}',
-        params: {
-          id: 'aUserId',
-        },
-      });
-    });
   });
 
   describe('#onCreate()', () => {
@@ -104,23 +87,6 @@ describe('DatabaseBuilder', () => {
         },
         resource: 'projects/_/instances/subdomains/refs/users',
       } as any);
-    });
-
-    it('should interpolate params until the server does it', () => {
-      let handler = database.ref('/users/{id}').onCreate(event => {
-        expect(event.resource).to.equal('projects/_/instances/subdomain/refs/users/aUserId');
-      });
-
-      return handler({
-        data: {
-          data: null,
-          delta: 'hello',
-        },
-        resource: 'projects/_/instances/subdomain/refs/users/{id}',
-        params: {
-          id: 'aUserId',
-        },
-      });
     });
   });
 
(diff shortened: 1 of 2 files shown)


piuccio:
this would be amazing!

laurenzlong:
Thanks for the enthusiasm @piuccio and everyone else! We're still waiting on some backend changes, but hope to have this in the next SDK release.

spec/providers/database.spec.ts
```diff
@@ -80,6 +80,136 @@ describe('DatabaseBuilder', () => {
...
+      return handler({
+        data: {
+          data: null,
+          delta: { foo: 'bar' },
+        },
+        resource: 'projects/_/instances/subdomains/refs/users',
+      } as any);
+    });
+
+    it('should interpolate params until the server does it', () => {
```
inlined: The server already does this; we should be able to yank this code.

laurenzlong: Removed.

laurenzlong:
@inlined I'll just leave this PR here then, and you can merge when the backend is ready. Since the other PR is in the other repo, using a launch branch will make things more complicated.

inlined:
I meant that this could go to a launch since it needs to be held back and the other can go to master. NBD either way. The backend change has been merged into GCF so it should go live this week.

1hakr:
@laurenzlong , the `deltasnapshot` return null in` onDelete()` trigger. so the `previous` will not work A user has reported this and i have also faced this [link](https://stackoverflow.com/questions/44998710/cloud-functions-for-firebase-how-to-get-the-value-of-the-data-that-was-deleted). 

I didnt want to create bug so commenting it here

laurenzlong:
@1hakr This is definitely a bug, thanks for reporting!

laurenzlong:
hey @1hakr we aren't able to reproduce this. Can you file an actual Github issue, and include repro instructions? Including exactly how you deleted the node (if it's via code, then add the client side code, if it's via CLI, include the command you used, if it's via console, then exactly what you clicked)

CodingDoug:
@1hakr I was unable to reproduce in the most simple case of an onDelete that does nothing but log the previous value of the data that was deleted:

```
exports.onDelete = functions.database.ref("/messages/{messageId}").onDelete(event => {
    console.log("Stuff got deleted ", event.data.previous.val())
})
```

The correctly logs the deleted values for me.  It sounds like you're seeing something different?

1hakr:
Found the culprit. I was using exists() method to check if data exists or not before doing any operation. I removed the code and it works now. I thought exists() was supposed to tell you if data exists or not. My bad. :neutral_face: 

Here is the code i was using.
```
exports.deletePriceAlert = functions.database.ref('/user_alerts/price/{uid}/{alertId}').onDelete(event => {
  const snapshot = event.data;
  const uid = event.params.uid;
  
  if (!snapshot.exists()) {
    return;
  }

  const comboKey = snapshot.previous.val().name;
  return admin.database().ref(`/alerts/price/${comboKey}/${uid}`).remove().then(result => {
    logInfo("Delete alert", {user: uid, key : comboKey});
  })
  .catch(error => {
    return reportError(error, {user: uid, type: 'database_write', context: 'delete alert'});
  });
});
```

PR #65, opened 2017-06-29
Fix Typescript compiler issues with TS 2.4.1

laurenzlong:
Need to upgrade to TS 2.4.1 and change code according to the stricter compiler, since our release system uses it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "nock": "^8.0.0",
     "sinon": "^1.17.4",
     "tslint": "^3.15.1",
-    "typescript": "^2.4.1"
+    "typescript": "^2.0.3"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "nock": "^8.0.0",
     "sinon": "^1.17.4",
     "tslint": "^3.15.1",
-    "typescript": "^2.0.3"
+    "typescript": "^2.4.1"
```
inlined: This isn't strictly necessary (or at least it can be postponed until after this release). When NPM run build runs, it uses the maximum version that is available. If we raise this version then our TypeScript customers must upgrade too.

laurenzlong: Makes sense, updated.

PR #66, opened 2017-06-30
Update to firebase-admin v5.0.0

laurenzlong:
### Description

Update dependency to firebase-admin v5.0.0, also added shim to transform auth event payload into the new format UserRecord format from firebase-admin.

### DIFF
--- a/spec/providers/auth.spec.ts
+++ b/spec/providers/auth.spec.ts
@@ -91,6 +91,9 @@ describe('AuthBuilder', () => {
       ]);
     });
 
+    // createdAt and lastSignedIn are fields of admin.auth.UserMetadata below v5.0.0
+    // We want to add shims to still expose these fields so that user's code do not break
+    // The shim and this test should be removed in v1.0.0 of firebase-functions
     it('should still retain createdAt and lastSignedIn', () => {
       return Promise.all([
         cloudFunctionCreate(event).then(data => {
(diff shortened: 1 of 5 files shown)


spec/providers/auth.spec.ts
```diff
@@ -102,11 +102,67 @@ describe('AuthBuilder', () => {
...
+        },
+        data: {
+          uid: 'abcde12345',
+          email: 'foo@bar.baz',
+          emailVerified: false,
+          displayName: 'My Display Name',
+          photoURL: 'bar.baz/foo.jpg',
+          disabled: false,
+          metadata: {
+              creationTime: '2016-12-15T19:37:37.059Z',
```
rjhuijsman: Can you add a small comment to call out the difference between the old and new wire formats here?

laurenzlong:
Made some changes according to our deprecation discussion, @rjhuijsman can you take another look? I also removed the test for "should tolerate missing fields in the payload" since the test events for the other tests is already missing fields.

spec/providers/auth.spec.ts
```diff
@@ -61,90 +61,69 @@ describe('AuthBuilder', () => {
...
+          expect(data.metadata.lastSignInTime).to.equal('2017-01-01T00:00:00.000Z');
+        }),
       ]);
     });
 
-    // This isn't expected to happen in production, but if it does we should
-    // handle it gracefully.
-    it('should tolerate missing fields in the payload', () => {
-      const cloudFunction = auth.user().onCreate((ev: Event<firebase.auth.UserRecord>) => ev.data);
+    it('should still retain createdAt and lastSignedIn', () => {
```
rjhuijsman: A comment on what "createdAt" and "lastSignedIn" are and why they should be retained would be useful.

spec/providers/auth.spec.ts
```diff
@@ -61,90 +61,69 @@ describe('AuthBuilder', () => {
...
-          displayName: 'sample user',
-          photoURL: '',
-          disabled: false,
         },
-      } as any;
+      };
 
-      return expect(cloudFunction(event)).to.eventually.deep.equal(event.data);
+      return Promise.all([
+        cloudFunctionCreate(newEvent).then(data => {
```
rjhuijsman: We should also expect that the fields with the old name are still present in this case with the new wire format.

laurenzlong: done

PR #76, opened 2017-07-25
Update peer dependency to firebase-admin v5.1.0

laurenzlong:
### Description

firebase-admin just had a very exciting release that introduces support for phone auth! (https://github.com/firebase/firebase-admin-node/releases/tag/v5.1.0)

This PR updates the peer dependency. I also added firebase-admin back as a dev dependency so that Travis can install the exact version of firebase-admin necessary when running tests. I removed caching of node_modules for Travis so that the tests will be more consistent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,3 @@ node_js:
 - '6.11.1'
 - stable
 sudo: false
-cache:
-  directories:
-  - node_modules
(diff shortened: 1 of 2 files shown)


laurenzlong:
@inlined see my description, I had to add it back to fix Travis.

PR #81, opened 2017-07-27
Make the integration test fully-automatic.

rjhuijsman:
Make the integration test fully-automatic. This entails two major changes:
- In the HTTP Cloud Function that sets off the server-side part of the test, keep track of which tests pass, and return 200 "PASS" when all tests pass, or 500 "FAIL - [debug-info]" when they don't.
- On the client-side, automatically trigger that HTTP function and interpret the return code as a pass or fail.

Also tries to improve the readability of 'run_tests.sh' by indenting appropriately, and tries to improve its usability by more clearly indicating whether the tests have passed or failed.

### DIFF
--- a/integration_test/functions/src/index.ts
+++ b/integration_test/functions/src/index.ts
@@ -7,6 +7,7 @@ import { Request, Response } from 'express';
 export * from './pubsub-tests';
 export * from './database-tests';
 export * from './auth-tests';
+let numTests = 4;  // Make sure to increase this as you add tests!
 
 firebase.initializeApp(_.omit(functions.config().firebase, 'credential'));  // Explicitly decline admin privileges.
 admin.initializeApp(functions.config().firebase);
@@ -16,7 +17,7 @@ export const integrationTests: any = functions.https.onRequest((req: Request, re
 
   const testId = firebase.database().ref().push().key;
 
-  Promise.all([
+  return Promise.all([
     // A database write to trigger the Firebase Realtime Database tests.
     // The database write happens without admin privileges, so that the triggered function's "event.data.ref" also
     // doesn't have admin privileges.
@@ -33,8 +34,33 @@ export const integrationTests: any = functions.https.onRequest((req: Request, re
       admin.auth().deleteUser(userRecord.uid);
     }),
   ]).then(() => {
-    // On test completion, redirect the user to the Firebase RTDB dashboard, to
-    // see the results stored there.
-    resp.redirect(`https://${process.env.GCLOUD_PROJECT}.firebaseio.com/testRuns/${testId}`);
+    // On test completion, check that all tests pass and reply "PASS", or provide further details.
+    console.log('Waiting for all tests to report they pass...');
+    let ref = admin.database().ref(`testRuns/${testId}`);
+    return new Promise((resolve, reject) => {
+      let testsExecuted = 0;
+      ref.on('child_added', (snapshot) => {
+        testsExecuted += 1;
+        if (!snapshot.val().passed) {
+          reject(new Error(`test ${snapshot.key} failed; see database for details.`));
+          return;
+        }
+        console.log(`${snapshot.key} passed (${testsExecuted} of ${numTests})`);
+        if (testsExecuted < numTests) {
+          // Not all tests have completed. Wait longer.
+          return;
+        }
+        // All tests have passed!
+        resolve();
+      });
+    }).then(() => {
+      ref.off();  // No more need to listen.
+      console.log('All tests pass!');
+      resp.status(200).send('PASS');
+    }).catch(err => {
+      ref.off();  // No more need to listen.
+      console.log(`Some tests failed: ${err}`);
+      resp.status(500).send(`FAIL - details at https://${process.env.GCLOUD_PROJECT}.firebaseio.com/testRuns/${testId}`);
+    });
   });
 });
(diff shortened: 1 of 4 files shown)


laurenzlong:
Biggest piece of feedback is that when something fails, it just prints out XXXXX Something failed XXXXXXX

It should at the minimum print out the URL for the database so the user can go check what's wrong.

(PS. you can easily simulate a failure by changing one of the promises in functions/index.ts to RSVP.reject())

failedintegration

integration_test/functions/src/index.ts
```diff
@@ -7,6 +7,7 @@ import { Request, Response } from 'express';
 export * from './pubsub-tests';
 export * from './database-tests';
 export * from './auth-tests';
+let numTests = 4;  // Make sure to increase this as you add tests!
```
rjhuijsman: The tests are not actually not quite the same as the number of promises, but they _are_ the same as the number of exported functions - so I've gone off that. :)

integration_test/run_tests.sh
```diff
@@ -1,28 +1,46 @@
...
+  ) &&
+  echo "##### Installing dependencies..." &&
+  (
+    cd functions &&
+    npm install
+  ) &&
+  echo "##### Deploying empty index.js to project..." &&
+  ./functions/node_modules/.bin/tsc -p functions/ &&  # Make sure the functions/lib directory actually exists.
+  echo "" > functions/lib/index.js &&
+  firebase deploy --debug 2> /dev/null &&
```
rjhuijsman: Done.

integration_test/run_tests.sh
```diff
@@ -1,28 +1,46 @@
...
+  firebase deploy --only functions --debug 2> /dev/null &&
+  echo &&
+  echo "##### Running the integration tests..." &&
+  wget --content-on-error -qO- $funcURL &&
+  echo &&
+  echo "##### Integration test run passed!" &&
+  echo "##### Removing all functions" &&
+  echo "" > functions/lib/index.js &&
+  firebase deploy --debug 2> /dev/null &&
+  rm functions/firebase-functions.tgz &&
```
rjhuijsman: Done.

rjhuijsman:
Good catch that that type of failure got less logs than they should - I've rejiggered the promises in index.ts so that those failures are also caught and logged with error-500.

integration_test/run_tests.sh
```diff
@@ -15,8 +15,8 @@
   echo "##### Deploying empty index.js to project..." &&
   ./functions/node_modules/.bin/tsc -p functions/ &&  # Make sure the functions/lib directory actually exists.
   echo "" > functions/lib/index.js &&
-  firebase deploy --debug 2> /dev/null &&
-  echo &&
+  firebase deploy --only functions --debug 2> /dev/null &&
+ echo &&
```
laurenzlong: I think you're missing one space here.

rjhuijsman: Fixed.

integration_test/functions/src/index.ts
```diff
@@ -33,8 +34,37 @@ export const integrationTests: any = functions.https.onRequest((req: Request, re
       admin.auth().deleteUser(userRecord.uid);
     }),
   ]).then(() => {
-    // On test completion, redirect the user to the Firebase RTDB dashboard, to
-    // see the results stored there.
-    resp.redirect(`https://${process.env.GCLOUD_PROJECT}.firebaseio.com/testRuns/${testId}`);
+    // On test completion, check that all tests pass and reply "PASS", or provide further details.
```
inlined: Eventually we may want to break this into client + server. The HTTP function could return the test ID and then a client would poll for the ID. This would help us avoid stricter cutoffs in maximum HTTP request duration.

integration_test/functions/src/index.ts
```diff
@@ -33,8 +34,37 @@ export const integrationTests: any = functions.https.onRequest((req: Request, re
...
+    console.log('Waiting for all tests to report they pass...');
+    let ref = admin.database().ref(`testRuns/${testId}`);
+    return new Promise((resolve, reject) => {
+      let testsExecuted = 0;
+      ref.on('child_added', (snapshot) => {
+        testsExecuted += 1;
+        if (!snapshot.val().passed) {
+          reject(new Error(`test ${snapshot.key} failed; see database for details.`));
```
inlined: Can we add a link?

integration_test/functions/src/index.ts
```diff
@@ -33,8 +34,37 @@ export const integrationTests: any = functions.https.onRequest((req: Request, re
...
+        if (testsExecuted < numTests) {
+          // Not all tests have completed. Wait longer.
+          return;
+        }
+        // All tests have passed!
+        resolve();
+      });
+    }).then(() => {
+      ref.off();  // No more need to listen.
+      return Promise.resolve();
```
inlined: Unnecessary

integration_test/functions/src/index.ts
```diff
@@ -33,8 +34,37 @@ export const integrationTests: any = functions.https.onRequest((req: Request, re
...
+    }).catch(err => {
+      ref.off();  // No more need to listen.
+      return Promise.reject(err);
+    });
+  }).then(() => {
+    console.log('All tests pass!');
+    resp.status(200).send('PASS');
+  }).catch(err => {
+    console.log(`Some tests failed: ${err}`);
+    resp.status(500).send(`FAIL - details at https://${process.env.GCLOUD_PROJECT}.firebaseio.com/testRuns/${testId}`);
```
inlined: Nits:
1. the GCLOUD_PROJECT is not necessarily the subdomain to firebaseio.com; you can use functions.config().firebase.databaseUrl.
2. I think the URL has to end in .json if you're linking to a database URL.
3. You might actually want to consider a pantheon so it will try to load the database using the developer credentials rather than unauthenticated credentials.

PR #92, opened 2017-08-24
Preserve params passed with the original event

### LINKED ISSUE #88: `event.params` is undefined in unit tests
### Version info
**firebase-functions:** 0.6.2
**firebase-tools:** 3.9.2
**firebase-admin:** 5.2.1

### Test case

```ts
//
// In unit test
//
// ...
event = {
  params: {
    userId: 'user-id',
  },
  data: new functions.database.DeltaSnapshot(null, admin, null, user)
};
console.log(event); // 1st log
await myFunctions.onUserWrite(event);
// ...
```
```ts
//
// In onUserWrite
//
export.onUserWrite = functions.database.ref('/users/userId').onWrite(async event => {
  // ...
  console.log(event); // 2nd log
  // ...
);
```

No, since unit tests fail

### Expected behavior

1st log should equal 2nd log, showing both `event.params.userId` as `user-id`

### Actual behavior

1st log shows `event.params.userId` as `user-id`
2nd log shows `event.params.userId` as `undefined`; However, the `event.data` element is equal to the 1st log!

---

Reverting back to **firebase-functions:** 0.6.1 fixes the issue.

Crazometer:
### Description

Fixes #88. Parameters assigned to event.params take precedence over those from the resource string.

### DIFF
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -74,7 +74,7 @@ export interface MakeCloudFunctionArgs<EventData> {
 }
 
 function _makeParams(event: Event<any>, triggerResource: string): { [option: string]: any } {  
-  if (!event.resource) {
+  if (!event.resource) { // In unit testing, "resource" may not be populated for a test event.
     return event.params || {};
   }
 
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -73,7 +73,7 @@ export interface MakeCloudFunctionArgs<EventData> {
   after?: (raw: Event<any>) => void;
 }
 
-function _makeParams (event: Event<any>, triggerResource: string): { [option: string]: any } {
+function _makeParams(event: Event<any>, triggerResource: string): { [option: string]: any } {
   let wildcards = triggerResource.match(WILDCARD_REGEX);
   let params = {};
   if (wildcards) {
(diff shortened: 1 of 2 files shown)


src/cloud-functions.ts
```diff
@@ -86,6 +86,9 @@ function _makeParams (event: Event<any>, triggerResource: string): { [option: st
       params[wildcardNoBraces] = eventResourceParts[position];
     });
   }
+
+  Object.assign(params, event.params);
```
laurenzlong: The general principle I'd like to follow is that event.resource is the source of truth. But if that is absent (as in the case of some unit tests), then event.params should be used. I'm not sure how I feel about merging.

Could you do something to the beginning of the function along the lines of:
```
if (!triggerResource) {
   return event.params || {};
}
```

Crazometer:
@laurenzlong  Is this what you had in mind?

laurenzlong:
Yes, thanks for catching that I meant "event.resource", not "triggerResource". Could you also add a comment to explain that this is to ensure that developers can write unit tests where they only define params, and not resources in the test event. (To ensure someone else doesn't accidentally remove this in the future). Other than that, it is good to go!

laurenzlong:
Actually, never mind, I added the comment. Thanks for making the PR!

PR #97, opened 2017-08-31
Refactor Integration Test Script

justinrosenthal:
This PR refactors the integration test script:
1. Require a command line parameter for specifying project_id.
2. Use `set -e` instead of `&&` chaining.
3. Modularize into reusable functions.
4. Use curl instead of wget as it's available on OS X by default.
5. Use absolute paths so the script can be run from outside the `integration_test` directory, and to make the `cd` statements independent of each other.
6. Don't suppress stderr during `firebase deploy`.

### DIFF
--- a/integration_test/run_tests.sh
+++ b/integration_test/run_tests.sh
@@ -1,47 +1,76 @@
-#!/bin/bash
-(
-  echo "##### Building SDK..." &&
-  (
-    cd ../ &&
-    rm -f firebase-functions-*.tgz &&
-    npm run build:pack &&
-    mv firebase-functions-*.tgz integration_test/functions/firebase-functions.tgz
-  ) &&
-  echo "##### Installing dependencies..." &&
-  (
-    cd functions &&
-    npm install
-  ) &&
-  echo "##### Deploying empty index.js to project..." &&
-  ./functions/node_modules/.bin/tsc -p functions/ &&  # Make sure the functions/lib directory actually exists.
-  echo "" > functions/lib/index.js &&
-  firebase deploy --debug 2> /dev/null &&
-  echo &&
-  echo "##### Project emptied. Deploying functions..." &&
-  ./functions/node_modules/.bin/tsc -p functions/ &&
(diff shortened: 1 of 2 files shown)


integration_test/run_tests.sh
```diff
@@ -1,47 +1,76 @@
...
+  echo "" > functions/lib/index.js
+  firebase deploy --project=$PROJECT_ID --only functions
+  announce "Project emptied."
+}
+
+function deploy {
+  announce "Deploying functions..."
+  cd $DIR
+  ./functions/node_modules/.bin/tsc -p functions/
+  firebase deploy --project=$PROJECT_ID --only functions
```
rjhuijsman: This does need "--debug 2> /dev/null" to generate the firebase-debug.log file. Else a clean user will get this error:

> ##### Running the integration tests...
> tail: cannot open ‘/usr/local/google/home/rjh/src/firebase-functions/integration_test/firebase-debug.log’ > for reading: No such file or directory
> curl: no URL specified!
> curl: try 'curl --help' or 'curl --manual' for more information

justinrosenthal: Oh I see, since we're depending on scraping the debug log and usually it gets deleted.  The `2> /dev/null` is the unfortunate part because it suppresses deploy errors.  Without it, though, we get tons of debug lines spamming the output.

justinrosenthal: In 1937e8e I removed the log scraping in favor of just manually constructing the URL.  I'm open to other suggestions, but this seemed like the least-bad approach.  Additionally, I'd argue it's maybe even less brittle than relying on a very specific debug output format of `firebase-tools`.  Lmk what you think.

rjhuijsman: Thanks! This seems the least-bad approch.

rjhuijsman:
I will take the liberty of merging this PR, so that the Warsaw team is unblocked if they try to do integration testing their-daytime-today (our tonight).

PR #98, opened 2017-08-31
Add code coverage reports to unit tests

inlined:
### Description

Adds automatic test coverage reports to `npm test` via the `nyc` commandline tool for `istanbul`. Future changes may cause a test failure if code coverage drops too low.

### Code sample

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.nyc_output/
 .idea
 .npm
 .tmp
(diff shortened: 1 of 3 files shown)


inlined:
I think we've let good be the enemy of great since we still don't have code coverage. OTOH, I'm going to close this so I can have a clean inbox.

PR #103, opened 2017-09-13
Removed obsolete folder spec/fixtures

laurenzlong:
### Description

These test fixtures are no longer needed.


spec/fixtures/credential/key.json
```diff
@@ -1,7 +0,0 @@
-{
```
bklimt: Would you mind leaving this one file here for now? I'm going to need it in a change I'm working on right now, and I can move it out as part of that change.

laurenzlong:
Let's just remove these later then.

PR #108, opened 2017-09-19
Add a warning when the developer's function does not return a Promise or value.

inlined:
### Description

This fixes a regression introduced long ago when we added our own Promise wrapper in
the SDK. Because our SDK always returns a promise, we cannot rely on the Cloud Functions
environment to emit the warning log.

As a minor cleanup, I rewrote makeCloudFunction to use await/async. This hopefully makes
before/after much easier to read because it uses try/finally.

### Code sample

The following code sample:

```typescript
export let noReturn = functions.database.ref('uppercaser/{message}').onCreate(event => {
  event.data.ref.set(event.data.val().toUpperCase());
});
```

will now log the following warning:

```bash
Function returned undefined, expected Promise or value
```

### DIFF
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -92,7 +92,7 @@ function _makeParams(event: Event<any>, triggerResource: string): { [option: str
   }
 
   return params;
-};
+}
 
 /** @internal */
 export function makeCloudFunction<EventData>({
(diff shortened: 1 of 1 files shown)


mbleigh:
@inlined why merged without review? Is this super-duper urgent?

laurenzlong:
LGTM

laurenzlong:
@mbleigh It was an accident, Thomas mistook the green checkmark from Travis as an approved review. I reviewed it after the fact.

inlined:
Yeah, sorry. I was quickly going through a list of 6 CRs and wasn't thinking clearly.

PR #123, opened 2017-10-23
Handle Firestore timestamp values with precision to the second

laurenzlong:
### Description

Fixes bug where timestamp values that didn't have millisecond or nanosecond precision resulted in "Invalid date" when event.data.data() is called.

### DIFF
--- a/src/encoder.ts
+++ b/src/encoder.ts
@@ -32,9 +32,5 @@ export function dateToTimestampProto(timeString) {
     const trailingZeroes = 9 - nanoString.length;
     nanos = parseInt(nanoString, 10) * Math.pow(10, trailingZeroes);
   }
-
-  return {
-    seconds: seconds,
-    nanos: nanos,
-  };
+  return { seconds, nanos };
 };
--- a/src/encoder.ts
+++ b/src/encoder.ts
@@ -20,18 +20,20 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+import * as _ from 'lodash';
+
 export function dateToTimestampProto(timeString) {
   if (typeof timeString === 'undefined') {
     return;
   }
   let date = new Date(timeString);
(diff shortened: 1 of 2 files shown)


src/encoder.ts
```diff
@@ -20,18 +20,20 @@
...
-  if (nanoString.length === 3) {
-    nanoString = `${nanoString}000000`;
-  } else if (nanoString.length === 6)  {
-    nanoString = `${nanoString}000`;
+  let nanoString;
+  if (timeString.length <= 20) {
+    nanoString = '000000000';
+  } else {
+    nanoString = timeString.substring(20, timeString.length - 1);
+    nanoString += _.reduce(new Array(9 - nanoString.length), i => {return i + '0';}, '');
```
inlined: Woah... String -> array -> reduce -> parseInt? It may be the c programmer in me, but string manipulation of numbers feels really wonky.

why not something like

```typescript
let nanos: 0;
if (timeString.length > 20) {
  const nanoString = timeString.substring(20, timeString.length - 1);
  const trailingZeros = 9 - nanoString.length;
  nanos = parseInt(nanoString, 10) * Math.pow(10, trailingZeroes);
}
```

laurenzlong:
@inlined That is way cleaner, thank you.

src/encoder.ts
```diff
@@ -26,16 +26,15 @@ export function dateToTimestampProto(timeString) {
...
-    nanoString = `${nanoString}000`;
+  let nanos = 0;
+  if (timeString.length > 20) {
+    const nanoString = timeString.substring(20, timeString.length - 1);
+    const trailingZeroes = 9 - nanoString.length;
+    nanos = parseInt(nanoString, 10) * Math.pow(10, trailingZeroes);
   }
-  let proto = {
+
+  return {
```
inlined: You can also say `return {seconds, nanos}`
It felt weird to me at first, but it starts to feel pretty natural and expressive. It also encourages you to use variable names as they'll appear in your return values/API calls which tends to lead to more readable code.

laurenzlong: nifty! thanks for the tip!

PR #136, opened 2017-11-28
Always construct DocumentSnapshot for event.data.previous for Firestore functions

laurenzlong:
### Code sample

This changes what event.data.previous is when there's no previous data (i.e. the Firestore document was just created)

```
// Before this PR:
event.data.previous // null
event.data.previous.exists // would throw

// After this PR
event.data.previous // is a firestore.DocumentSnapshot
event.data.previous.exists // false
```
IMPORTANT: need to make sure docs/sample code isn't suggesting the wrong thing before we merge this PR (tracking bug 69812710)

### DIFF
--- a/spec/providers/firestore.spec.ts
+++ b/spec/providers/firestore.spec.ts
@@ -145,7 +145,10 @@ describe('Firestore Functions', () => {
       let testFunction = firestore.document('path').onCreate((event) => {
         expect(event.data.data()).to.deep.equal({key1: true, key2: 123});
         expect(event.data.get('key1')).to.equal(true);
-        expect(event.data.previous).to.equal(null);
+        expect(event.data.previous).to.not.equal(null);
+        expect(event.data.previous.exists).to.be.false;
+        expect(event.data.previous.data).to.throw(Error);
+        expect(() => {return event.data.previous.get('key1');}).to.throw(Error);
       });
       let data = constructEvent({}, createValue());
       return testFunction(data);
--- a/spec/providers/firestore.spec.ts
+++ b/spec/providers/firestore.spec.ts
@@ -333,7 +333,7 @@ describe('Firestore Functions', () => {
         let snapshot = firestore.dataConstructor({
           data: { value: raw },
         });
-        expect(snapshot.data()).to.deep.equal({'binaryVal': 'Zm9vYmFy'});
+        expect(snapshot.data()).to.deep.equal({'binaryVal': new Buffer('foobar')});
       });
     });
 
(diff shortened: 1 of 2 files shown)


laurenzlong:
Some additions:
- Behavior for byte values changed in the new Firestore SDK, and now a buffer is returned in snapshot.data()
- Stopped converting JSON to fieldsProto since snapshot constructor takes in json.

spec/providers/firestore.spec.ts
```diff
@@ -145,7 +145,10 @@ describe('Firestore Functions', () => {
       let testFunction = firestore.document('path').onCreate((event) => {
         expect(event.data.data()).to.deep.equal({key1: true, key2: 123});
         expect(event.data.get('key1')).to.equal(true);
-        expect(event.data.previous).to.equal(null);
+        expect(event.data.previous).to.not.equal(null);
+        expect(event.data.previous.exists).to.be.false;
+        expect(event.data.previous.data).to.throw(Error);
```
inlined: Why isn't event.data.previous.data just null?

mikelehen: cc/ @schmidt-sebastian 

Lauren's implementation matches the client implementation (thanks!) but we are actually intending to change this so that .data is null for nonexistent documents (see http://b/68270401).  So I think it would probably be fine (and future looking even) to make .data null here. I believe the difficulty that the clients have in making this change (that DocumentSnapshots are used to surface query results where the document is guaranteed to exist and therefore .data should not be nullable) doesn't apply to Functions.

laurenzlong: @mikelehen `event.data.previous` is a constructed DocumentSnapshot. So I didn't do any additional processing to it. Once the Firestore Node SDK makes .data null, this will be null as well.

mikelehen: Ah, right! I forgot we're sharing code (yay!).

inlined: Ooof. Should we make the unit test support both cases since it will otherwise randomly break in CI once the latest node module is picked up (assuming it's considered a revisional fix and not a breaking change)

laurenzlong:
@inlined I modified the tests to not check for throw error behavior on .data

spec/providers/firestore.spec.ts
```diff
@@ -147,8 +147,6 @@ describe('Firestore Functions', () => {
         expect(event.data.get('key1')).to.equal(true);
         expect(event.data.previous).to.not.equal(null);
         expect(event.data.previous.exists).to.be.false;
```
inlined: If you prefer we can still keep the value of the test with something like

```typescript
// older versions throw, newer versions will return null
try {
  let val = event.data.previous.get('key1');
  expect(val).to.be.null;
  let data = event.data.previous.data;
  expect(data).to.be.null;
} catch (err) {
  // ignored; throwing is OK for older SDKs
}
```

laurenzlong: I'm tempted to remove them to keep things cleaner.

laurenzlong:
I also submitted a documentation change (#178566239)

Issue #137, opened 2017-11-30
`event.data.val()` is `null` in unit tests


zgosalvez:
### Version info
**firebase-functions:** 0.7.3
**firebase-tools:** 3.15.3
**firebase-admin:** 5.5.1

### Test case

```ts
//
// In unit test
//
// ...
user = {
  firstName: 'me'
};
event = {
  params: {
    userId: 'user-id',
  },
  data: new functions.database.DeltaSnapshot(null, null, null, user)
};
console.log(event.data.val()); // 1st log
await myFunctions.onUserWrite(event);
// ...
```
```ts
//
// In onUserWrite
//
export.onUserWrite = functions.database.ref('/users/userId').onWrite(async event => {
  // ...
  console.log(event.data.val()); // 2nd log
  // ...
);
```

### Were you able to successfully deploy your functions?

No, since unit tests fail, but manually deploying (skipping unit tests) shows no issues.

### Expected behavior

1st and 2nd logs, should both show `user` object.

### Actual behavior

- 1st log shows `null`
- 2nd log shows `null`

---

Reverting back to **firebase-functions:** 0.7.2 fixes the issue.

laurenzlong:
@zgosalvez Thanks for pointing this out! We recently changed the signature of DeltaSnapshot, and this led to the accidental affect of breaking unit tests. I'll make a PR to mend this.

zgosalvez:
Thanks, @laurenzlong ! When can we expect this fix to be released?

laurenzlong:
It'll be released in the next few days, you can see the release notes at https://github.com/firebase/firebase-functions/pull/145

PR #139, opened 2017-11-30
Change order of parameters for DeltaSnapshot constructor

laurenzlong:
### Description

Fixes https://github.com/firebase/firebase-functions/issues/137, so users doing uni testing can still construct DeltaSnapshots in the same way as before multi-DB was introduced.

Also made instance both a public and an optional field. This is ok since the admin SDK will initialize the database for the default instance if no parameter is supplied.


laurenzlong:
Yeah, I think this was a case of: accidentally changed a public API, and now reversing it.

PR #150, opened 2017-12-20
Revert "Always construct DocumentSnapshot for event.data.previous for Firestore functions"

laurenzlong:
Reverts firebase/firebase-functions#136


inlined:
LGTM

PR #162, opened 2018-01-09
Correct the type signature of DeltaDocumentSnapshot

Byeol:
### Description

Correct the type signature of `DeltaDocumentSnapshot` to make it correspond to the [reference documentation](https://firebase.google.com/docs/reference/functions/functions.firestore.DeltaDocumentSnapshot).

### Code sample

### DIFF
--- a/src/providers/firestore.ts
+++ b/src/providers/firestore.ts
@@ -70,12 +70,12 @@ export class NamespaceBuilder {
 
 export interface DeltaDocumentSnapshot {
   exists: Boolean;
-  ref: any;
+  ref: firebase.firestore.DocumentReference;
   id: string;
-  createTime: string;
-  updateTime: string;
-  readTime: string;
-  previous: any;
+  createTime?: string;
+  updateTime?: string;
+  readTime?: string;
+  previous: DeltaDocumentSnapshot;
   data: () => any;
   get: (key: string) => any;
 };


src/providers/firestore.ts
```diff
@@ -70,12 +71,12 @@ export class NamespaceBuilder {
...
 export interface DeltaDocumentSnapshot {
   exists: Boolean;
-  ref: any;
+  ref: firestore.DocumentReference;
   id: string;
-  createTime: string;
-  updateTime: string;
-  readTime: string;
-  previous: any;
+  createTime: string | null;
```
laurenzlong: Can you change the cases of "field: string | null" to be "field?: string"

Byeol: @laurenzlong Changed. Thanks! :)

src/providers/firestore.ts
```diff
@@ -23,6 +23,7 @@
 import { posix } from 'path';
 import * as _ from 'lodash';
 import * as firebase from 'firebase-admin';
+import * as firestore from '@google-cloud/firestore';
```
laurenzlong: sorry i previously missed this, but we don't actually pull in @google-cloud/firestore as a dependency ,since it's already a part of firebase-admin. So please remove this line and change line 74 to "ref: admin.firestore.DocumentReference". You'll have to test this, but I'm pretty sure that will work.

laurenzlong: Never mind, I take that back. It turns out firebase-admin only exposes firestore types that have a public constructor, which does not include DocumentReference. 

I think the way you have it should be fine, but I'm not sure if it's against best practice to require a module that is not in package.json, but is a sub-dependency. @inlined Can you comment on this?

Byeol: @laurenzlong You're right. I tried to do that, but `firebase-admin` doesn't expose `DocumentReference`.

IMHO, I'm not pretty sure if it's a good idea to require a module that is not in package.json. What do you think about adding `@google-cloud/firestore` as a `devDependency` explicitly? I think there is no other way to access type declaration of `DocumentReference`.

inlined: 1. This is a bug on Firestore/firebase-admin to export all of their types. We can work with @schmidt-sebastian. It seems like firebase-admin could use some of the tricks we use in firebase-functions to reexport namespaces.
2. If we don't actually `require` the module in the transpiled JavaScript I'm fine moving this as a dev dependency, though we might want to track an issue to clean this up. I suspect it might be elided since firestore is never used directly, though I can suggest some tricks if it is showing up.

schmidt-sebastian: @hiranya911 regarding types in firebase-admin.

laurenzlong: FYI fix in firebase-admin has been merged to master: https://github.com/firebase/firebase-admin-node/pull/193f

Byeol: @laurenzlong Thanks for letting me know. I'll update this PR once the next version of `firebase-admin` is released.

laurenzlong:
Hey @Byeol firebase-admin v5.8.2 has been released now, so you should be able to get rid of the google-cloud/firestore dependency now

SUPERCILEX:
@laurenzlong I'm pretty sure the issues you brought up were rebased into this PR since it looks good to go. Would be nice to have real types! 😄

PR #168, opened 2018-01-19
Normalize path in InstanceBuilder

### LINKED ISSUE #167: If specifying "instance" on RTDB trigger and path has leading '/' then two '/'s are created in Resource path
When creating an RTDB trigger for a cloud function using in `database.instance('some_instance')`, the resource path appears to get an extra slash in the path if the string you feed to `.ref()` includes a leading slash.

If the ref string has a leading slash when not specifying an instance, there is no second slash.

I believe the library needs to call `normalizePath` somewhere in here: https://github.com/firebase/firebase-functions/blob/bb51613c9188ecbbed635210dd8d1ac7a1ab085c/src/providers/database.ts#L47

### Version info

"firebase-admin": "~5.6.0",
"firebase-functions": "~0.8.0",
"firebase-tools": "3.17.1",

### Test case

Create two function triggers like this:

```
const trigger1 = functions.database.ref('/trigger_path').onCreate(someHandler);
const trigger2 = functions.database.instance('instance2').ref('/trigger_path').onCreate(someHandler);
```
Then go to the Google Cloud Functions console, go to `trigger2` function and observe that the Resource will look like: `projects/_/instances/instance2/refs//trigger_path`

(Note extra / between refs and trigger_path)

deremer:
Fixes #167

### DIFF
--- a/src/providers/database.ts
+++ b/src/providers/database.ts
@@ -45,7 +45,8 @@ export class InstanceBuilder {
   constructor(private instance: string) {}
 
   ref(path: string): RefBuilder {
-    return new RefBuilder(apps(), `projects/_/instances/${this.instance}/refs/${path}`);
+    const normalized = normalizePath(path);
+    return new RefBuilder(apps(), `projects/_/instances/${this.instance}/refs/${normalized}`);
   }
 }
 


deremer:
I signed it

PR #195, opened 2018-02-28
Update firebase-admin to version ^5.9.0

ajcrews:
### Description

Upgrading `firebase-admin` to the latest. Also relaxing the semver to a caret so we don't have to patch this package for every minor update to `firebase-admin`. I noticed that a caret was previously used, then it was a mix (dep used caret, peer dep used tilde), now both are on the tilde. Please confirm the intention.


laurenzlong:
Thanks for the PR! We now use tilde because minor updates to firebase-admin can introduce breaking changes (at least for Typescript type definitions).

ajcrews:
Thanks @laurenzlong! Updated.

PR #205, opened 2018-03-20
Warn users about breaking changes

tinaliang:
### Description

firebase-functions SDK will print out a message warning users about breaking changes when being upgraded. This postinstall script can be removed in later versions.

To test: run `npm install` inside a functions directory after upgrading the `package.json` version number to 1.0 for firebase-functions.

### DIFF
--- a/upgrade-warning
+++ b/upgrade-warning
@@ -1,0 +1,14 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const message = `
+Congratulations! You have just updated to firebase-functions 1.0.
+
+This means you've installed some breaking changes. To see a complete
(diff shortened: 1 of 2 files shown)


upgrade-warning
```diff
@@ -0,0 +1,14 @@
...
+const message = `
+Congratulations! You have just updated to firebase-functions 1.0.
+
+This means you've installed some breaking changes. To see a complete
+list of these breaking changes, please go to
+
+https://firebase.google.com/docs/functions/early-access/beta-v1-diff
+`;
+
+console.log(message);
```
laurenzlong: should add newline to end of file.

PR #206, opened 2018-03-21
 Change https callables to use apps.admin.auth instead of firebase.auth.

bklimt:
This fixes the problem where an https callable function will reject all requests with an auth token if the function has not called firebase.initializeApp.

I have tested this with `npm test`, and with a function deployed to an app in production using this branch.

### DIFF
--- a/spec/providers/https.spec.ts
+++ b/spec/providers/https.spec.ts
@@ -21,15 +21,12 @@
 // SOFTWARE.
 
 import * as express from 'express';
-import * as firebase from 'firebase-admin';
 import * as https from '../../src/providers/https';
 import * as jwt from 'jsonwebtoken';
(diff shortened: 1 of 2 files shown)


spec/providers/https.spec.ts
```diff
@@ -361,7 +375,7 @@ describe('callable.FunctionBuilder', () => {
 
     it('should handle auth', async () => {
       const mock = mockFetchPublicKeys();
-      const projectId = config.singleton.firebase['projectId'];
+      const projectId = 'aProjectId';
```
laurenzlong: replace with: 
const projectId = apps().admin.options.projectId;

bklimt: Done

PR #229, opened 2018-04-24
Update firebase-admin in integration tests to v5.12.0

swalkowski:
Updating firebase-admin version in the integration test dependencies to v5.12.0 so that it matches the peer dependency requirement of firebase-functions. This fixes the integration tests.

/cc @laurenzlong for review


laurenzlong:
Thanks for contributing!

PR #232, opened 2018-04-24
`context` parameter is no longer optional

### LINKED ISSUE #227: Event context can be undefined according to type definitions
According to the 1.0.0 Typescript definitions the context part of an event is defined as `EventContext | undefined`. For example:

```ts
const myFunction = firestore
  .document(`collection/{documentId}`)
  .onCreate((data, context) => {
    // Check if context is defined?
  });
```

I have a hard time imagining why this would ever be undefined, so I am currently ignoring it. 

Context as the second parameter to the handler function is optional, because you can supply a handler with just the `data` argument. But shouldn't that be solved via overloading? If I supply a function with both arguments then context should be guaranteed to exist, no?

merlinnot:
I've gone through the source code and it looks like `context` is always defined, so I made a type definition more strict.

Sorry that I didn't fit a changelog entry within 80 columns, I didn't know how your helper tools parses these entries so I didn't want to split it into multiple lines.
 
Fixes #227

### DIFF
--- a/integration_test/functions/src/testing.ts
+++ b/integration_test/functions/src/testing.ts
@@ -2,7 +2,7 @@ import * as firebase from 'firebase-admin';
 import * as _ from 'lodash';
 import { EventContext } from 'firebase-functions';
 
-export type TestCase<T> = (data: T, context?: EventContext) => any
+export type TestCase<T> = (data: T, context: EventContext) => any
 export type TestCaseMap<T> = { [key: string]: TestCase<T> };
 
 export class TestSuite<T> {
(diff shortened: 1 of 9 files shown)


merlinnot:
@laurenzlong May I kindly ask for a review?

laurenzlong:
Sorry this slipped through the cracks! Thanks for the reminder. Everything looks good, except can you remove the changes you made to the changelog? We generally only update the changelog right before each release, with the full list of features in that release.

merlinnot:
Done :)

PR #254, opened 2018-05-28
Make `context` argument in `run` optional again

swalkowski:
This fixes https-tests in integration_test. An HTTP-triggered function takes only one argument and passes it to `run` so the `context` argument cannot be required in this case.


laurenzlong:
Closing in lieu of https://github.com/firebase/firebase-functions/pull/250

PR #262, opened 2018-06-09
Pass request headers to CallableContext

lpellegr:
### Description

The purpose of this PR is to make the headers of the request handled by an HTTPS Callable accessible to API users by passing request headers to `CallableContext`.

Getting access to headers is convenient for several use cases. For instance, retrieving the IP of the client invoking the callable (`x-forwarded-for`), detecting client language (`accept-language`), etc. 

### Code sample

Here is how to use the new property:

```
functions.https.onCall(async (data, context) => {
    const clientIp = context.headers['x-forwarded-for'];
    // ...
});
```

### DIFF
--- a/spec/providers/https.spec.ts
+++ b/spec/providers/https.spec.ts
@@ -443,6 +443,27 @@ describe('callable.FunctionBuilder', () => {
         },
       });
     });
+
+    it('should handle headers', async () => {
+      const mockRequest = request(null, 'application/json', {});
+      await runTest({
+        httpRequest: mockRequest,
+        expectedData: null,
+        callableFunction: (data, context) => {
+          expect(context.auth).to.be.undefined;
+          expect(context.headers).to.not.be.undefined;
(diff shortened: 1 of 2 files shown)


lpellegr:
@laurenzlong Could you please have a look at this pull-request?

laurenzlong:
Thanks for the PR! We are actually planning to expose the full raw request (not just the header) as a field in the context. We're currently going through an internal API discussion/approval process for this. Once approved, I'd love work with you to tweak this PR to accommodate that change.

lpellegr:
Great! Thanks for the update. I would be happy to help.

laurenzlong:
Awesome! Thanks for your willingness to help! The approved API is to add a field to CallableContext called "rawRequest", which would be the type "express.Request", and it would be the entire HTTP request. Let me know if you have any questions or need any help.

lpellegr:
I have created a new PR (#269) to prevent any confusion due to the title of this PR.

PR #269, opened 2018-06-19
Pass the raw Express request to CallableContext

lpellegr:
This PR generalizes what was originally proposed in #262. 

Here is a summary of the spec shared by laurenzlong:

> The approved API is to add a field to CallableContext called "rawRequest", which would be the type "express.Request", and it would be the entire HTTP request. Let me know if you have any questions or need any help.

### DIFF
--- a/spec/providers/https.spec.ts
+++ b/spec/providers/https.spec.ts
@@ -443,6 +443,24 @@ describe('callable.FunctionBuilder', () => {
         },
       });
     });
+
+    it('should expose raw request', async () => {
+      const mockRequest = request(null, 'application/json', {});
+      await runTest({
+        httpRequest: mockRequest,
+        expectedData: null,
+        callableFunction: (data, context) => {
+          expect(context.rawRequest).to.not.be.undefined;
+          expect(context.rawRequest).to.equal(mockRequest);
+          return null;
+        },
+        expectedHttpResponse: {
+          status: 200,
+          headers: expectedResponseHeaders,
+          body: {result: null},
+        },
+      });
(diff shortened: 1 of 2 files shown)


spec/providers/https.spec.ts
```diff
@@ -443,6 +443,26 @@ describe('callable.FunctionBuilder', () => {
         },
       });
     });
+
+    it('should handle raw request', async () => {
```
laurenzlong: rephrase to: "should expose raw request"

spec/providers/https.spec.ts
```diff
@@ -443,6 +443,26 @@ describe('callable.FunctionBuilder', () => {
...
+      const mockRequest = request(null, 'application/json', {});
+      await runTest({
+        httpRequest: mockRequest,
+        expectedData: null,
+        callableFunction: (data, context) => {
+          expect(context.auth).to.be.undefined;
```
laurenzlong: This line can be removed, distracts from seeing what the point of the test is.

spec/providers/https.spec.ts
```diff
@@ -443,6 +443,26 @@ describe('callable.FunctionBuilder', () => {
...
+          expect(context.instanceIdToken).to.be.undefined;
```
laurenzlong: Same here, this line can be removed.

lpellegr:
Thanks for the review. Your suggestions have been applied.

laurenzlong:
Looks great! Thanks again for taking the time to do this!

Issue #283, opened 2018-07-25
version 2.0.0 causes timeout everywhere


NewITCorp:
Hi,
Today I upgrade firebase functions from 1.1.0 to 2.0.0 and deploy using exactly the same index.js
but all my 22 functions are now timeout every time and everywhere. (almost none is success)

![untitled](https://user-images.githubusercontent.com/26928577/43177103-d039f428-8ff0-11e8-93ce-10aa08fe7f4b.png)

after one hour, I rolled back a version in package.json to 1.1.0 and re-deploy ,everything is back to normal now. 

this is extremely scary for me since my project is on production right now.
any suggestion what i am doing wrong?

thanks
Paul

laurenzlong:
Hi Paul, did you by any chance update your runtime to Node 8 by changing your functions/package.json? And if you wouldn't mind, could you share the code for one of your functions?

laurenzlong:
And are you using any Firestore-triggered functions, we have a breaking change in v2.0.0 for timestamp values.

NewITCorp:
Hi @laurenzlong,
Yes, I do using a firestore-trigger functions.
I have already changed the timestamp values by adding a setting to it but not sure if it is correct.

here is an index.js
[index.txt](https://github.com/firebase/firebase-functions/files/2226377/index.txt)

here is package.json (I already changed the firebase functions from 2.0.0 to 1.1.0)
[package.txt](https://github.com/firebase/firebase-functions/files/2226386/package.txt)

> did you by any chance update your runtime to Node 8 by changing your functions/package.json?

not sure, could you please advise how to check?

thanks for a nice support

walleXD:
I am getting the same issue. I am using `auth.user().onCreate` & `auth.user().onDelete` functions to create using profile documents in `firestore`. For some reason both functions end with a timeout and don't end cleanly even though the functions are able to successfully create the document in `firestore`

Here are my functions:
``` es6
export const createNewUserProfile = auth.user().onCreate(async ({ uid }) => {
  const db = admin.firestore()
  const collection = db.collection("/userProfiles")
  const doc = collection.doc(`${uid}`)

  try {
    await doc.set({})
  } catch (e) {
    throw new Error(e.message)
  }
})

export const deleteUserProfile = auth.user().onDelete(async ({ uid }) => {
  const db = admin.firestore()
  const collection = db.collection("/userProfiles")
  const doc = collection.doc(`${uid}`)

  try {
    await doc.delete()
[...]

PR #286, opened 2018-07-25
Fix error where Node 6 functions were timing out

laurenzlong:
### Description

Fixes https://github.com/firebase/firebase-functions/issues/283

### DIFF
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -234,7 +234,6 @@ export function makeCloudFunction<EventData>({
   let cloudFunction;
 
   let cloudFunctionNewSignature: any = (data: any, context: any) => {
-    console.log('inside CloudFunctionNewSignature, context is: ', context)
     if (legacyEventType && context.eventType === legacyEventType) {
       // v1beta1 event flow has different format for context, transform them to new format.
       context = {
@@ -266,7 +265,6 @@ export function makeCloudFunction<EventData>({
     before(event);
 
     let dataOrChange = dataConstructor(event);
-    console.log('right before calling the handler, event is: ', event)
     let promise = handler(dataOrChange, context);
     if (typeof promise === 'undefined') {
       console.warn('Function returned undefined, expected Promise or value');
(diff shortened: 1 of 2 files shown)


src/cloud-functions.ts
```diff
@@ -276,19 +263,44 @@ export function makeCloudFunction<EventData>({
...
-      let dataOrChange = dataConstructor(event);
-      let promise = handler(dataOrChange, context);
-      if (typeof promise === 'undefined') {
-        console.warn('Function returned undefined, expected Promise or value');
-      }
-      return await promise;
-    } finally {
-      after(event);
+    let dataOrChange = dataConstructor(event);
+    console.log('right before calling the handler, event is: ', event)
```
mbleigh: Remove

src/cloud-functions.ts
```diff
@@ -231,23 +231,10 @@ export function makeCloudFunction<EventData>({
...
-        context = _.cloneDeep(eventOrData.context);
-      } else {
-        // eventflow v1beta1 format
-        context = _.omit(eventOrData, 'data');
-      }
-    }
+  let cloudFunction;
 
+  let cloudFunctionNewSignature: any = (data: any, context: any) => {
+    console.log('inside CloudFunctionNewSignature, context is: ', context)
```
mbleigh: Remove

src/cloud-functions.ts
```diff
@@ -276,19 +263,44 @@ export function makeCloudFunction<EventData>({
...
+        after(event);
+        return result;
+      })
+      .catch(err => {
+        after(event);
+        return Promise.reject(err);
+      });
   };
+
+  if (process.env.NEW_FUNCTION_SIGNATURE) {
```
mbleigh: Let's be explicit, `process.env.NEW_FUNCTION_SIGNATURE === 'true'`

src/cloud-functions.ts
```diff
@@ -276,19 +263,44 @@ export function makeCloudFunction<EventData>({
...
+    console.log('in if loop')
```
mbleigh: Remove

src/cloud-functions.ts
```diff
@@ -276,19 +263,44 @@ export function makeCloudFunction<EventData>({
...
+    cloudFunction = cloudFunctionNewSignature;
```
mbleigh: Since this is a single line, how about we remove the `if/else` in favor of:

```js
let cloudFunction = cloudFunctionNewSignature = ...;

// wrap in a single-argument function for legacy signature
if (!process.env.NEW_FUNCTION_SIGNATURE === 'true') {
  cloudFunction = (raw) => {
    ...
    return cloudFunctionNewSignature(data, context);
  }
}
```

Issue #297, opened 2018-08-01
Support failurePolicy - retry


hobochild:
Currently, you can't pass failure policy to firebasefunctions which means you cant deploy retryable functions with `firebase-tools`. 

Not sure I'll have time to submit a PR but my two-cents on the API would be. 

 ```
exports.webhookAsia = functions
    .region('asia-northeast1')
    .failurePolicy({ retry: true }) // set retry on failure to true
    .user.onCreate((req, res) => {
            res.send("Hello");
    });
```

NOTE the gcloud functions api expects the failure policy to be formed - assume this is to future proof retry options.
```
{ failurePolicy: { retry: {}}}
```

**Comment by hobochild** [comment]
Alternatively, you could and a cli flag to the firebase-tools which patches the eventTrigger before posting it. 

like 
```
firebase deploy ----retry
```

laurenzlong:
Thanks for filing! We've discussed this a lot internally and ultimately decided to not expose retry policies via the SDK for now, since the current retry policy retries your function repeatedly for days, which can lead to unexpectedly large bills if the error is due to a programming error. If you know what you are doing, and are ok with the consequences of enabling retries you can do so through the Google Cloud Console (console.cloud.google.com). The Google Cloud Functions team is also working on more customizable retry policies, and when that is launched, we will expose the options through the firebase-functions SDK.

0x80:
@laurenzlong Any update on the "more customizable retry policies"?  It looks we can still only enable retries via cloud function runtimeOptions), without specifying how many / for how long, so using this setting is still quite risky.

PR #370, opened 2019-01-03
adds retry option to event triggered functions

joehan:
### Description
Adds retry option for event triggered functions, to match gcloud functions. https://cloud.google.com/functions/docs/reference/rpc/google.cloud.functions.v1#google.cloud.functions.v1.EventTrigger

### Code sample
//adds an onCreate function for auth, with retry enabled
functions.runWith({retry:true})
      .auth.user()
      .onCreate(user => user);

### DIFF
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -168,6 +168,7 @@ export interface TriggerAnnotated {
     httpsTrigger?: {};
     eventTrigger?: {
       eventType: string;
+      failurePolicy?: object;
       resource: string;
       service: string;
     };
@@ -303,7 +304,9 @@ export function makeCloudFunction<EventData>({
           service,
         },
       });
-
+      if (opts.retry) {
+        trigger.eventTrigger.failurePolicy = { retry: {} };
+      }
       return trigger;
     },
   });
--- a/spec/cloud-functions.spec.ts
+++ b/spec/cloud-functions.spec.ts
@@ -58,6 +58,43 @@ describe('makeCloudFunction', () => {
     });
   });
 
+  it('should put a __trigger that includes failurePolicy on the returned CloudFunction if opts.retry = true', () => {
+    let cf = makeCloudFunction({
+      provider: 'mock.provider',
+      eventType: 'mock.event',
+      service: 'service',
+      triggerResource: () => 'resource',
+      handler: () => null,
+      opts: { retry: true },
+    });
+    expect(cf.__trigger).to.deep.equal({
+      eventTrigger: {
+        eventType: 'mock.provider.mock.event',
+        resource: 'resource',
+        service: 'service',
+        failurePolicy: { retry: {} },
+      },
+    });
+  });
+
+  it('should put a __trigger that does not include failurePolicy on the returned CloudFunction if opts.retry = false', () => {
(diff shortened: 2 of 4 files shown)


src/cloud-functions.ts
```diff
@@ -168,6 +168,7 @@ export interface TriggerAnnotated {
     httpsTrigger?: {};
     eventTrigger?: {
       eventType: string;
+      failurePolicy?: object;
```
thechenky: Curious as to why failurePolicy is inside the eventTrigger and not outside like the other options like timeout and memory? was that because it does not apply to http functions?

src/cloud-functions.ts
```diff
@@ -303,7 +304,9 @@ export function makeCloudFunction<EventData>({
           service,
         },
       });
-
+      if (opts.retry) {
+        trigger.eventTrigger.failurePolicy = { retry: {} };
```
thechenky: Should this propagate the retry boolean instead of creating an empty object?

joehan:
Hmm, just saw this old issue that was closed, with reasoning for NOT implementing this: https://github.com/firebase/firebase-functions/issues/297 - thoughts?

src/cloud-functions.ts
joehan: Nope - strangely enough, google cloud expects an empty object of type Retry here: https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions#EventTrigger

src/cloud-functions.ts
joehan: This is per the Google cloud docs - since it only applies to event-triggered functions, its a property of EventTrigger: https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions#EventTrigger

joehan:
To your comment about propagating - I believe the CLI already picks up the eventTrigger and sends the whole thing, so it shouldn't need any changes

src/cloud-functions.ts
thechenky: ah, makes sense

src/cloud-functions.ts
thechenky: noted, interesting

thechenky:
Makes sense, thanks for explaining! Regarding whether we should implement this, I think @laurenzlong would have the context here. Lauren - have we set on the decision to not implement retries or have things changed since that time?

thechenky:
Chatted with Lauren offline - because this option will [retry your function for 7 days](https://cloud.google.com/functions/docs/reference/rpc/google.cloud.functions.v1#google.cloud.functions.v1.FailurePolicy), it's rather dangerous to expose as a native functionality in firebase-functions, as even the slightest error in your functions code can seriously rack up your bill. We will still continue exposing the ability to add a retry through the GCP console after a function has been deployed.

joehan:
Closing since Firebase has chosen not to expose this potentially dangerous functionality - users who need this option can access it through GCF

merlinnot:
Is there any way it could be still exposed, but would require an explicit acknowledgement of a risk, either as a confirmation question in interactive shells or a flag for CI/CD systems?

joehan:
@merlinnot : @laurenzlong - did this option come up in the discussion? Thoughts?

PR #371, opened 2019-01-04
adding errors for invalid region and timeoutSeconds values

joehan:
### Description
Adds errors when invalid values are passed to functions.runWith or functions.region

Internal bug reference: 115062564


thechenky:
oops I commented on the wrong PR there with that regex comment, please ignore. This is approved.

PR #378, opened 2019-01-10
fixes error during tsc

joehan:
### Description
In master, when you run tsc, there is an error because field, which infers to type string | number | symbol, isn't a valid for a param of type string.

Additonaly, it seems like 'fromField' is meant to represent the valid types for the key of an object, so excluding number and symbol was an oversight


joehan:
closing this and reopening separately since i forgot to change my git email on my mac and CLA bot is mad at me for it

PR #382, opened 2019-01-14
Update issue templates

thechenky:
Adding two templates for easier filing. Also adding node version to the bug report template. These templates are based loosely on https://github.com/firebase/firebase-android-sdk/issues/new/choose


.github/ISSUE_TEMPLATE/---report-a-bug.md
```diff
@@ -0,0 +1,49 @@
...
+name: "⚠️ Report a Bug"
+about: Think you found a bug in the SDK? Report it here.
+title: ''
+labels: ''
+assignees: ''
+
+---
+### Related issue
+<!-- Are there any related firebase-functions issues that you found on this topic before deciding to open a new issue? Please link them here-->
+
```
kevinajian: Some of these have an empty line after the header, others have 2 extra lines before the next header. Small nit but do we want to make the format consistent?

thechenky: ah good eye, yes.

PR #384, opened 2019-01-15
Improving stability and speed of intergration tests

joehan:
### Description

- Integration tests now wait 50 seconds after deploying to ensure that function changes propagate
- Intergration tests now retry deployment 2 times in the case of failure
- integration_tests/run_tests.sh now can be given 2 project ids, and will run node6 and node8 tests in parallel on the 2 different projects.

### Code sample

./run_tests.sh project1 project2 // runs node6 tests on project1 and node8 tests on project2, in parallel
./run_tests.sh projectId // runs node6 and node8 tests on projectId, in series

### DIFF
--- a/integration_test/run_tests.sh
+++ b/integration_test/run_tests.sh
@@ -77,7 +77,7 @@ function deploy {
   # Deploy functions, and security rules for database and Firestore
   firebase deploy --project=$PROJECT_ID --only functions,database,firestore || \
   firebase deploy --project=$PROJECT_ID --only functions,database,firestore || \
-  firebase deploy --project=$PROJECT_ID --only functions,database,firestore ## TODO: try using $() and sed to parse echo to find the printed deploy the rest text
+for i in 1 2 3; do firebase deploy --project=$PROJECT_ID --only functions,database,firestore && break; done
 }
 
 # At the moment, functions take 30-40 seconds AFTER firebase deploy returns successfully to go live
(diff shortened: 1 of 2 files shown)


integration_test/run_tests.sh
```diff
@@ -87,16 +133,32 @@ function cleanup {
...
+  install_deps
+  delete_all_functions
+  announce "Deploying functions to Node 8 runtime ..."
+  deploy
+  pick_node6
+  announce "Re-deploying the same functions to Node 6 runtime ..."
+  deploy
+  waitForPropagation
+  run_all_tests
+fi
```
thechenky: For these lines:

https://github.com/firebase/firebase-functions/blob/baa48d5a72c1ee7bf532c6b9078673beda130046/integration_test/run_tests.sh#L136-L162

Can we remove the overall if/else and replace with something like this:

```suggestion
build_sdk
pick_node8
install_deps
delete_all_functions
announce "Deploying functions to Node 8 runtime ..."
deploy
if [[ $PROJECT_ID_NODE_6 == $PROJECT_ID_NODE_8 ]]; then
  waitForPropagation
  run_tests
fi
pick_node6
announce "Re-deploying the same functions to Node 6 runtime ..."
deploy
waitForPropagation
if [[ $PROJECT_ID_NODE_6 == $PROJECT_ID_NODE_8 ]]; then
  run_tests
else
  run_all_tests
fi
```

PR #390, opened 2019-01-18
Add issue template metadata for the OSS bot

samtstern:
The OSS bot was failing to validate issue templates because it was looking for `.github/ISSUE_TEMPLATE.md` which doesn't exist.

Adding this to the top of the template lets the bot work backwards from the issue text to the original template (since the GitHub API does not preserve that information).

I will also submit a PR to the bot to be more resilient to this failure.


thechenky:
Ah, my bad, didn't realize the issue template changes would affect the bot, but of course, that makes sense now since it has to have the template to validate :)

PR #394, opened 2019-01-23
Modify gitignore

thechenky:
Update .gitignore to not track generating JS files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ node_modules
 npm-debug.log
 typings
 yarn.lock
+**/*.js.map
+**/*.js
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -13,3 +13,5 @@ node_modules
 npm-debug.log
 typings
 yarn.lock
+**/*.js.map
+**/*.js
```
laurenzlong: I'm surprised you need to do this since the lib folder is already ignored.

thechenky: Yeah, I wanted to have my IDE run unit tests automatically on any source change, and the output directory in .tsconfig was messing with that - so I got the tests set up but that resulted in a mess of generated files being put alongside the ts files. I'll take a closer look though to see what exactly caused this and see if I can make it work without modifying output dir.

thechenky:
I'm going to close this PR for now and try to modify my IDE settings so that these changes are not needed. If that's harder to do, I'll reopen.

PR #400, opened 2019-02-01
Jh expose app

joehan:
### Description
Users can now access the default app instance by calling functions.app().

### Code sample
functions.app().firestore() // equivalent to admin.app().firestore()...


changelog.txt
```diff
@@ -1 +1,2 @@
+feature - Adds functions.app() api to access the app instance used by functions
```
kevinajian: this should be `functions.apps()` right? same with your commit desc / sample

joehan: nope, its exported as app to match the api that was approved

PR #415, opened 2019-02-23
Make Event Context Generic

polRk:
### Description

Add EventContext generic type. And add supporting generic in firestore document functions

### Code sample

```
export const test = functions.firestore.document('test/{testID}').onCreate<{ testID: string }>((snapshot, context) => {
  return context.params.testID;
});
```

### DIFF
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -86,6 +86,31 @@ export interface EventContext {
   };
 }
 
+export interface EventContextGeneric<P> {
+  /** ID of the event */
+  eventId: string;
+  /** Timestamp for when the event occured (ISO string) */
+  timestamp: string;
+  /** Type of event */
+  eventType: string;
+  /** Resource that triggered the event */
+  resource: Resource;
+  /** Key-value pairs that represent the values of wildcards in a database reference */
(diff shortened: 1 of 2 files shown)


polRk:
When i work this my code on typescript, i usually use IDE, and i want use autocompletion. But i my IDE i see 'any'. And i usually make spelling mistake, use undefined props and so on.

polRk:
I want to make as much as possible Generic types. 
I see DocumentData, UpdateData and etc. use `{[field: string]: any};` , i want to make like this 
`DocumentData<T ={ [field: string]: any }>` It doesn't have any broken changes and it has backward compatibility

thechenky:
@polRk I'm really not sure why this just duplicated the original `EventContext` - I would expect a generic version of something to replace the non-generic one?

polRk:
This is only a demo version, if you do something gengeric, it entails big changes to the other types, so I made a copy

thechenky:
I don't think a demo version would be okay to keep in this repo - you're welcome to fork this and add this to your workflow, but I'm not comfortable merging in a half-baked solution. I'm going to close this out now but I've noted your needs in an internal tracker.

PR #431, opened 2019-04-19
removes extra broken call of callHttpsTrigger

joehan:
### Description
Removes an extra, broken call to callHttpsTrigger that was breakign integration tests
Screen Shot 2019-04-19 at 10 08 01 AM

### DIFF
--- a/integration_test/functions/src/index.ts
+++ b/integration_test/functions/src/index.ts
@@ -124,6 +124,7 @@ export const integrationTests: any = functions
         .collection('tests')
         .doc(testId)
         .set({ test: testId }),
+      // Invoke a callable HTTPS trigger.
       callHttpsTrigger('callableTests', { foo: 'bar', testId }, baseUrl),
       // A Remote Config update to trigger the Remote Config tests.
       // admin.credential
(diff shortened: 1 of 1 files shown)


integration_test/functions/src/index.ts
```diff
@@ -150,8 +151,6 @@ export const integrationTests: any = functions
         .storage()
         .bucket()
         .upload('/tmp/' + testId + '.txt'),
-      // Invoke a callable HTTPS trigger.
-      callHttpsTrigger('callableTests', { foo: 'bar', testId }),
```
thechenky: Was this call just not needed? Trying to understand why it was removed and not modified to include three fields.

joehan: We only need to invoke this once, which we do earlier in the file. I think this was the result of bad merge conflict resolution by git.

It looks like when the merge between public and private happened, this line was moved in one branch and modified to the new signature in the other, so git left both changes in the merge.

thechenky: Hm...that sounds odd that this happened. Also we should always run the integration tests prior to releasing to avoid this in the future.

joehan: Yeah, this one is on me - i ran them before the merge and must have forgotten to rerun afterwards

Issue #433, opened 2019-05-04
FIREBASE_CONFIG not set when Node.js 10 function deployed


Bob-Snyder:
### [REQUIRED] Environment info

**firebase-tools:** 6.8.0

**Platform:** Windows 10

### [REQUIRED] Test case
I had a set of functions built using Node.js version 6.  They were successfully deployed and running.

I took the following actions to upgrade to use a newer version of Node:
  -- Upgraded my Windows development system to Node.js version 10.15.3
  -- Upgraded to latest Firebase tools: `npm install -g firebase-tools`
  -- Upgraded Firebase SDK: `npm install firebase-functions@latest firebase-admin@latest --save`
  -- Edited the project's `package.json` file to include `"engines" : { "node" : "10" }`

With those changes, the `firebase deploy --only functions` command  completed without error.  However, immediately after deployment, an error log appeared in the Firebase console for each function: `Warning, FIREBASE_CONFIG environment variable is missing. Initializing firebase-admin will fail`.

Changing `package.json file` to specify version 8 instead of 10 resolved the problem:  `"engines" : { "node" : "8" }` .  Functions initialize and execute as expected.

### [REQUIRED] Steps to reproduce
Described in test case

### [REQUIRED] Expected behavior
Firebase tools should set FIREBASE_CONFIG to the appropriate default for the project.

### [REQUIRED] Actual behavior
FIREBASE_CONFIG is not set if Node version of 10 is specified in `package.json` file.

Bob-Snyder:
@thechenky
When the function executes, the only logs produced are these two:
 
`Function execution started`
`Function execution took 392 ms, finished with status code: 401`

The global initialization for my functions is:

```
const functions = require('firebase-functions');
const admin = require('firebase-admin');
admin.initializeApp();

const db = admin.firestore();

const settings = {timestampsInSnapshots: true};
db.settings(settings);

```

Here's one of the functions.  As I noted above, when execution fails, none of the expected log messages appear; only the exit status of 401.

```
exports.httpReportMessage2 = functions.https.onCall((reportMessage, context) => {
  // User must be authenticated.
  if (!context.auth) {
    throw new functions.https.HttpsError('failed-precondition', 'Not authenticated');
  }
  console.log('uid=', context.auth.uid);

  const timeSent = new Date(reportMessage.timeSent);

  console.log('timeSent=', timeSent.toString());

  const elapsed = Date.now() - timeSent;

  if (elapsed > 5000) {
[...]

Issue #432, opened 2019-05-02
Node 10 callable function returning unauthenticated


doowb:
### [REQUIRED] Version info

**node:**

`v10.15.2`

**firebase-functions:**
`2.3.0`

**firebase-tools:**

`6.8.0`

**firebase-admin:**
`7.3.0`

### [REQUIRED] Test case

Create a callable function:

```js
exports.hello = functions.https.onCall(async (data, context) => {
  return { message: 'Hello World' };
});
```

In client side javascript, call the function after the user has logged in:

```js
firebase.auth().onAuthStateChanged(user => {
  if (user) {
    firebase.functions()
      .httpsCallable('hello')()
      .then(result => {
        console.log(result);
      })
      .catch(err => {
        console.error(err);
      });
  }
});
```

### [REQUIRED] Steps to reproduce

Deploy the function using the `node 10` engine configured in the functions `package.json`.
Deploy the html with the client side code for authenticating and calling the function to firebase hosting.
Navigate to the page hosted on firebase and login.

### [REQUIRED] Expected behavior

See the response from the callable function output in the browser console.
This is when deployed using `node 8`:

```js
{ data: { message: 'Hello World' } }
```

### [REQUIRED] Actual behavior

`Unauthenticated` error response from calling the function. This does not happen when the function is deployed using `node 8` as the engine.

<summary>
<details>

```
Navigated to https://boilerplate-user-management.firebaseapp.com/
hello:1 POST https://us-central1-boilerplate-user-management.cloudfunctions.net/hello 401
index.js:26 Error: Unauthenticated
    at new i (error.ts:65)
    at error.ts:172
    at e.<anonymous> (service.ts:215)
    at tslib.es6.js:97
    at Object.next (tslib.es6.js:78)
    at n (tslib.es6.js:68)
(anonymous) @ index.js:26
Promise.catch (async)
(anonymous) @ index.js:25
(anonymous) @ subscribe.ts:104
(anonymous) @ subscribe.ts:225
Promise.then (async)
t.sendOne @ subscribe.ts:222
t.forEachObserver @ subscribe.ts:213
t.next @ subscribe.ts:103
n @ auth.js:1281
(anonymous) @ auth.js:1807
e.g @ promise.js:832
Ct @ promise.js:1171
Rt @ promise.js:1145
t.Yb @ promise.js:1116
dt @ run.js:132
Promise.then (async)
rt @ run.js:63
ht @ run.js:42
_t @ promise.js:1034
[...]

PR #434, opened 2019-05-07
Fix Node 10 functions not setting FIREBASE_CONFIG and functions.config()

thechenky:
### Description
Fix a bug that is occurring in Node 10 functions only where `FIREBASE_CONFIG` env var is not being set, and `functions.config()` returns `undefined`. Both of these values are coming from `.runtimeconfig.json`, which I realized is not being imported correctly in Node 10 runtime because of a changed directory structure. I verified that the Firebase CLI is indeed pulling in `.runtimeconfig.json` and uploading it properly in Node 10 so the problem is on firebase-functions SDK.

All three runtimes have different root directory defined for where the user code lives, and is represented by the environment variable `PWD` in each runtime: 
* Node 6: `/user_code`
* Node 8: `/srv`
* Node 10: `/srv/functions`

Because we were accessing `.runtimeconfig.json` with relative paths like `../../../.runtimeconfig.json`, it worked for Node 6 and 8 but the addition of an extra `functions` dir in Node 10 runtime broke it.

Fixes https://github.com/firebase/firebase-functions/issues/433 and probably https://github.com/firebase/firebase-functions/issues/432.

### Tests
* Added tests to `config.spec.ts` covering the different runtimes with different env vars set to mimic the runtimes: `process.env.PWD`
* All unit tests are passing
* Ran integration tests: `integration_test/run_tests.sh` - all passing.
* Did quite a lot of manual deploys of the same function that uses the RTDB and also sets config vars with `firebase functions:config:set key=value`. Validated that `FIREBASE_CONFIG` is set and `functions.config()` returns expected results in Node 6, 8, and 10. 

### Legacy Behavior
I removed legacy behavior from two years ago when we were anticipating `FIREBASE_PROJECT` and `CLOUD_RUNTIME_CONFIG` env vars to be set by GCF (https://github.com/firebase/firebase-functions/pull/51). This did not happen and as far as I can tell we are mostly relying on Runtime Config to set env vars now. 

@mbleigh @laurenzlong you both probably have more context here, does this sound right?

### DIFF
--- a/spec/config.spec.ts
+++ b/spec/config.spec.ts
@@ -27,7 +27,7 @@ import { config, firebaseConfig } from '../src/config';
 const NODE8_NODE10_PATH = '/srv/.runtimeconfig.json';
 const NODE6_PATH = '/user_code/.runtimeconfig.json';
 
-describe.only('config()', () => {
+describe('config()', () => {
   afterEach(() => {
     mockRequire.stopAll();
     delete config.singleton;
@@ -46,7 +46,7 @@ describe.only('config()', () => {
     );
   });
 
-  it('does not provide firebase config if .runtimeconfig.json not invalid', () => {
+  it('does not provide firebase config if .runtimeconfig.json is invalid', () => {
     mockRequire(NODE8_NODE10_PATH, 'does-not-exist');
 
     expect(firebaseConfig()).to.be.null;
@@ -69,6 +69,7 @@ describe.only('config()', () => {
         cloudResourceLocation: 'us-central',
       },
     });
+
     let loaded = firebaseConfig();
 
     expect(loaded).to.have.keys(
@@ -90,6 +91,7 @@ describe.only('config()', () => {
         cloudResourceLocation: 'us-central',
       },
     });
+
     let loaded = firebaseConfig();
 
     expect(loaded).to.have.keys(
@@ -111,6 +113,7 @@ describe.only('config()', () => {
         cloudResourceLocation: 'us-central',
       },
     });
+
     let loaded = firebaseConfig();
 
     expect(loaded).to.have.keys(
@@ -124,6 +127,7 @@ describe.only('config()', () => {
   it('loads config values from .runtimeconfig.json properly for Node 6 functions', () => {
     process.env.PWD = '/user_code';
     mockRequire(NODE6_PATH, { foo: 'bar', firebase: {} });
+
     let loaded = config();
 
     expect(loaded).to.not.have.property('firebase');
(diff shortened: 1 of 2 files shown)


spec/config.spec.ts
```diff
@@ -24,104 +24,127 @@ import * as mockRequire from 'mock-require';
...
+      foo: 'bar',
+      firebase: {
+        projectId: 'chenky-test',
+        databaseURL: 'https://chenky-test.firebaseio.com',
+        storageBucket: 'chenky-test.appspot.com',
+        cloudResourceLocation: 'us-central',
+      },
     });
-    expect(firebaseConfig()).to.have.property(
+    let loaded = firebaseConfig();
```
joehan: Could you add a new line between setting up mocks and calling the method under test, in each test, to split them into arrange, act, assert style? bkend suggested using this style on  tests in another PR, and I'm a fan

spec/config.spec.ts
```diff
@@ -24,104 +24,127 @@ import * as mockRequire from 'mock-require';
...
+      databaseURL: 'foo@firebaseio.com',
+    });
+
+    expect(firebaseConfig()).to.have.property(
+      'databaseURL',
+      'foo@firebaseio.com'
+    );
   });
 
   it('does not provide firebase config if .runtimeconfig.json not invalid', () => {
```
joehan: I might be misunderstanding this test, but I think the description either should be "... if .runtimeconfig.json is invalid", or " .... if runtimeconfig.json is valid", depending on what the expected behavior is

spec/config.spec.ts
```diff
@@ -24,104 +24,127 @@ import * as mockRequire from 'mock-require';
 import { expect } from 'chai';
 import { config, firebaseConfig } from '../src/config';
 
-describe('config()', () => {
+const NODE8_NODE10_PATH = '/srv/.runtimeconfig.json';
+const NODE6_PATH = '/user_code/.runtimeconfig.json';
+
+describe.only('config()', () => {
```
joehan: Take out this .only

thechenky: Thanks, good catch!

spec/config.spec.ts
thechenky: Hah didn't notice this was leftover from the previous code, nice catch.

spec/config.spec.ts
thechenky: Done.

thechenky:
Actually now that I think about this, not a fan of this implementation anymore since it will impact emulators, and I'm sure we won't be having `/srv` or `/user_code` defined there as root dirs. The relative path of `../../../.runtimeconfig.json` *will* work for both production and emulation environments, but it seems ugly and quite brittle as well :/ I'm torn.

thechenky:
On the other hand this looks like a bug in the functions backend with how the functions source is uploaded, possibly ignoring files that start with `.`. I've filed a bug with the internal team. Bug reference 132170650.

thechenky:
This has been fixed in the backend. Closing this out.

Issue #439, opened 2019-05-13
Update dependencies and enable automated updates


merlinnot:
I'm currently working on a feature (#425) and I'm bumping into issues because of outdated dependencies. I'd like to make a PR with dependency updates, so it doesn't decrease readability of a PR with the feature. I'll go ahead and bump all dependencies that are not breaking.

What do you think about enabling [Renovate](https://renovatebot.com) for this repository? Here's a config we could use:

```json
{
  "bumpVersion": "patch",
  "commitMessageAction": "update",
  "rangeStrategy": "bump",
  "rebaseStalePrs": true,
  "recreateClosed": true
}
```

merlinnot:
I can see that some of Firebase repositories use Repository Gardener. Let me know what's your preference and if I can help you somehow.

joehan:
I like the idea of enabling [Repository Gardener](https://github.com/GoogleCloudPlatform/repository-gardener) on this repo. The one snag I see is that we regularly sync this with a private repo, which might complicate things, but as long as we only enable it on the public side, I think it would work. @thechenky, what do you think?

thechenky:
@merlinnot can you give me an example of a Firebase repo that uses this tool?

merlinnot:
Here you go:

https://github.com/GoogleCloudPlatform/repository-gardener
https://github.com/firebase/firebase-js-sdk/pull/1599

thechenky:
Thanks @merlinnot! I talked with the devs maintaining that repo and they suggested they will likely be moving to Renovate. One issue I see with this is that Renovate is paid - so I'll have to get approval and we would have to use the hosted version. This may take some time to do, so I've filed a bug for myself to look into doing this in the future, at which point I'll loop you in. Unfortunately right now, we're pretty swamped with other higher priority work. Thank you again though for your PRs and your proactive involvement in trying to make this repo better! Really appreciate it. 

Internal bug reference: 134178323

I'm going to close out this issue in the meantime since we won't be addressing it in the near future.

PR #440, opened 2019-05-13
Update dependencies to latest versions

merlinnot:
### Description

Issue #439, partial resolution.

This PR:
- updates all dependencies to latest versions
- moves some `@types` dependencies to devDependencies
- refactors a few functions that did not compile because of improved types
- renames `typings` field to `types`, as it's the base name
- changes `node_modules/.bin/tsc` to `tsc` in all scripts in `package.json`
- orders fields in `package.json` in an alphabetical order, so it's easier to find relevant fields

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@types/cors": "^2.8.5",
     "@types/express": "^4.11.1",
     "@types/jsonwebtoken": "^8.3.2",
-    "@types/lodash": "^4.14.126",
+    "@types/lodash": "^4.14.132",
     "@types/mocha": "^5.2.6",
     "@types/mock-require": "^2.0.0",
     "@types/nock": "^10.0.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,70 +1,70 @@
...
+    "prettier": "^1.13.7",
+    "sinon": "^7.3.2",
+    "ts-node": "^8.1.0",
+    "typescript": "^3.4.0"
   },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/firebase/firebase-functions.git"
+  "engines": {
+    "node": ">=6.0.0"
```
mad-it: Support for engine 6.0.0 is being dropped right?

thechenky:
I'm not a fan of alphabetized `package.json` as it was quite difficult for me to find things like `name` which I'm used to seeing at the top. Additionally, I think it's odd to have `devDependencies` and `dependencies` not next to each other. Can we leave that file as is?

merlinnot:
I change the order of top-level properties back to original.

merlinnot:
Closing in favour of #450

PR #450, opened 2019-05-30
Lay the groundwork for v3

merlinnot:
### Description

This PR creates a foundation for a v3 release of the library, which drops support for Node.js 6. Package version is not changed deliberately, as more breaking changes might be staged in this release.

Although there's not that many changes, I made some well-named commits, so it might be easier to review one by one.

Supersedes #440


src/config.ts
```diff
@@ -40,10 +40,8 @@ export namespace config {
 
 /* @internal */
 export function firebaseConfig(): firebase.AppOptions | null {
-  // The FIREBASE_PROJECT environment variable was introduced to help local emulation with `firebase-tools` 3.18
-  // Unfortunately, API review decided that the name should be FIREBASE_CONFIG to avoid confusions that Firebase has
-  // a separate project from Google Cloud. This accepts both versions, preferring the documented name.
-  const env = process.env.FIREBASE_CONFIG || process.env.FIREBASE_PROJECT;
+  const env = process.env.FIREBASE_CONFIG;
+
```
joehan: Are we ok with getting rid of FIREBASE_PROJECT? @thechenky. If we do this, we should look over our docs to make sure its never mentioned.

tsconfig.json
```diff
@@ -1,17 +1,12 @@
 {
   "compilerOptions": {
-    "lib": ["es6"],
+    "lib": ["dom", "es2017"],
```
joehan: Some questions here:
1 - Why incldue DOM here but not in tsconfig.release.json?
2 - This ends up exlcuding a few things that es6 would include - per https://www.typescriptlang.org/docs/handbook/compiler-options.html it looks like 'es6' includes DOM.iterable and ScriptHost and this does not. If we don't use these libraries, I like excluding these since it should help reduce the file size for deployments. Are we sure these are unused?

package.json
```diff
@@ -2,69 +2,67 @@
...
+    "nock": "^10.0.6",
+    "prettier": "^1.17.1",
+    "sinon": "^7.3.2",
+    "ts-node": "^8.2.0",
+    "typescript": "^3.5.1"
   },
-  "typings": "lib/index.d.ts"
+  "peerDependencies": {
+    "firebase-admin": "^8.0.0",
+    "firebase-tools": "^5.0.0"
```
bkendall: Why was `firebase-tools` added here?

package.json
```diff
@@ -2,69 +2,67 @@
...
+  },
+  "engines": {
+    "node": "^8.13.0 || >=10.10.0"
```
bkendall: Why this specific minor version(s) of `node`? Why is anything less than `.13` and `.10` on those major versions not acceptable?

spec/fixtures/http.ts
```diff
@@ -39,15 +39,15 @@ export function mockRCVariableFetch(
...
   token: string = 'thetoken'
 ): nock.Scope {
-  let mock: nock.Scope = nock('https://runtimeconfig.googleapis.com').get(
+  let interceptor = nock('https://runtimeconfig.googleapis.com').get(
     `/v1beta1/projects/${projectId}/configs/firebase/variables/${varName}`
   );
 
   if (token) {
-    mock = mock.matchHeader('Authorization', `Bearer ${token}`);
+    interceptor = interceptor.matchHeader('Authorization', `Bearer ${token}`);
```
bkendall: well, since you're changing the name here...

`token` has a default value in the function signature. This `if` is unnecessary.

spec/fixtures/http.ts
```diff
@@ -39,15 +39,15 @@ export function mockRCVariableFetch(
   data: any,
   token: string = 'thetoken'
 ): nock.Scope {
-  let mock: nock.Scope = nock('https://runtimeconfig.googleapis.com').get(
+  let interceptor = nock('https://runtimeconfig.googleapis.com').get(
```
bkendall: Everything in this file seems to be a rename or a cleanup. Why is this necessary for `v3`?

spec/providers/https.spec.ts
```diff
@@ -213,14 +213,14 @@ const expectedResponseHeaders = {
...
-  );
   const mockedResponse = { [mocks.key_id]: mocks.public_key };
   const headers = {
     'cache-control': 'public, max-age=1, must-revalidate, no-transform',
   };
-  return mock.reply(200, mockedResponse, headers);
+
+  return nock('https://www.googleapis.com:443')
+    .get('/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com')
+    .reply(200, mockedResponse, headers);
```
bkendall: Why is this change necessary for `v3`?

tsconfig.json
```diff
@@ -1,17 +1,12 @@
...
     "module": "commonjs",
     "noImplicitAny": false,
     "outDir": ".tmp",
+    "resolveJsonModule": true,
```
bkendall: What is this necessary for?

package.json
merlinnot: My understanding is that firebase-functions is always used alongside firebase-tools. I thought it might be useful to specify which version of firebase-tools is supported.

As an example, this PR drops support for FIREBASE_PROJECT environment variable used in firebase-tools v3.18.

package.json
merlinnot: I wasn't sure if we'd ever want to use some HTTP/2 functionalities here, these versions ship with non-experimental HTTP/2 support. Firebase Admin SDK has the same limitation specified. Since it's a peer dependency and Firebase Functions meet the criteria, I thought we might as well make it a little more strict than `>=8.0.0`, just in case we ever need these functionalities. Do you see a scenario where Firebase users would like to use this library with version `<8.13.0`?

spec/fixtures/http.ts
merlinnot: I had to do a small refactoring in this file so it compiles, since an updated version of `nock` ships with more strict (more correct) type definitions.

Renames were not strictly necessary. They follow names of interfaces defined by `nock` and I tried to make it more precise and descriptive. I can extract renames into a separate PR if you feel like it's too much changes to comfortably review and approve.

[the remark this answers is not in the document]
merlinnot: See https://github.com/firebase/firebase-functions/pull/450/files#r290017363

As for the annotation, it's no longer necessary as the inferred type is now correct.

src/config.ts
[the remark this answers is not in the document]
merlinnot: Done in 3dc01290442b7c191dd3c913e0817904624252ca

spec/fixtures/http.ts
merlinnot: Nice catch, done in 76ce309dd2c2dc1b03f3f6bdd0c2dbcc7c5f2640

spec/providers/https.spec.ts
merlinnot: See https://github.com/firebase/firebase-functions/pull/450/files#r290017363 for reasoning on why I had to touch this in the first place. Since I had to make some changes anyway, I thought it might not be a bad idea to do some simple refactoring.

tsconfig.json
merlinnot: > 1 - Why incldue DOM here but not in tsconfig.release.json?

I had to add it as `protobufjs` shipped without these definitions, see https://travis-ci.org/firebase/firebase-functions/jobs/536773800. I just tested it with the newest version and it's fixed, so I removed it: 8fd948b6828865a1d88e5a3b40ba0db01a8ccff3

> 2 - This ends up exlcuding a few things that es6 would include - per https://www.typescriptlang.org/docs/handbook/compiler-options.html it looks like 'es6' includes DOM.iterable and ScriptHost and this does not. If we don't use these libraries, I like excluding these since it should help reduce the file size for deployments. Are we sure these are unused?

This target was repeatedly specified by core TypeScript maintainers as the correct one for Node.js 8, there's also a PR open by one of them: https://github.com/microsoft/TypeScript-wiki/pull/182

`lib` specifies available functions and properties, it does not affect transpilation in any way, so if it compiles (and the CI is green) we are good.

tsconfig.json
merlinnot: There's an import of a JSON file in one of the test files (`spec/providers/https.spec.ts`). This configuration is not used for production build.

merlinnot:
Thanks for the reviews @bkendall, @joehan, @thechenky! I replied to all of the comments and made some adjustments, I linked commits in some of the replies, so it's easier to review.

package.json
thechenky: Im torn here - I like the helpful warnings that tell the user they are using an outdated version of the Firebase CLI that is not compatible with this version of the functions SDK. I can see this reducing support requests. On the other hand, I can see some people using this without the Firebase CLI (like deploying with gcloud for example) and that may confuse them that the CLI is a requirement. However, deploying with gcloud is also not very well supported since most functions will require the FIREBASE_CONFIG env var which is only set if you deploy through Firebase CLI. @mbleigh what are your thoughts on this?

merlinnot: One thing to note is that all `peerDependencies` are optional, so people using it with `gcloud` will not experience errors, just warnings during `npm install`. I'd guess a majority of developers uses it with `firebase-tools`?

package.json
thechenky: I'm fine with this update for parity with firebase-admin. In any case, the GCF runtime supports only 8.15.0 and 10.15.3: https://cloud.google.com/functions/docs/concepts/exec#runtimes

spec/fixtures/http.ts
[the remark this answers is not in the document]
thechenky: Hm, that link just takes me to the changed files section without selecting a particular file - which one should I be looking at?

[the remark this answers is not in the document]
merlinnot: Hm, I guess GitHub broke something :) Here's a link to the same comment in the conversation view: https://github.com/firebase/firebase-functions/pull/450#discussion_r290017363

Just in case that one doesn't work either, here's a copy-paste:

> I had to do a small refactoring in this file so it compiles, since an updated version of nock ships with more strict (more correct) type definitions.
> 
> Renames were not strictly necessary. They follow names of interfaces defined by nock and I tried to make it more precise and descriptive. I can extract renames into a separate PR if you feel like it's too much changes to comfortably review and approve.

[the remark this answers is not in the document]
thechenky: Got it, thanks! This looks fine to me.

changelog.txt
```diff
@@ -0,0 +1,2 @@
+- important - [breaking change] The Firebase Functions SDK no longer supports Node 6. Developers must use Node 8.13.0 or higher.
+- important - [breaking change] The Firebase Functions SDK no longer supports Firebase Tools SDK below major version 5.
```
thechenky: Can you add another line here to highlight the env var deprecation? Something like this:

`- important - [breaking change] FIREBASE_PROJECT environment variable is no longer supported and has been supplanted by FIREBASE_CONFIG.`

thechenky: The CLI implemented support to deploy Node 10 functions in 6.8.0. Do you think we should use that version instead? That way, they have two runtimes to choose from, instead of just having 8. Technically it's not breaking though, so should work with v5 (btw, why did you choose v5? I saw `FIREBASE_PROJECT` was used in v3 so wondering what else was there?)

package.json
```diff
@@ -2,69 +2,67 @@
...
-    "express": "^4.16.2",
-    "jsonwebtoken": "^8.2.1",
+    "express": "^4.17.1",
+    "jsonwebtoken": "^8.3.2",
     "lodash": "^4.6.1"
   },
-  "engines": {
-    "node": ">=6.0.0"
+  "devDependencies": {
+    "@types/chai": "^4.1.7",
```
thechenky: Also, thanks for moving all the @types to the right place :)

tsconfig.release.json
```diff
@@ -1,18 +1,12 @@
...
   "compilerOptions": {
     "declaration": true,
-    "lib": ["es6"],
+    "lib": ["es2017"],
     "module": "commonjs",
     "noImplicitAny": false,
     "outDir": "lib",
     "stripInternal": true,
-    "target": "es6",
-    "typeRoots": [
```
thechenky: Just noticed this - do we not need this field anymore? This seems like it was limiting us to only pick up types in `node_modules` and not deeper down the tree.

changelog.txt
merlinnot: My intention was to choose the latest major version, I must have looked at the wrong `package.json`. I most definitely don't mind bumping it, but should we go for `^6.0.0` or `^6.8.0`? The fact that it's _supported_ does not imply that it is _required_. WDYT?

merlinnot: Added in 682e8c8f05619e8c0a8c9dbbce80f7699111c19d

tsconfig.release.json
merlinnot: It is indeed more strict, although it pulls types _more shallow_ in the tree, not _deeper_. So if there would be an enclosing directory that also have `node_modules`, then `@types` from this directory would be pulled too. I'm all for strictness, therefore I added it back in d9d9fcb48bcd7e8fef1ecd87a01384e332d7dc97.

merlinnot: Thanks for pointing this out, never thought about it like this.

changelog.txt
thechenky: Let's do `^6.8.0` since that has the functionality to deploy to Node 10, and generally brings us up to par with current node versions. I think it's better to offer this with ability to deploy to 10 than having someone realize later that they cannot deploy to 10 with the version they're on.

tsconfig.json
thechenky: I'm fine with `es2017` which I see you've changed this to. Resolving.

changelog.txt
thechenky: Let's do ^6.8.0 since that has the functionality to deploy to Node 10, and generally brings us up to par with current node versions. I think it's better to offer this with ability to deploy to 10 than having someone realize later that they cannot deploy to 10 with the version they're on.

tsconfig.release.json
thechenky: This looks good, thanks for adding back in.

tsconfig.json
thechenky: Ah is it this one: `import * as mocks from '../fixtures/credential/key.json';`? I'm all for adding better type support for imported jsons, so I don't have problems adding this in.

package.json
thechenky: Discussed this with the team - let's leave out `firebase-tools` peer dependency since the functions SDK really does not depend on it like it does on `firebase-admin` and it may cause more trouble down the line.

thechenky:
@merlinnot just those two changes in `changelog.txt` and `package.json` and we should be good to merge this in!

merlinnot:
Changelog entry added yesterday in 682e8c8, bump of `firebase-tools` today in 4abe9c0.

merlinnot:
Let me know if you need any help on other tasks towards v3.

changelog.txt
thechenky: bump on this

package.json
```diff
@@ -2,69 +2,67 @@
...
+    "nock": "^10.0.6",
+    "prettier": "^1.17.1",
+    "sinon": "^7.3.2",
+    "ts-node": "^8.2.0",
+    "typescript": "^3.5.1"
   },
-  "typings": "lib/index.d.ts"
+  "peerDependencies": {
+    "firebase-admin": "^8.0.0",
+    "firebase-tools": "^6.8.0"
```
thechenky: remove firebasee-tools?

thechenky:
Also @merlinnot FYI we are going to try to release this today as part of a suite of changes with other products dropping support.

changelog.txt
merlinnot: Done in 31b2d9e0ec65533fab0e156fa3d56d174d31aa01

package.json
merlinnot: Done in 68018f8794965985f7ed1670b4cf32e91da46754

merlinnot:
@thechenky Done!

PR #459, opened 2019-06-06
Remove Node 6 from Integration Tests

thechenky:
### Description
Removes Node 6 from integration tests, adds Node 10. Currently Node 10 is commented out since there are a couple outstanding issues with Node 10 that will need to be resolved. Additionally, update admin dependency in integration tests to `^8.0.0`.

### Testing
Tested running integration tests for Node 8. Tests are consistently passing.

### DIFF
--- a/integration_test/run_tests.sh
+++ b/integration_test/run_tests.sh
@@ -39,16 +39,16 @@ function build_sdk {
   mv firebase-functions-*.tgz integration_test/functions/firebase-functions.tgz
 }
 
-function pick_node10 {
-  cd $DIR
-  PROJECT_ID=$PROJECT_ID_NODE_10
-  cp package.node10.json functions/package.json
-}
-
 function pick_node8 {
   cd $DIR
   PROJECT_ID=$PROJECT_ID_NODE_8
   cp package.node8.json functions/package.json
+}
+
+function pick_node10 {
+  cd $DIR
+  PROJECT_ID=$PROJECT_ID_NODE_10
+  cp package.node10.json functions/package.json
 }
 
 function install_deps {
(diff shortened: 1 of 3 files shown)


integration_test/run_tests.sh
```diff
@@ -137,18 +137,18 @@ install_deps
...
-pick_node6
-announce "Re-deploying the same functions to Node 6 runtime ..."
-deploy
-waitForPropagation
-if [[ $PROJECT_ID_NODE_6 == $PROJECT_ID_NODE_8 ]]; then
-  run_tests
-else
-  run_all_tests
-fi
+# pick_node10
```
kevinajian: is there something tracking for when to uncomment this? can we add that here.

integration_test/run_tests.sh
```diff
@@ -39,10 +39,10 @@ function build_sdk {
   mv firebase-functions-*.tgz integration_test/functions/firebase-functions.tgz
 }
 
-function pick_node6 {
+function pick_node10 {
```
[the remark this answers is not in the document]
thechenky: done

integration_test/run_tests.sh
thechenky: done

integration_test/run_tests.sh
```diff
@@ -137,18 +137,19 @@ install_deps
...
-pick_node6
-announce "Re-deploying the same functions to Node 6 runtime ..."
-deploy
-waitForPropagation
-if [[ $PROJECT_ID_NODE_6 == $PROJECT_ID_NODE_8 ]]; then
-  run_tests
-else
-  run_all_tests
-fi
+# TODO(b/134418760) Uncomment this when Node 10 issues are fixed:
```
kevinajian: nit: fyi

Issue #447, opened 2019-05-24
TypeError: Cannot read property 'name' of undefined at RefBuilder.changeConstructor


jpreynat:
### Related issues

None really.

### [REQUIRED] Version info

**node:**
`v10.15.3`

**firebase-functions:**
`2.3.0`

**firebase-tools:**
`6.10.0`

**firebase-admin:**
`7.3.0`

### [REQUIRED] Test case

Deploy a database trigger on Node 10 runtime and make it run by updating the database.

### [REQUIRED] Steps to reproduce

Same as above.

### [REQUIRED] Expected behavior

I just tested upgrading our functions from Node 8 to Node 10. I expected the transition to go smoothly.

### [REQUIRED] Actual behavior

Everything is working fine locally with `firebase-tools@6.10.0` and `firebase-functions-test@0.16.0` (which works like a charm ATM, and is way faster than the previous emulator).
Deploying is working great too.
But the functions crash with the following logs:
```
Function execution took 8021 ms, finished with status: 'error'
```
```
TypeError: Cannot read property 'name' of undefined at RefBuilder.changeConstructor (/srv/functions/node_modules/firebase-functions/lib/providers/database.js:117:85)
  at cloudFunctionNewSignature (/srv/functions/node_modules/firebase-functions/lib/cloud-functions.js:119:34)
  at cloudFunction (/srv/functions/node_modules/firebase-functions/lib/cloud-functions.js:151:20)
  at Promise.resolve.then (/srv/node_modules/@google-cloud/functions-framework/build/src/invoker.js:330:28)
  at process._tickCallback (internal/process/next_tick.js:68:7)
```

### Were you able to successfully deploy your functions?

Some of our ~60 functions failed to deploy, but not more than the usual number, and not the ones that were tested.

thechenky:
Hi @jpreynat thanks for filing! This does look like an issue on our end - could you provide a quick sample of your function code? Which specific trigger are you using - `onUpdate`, `onWrite`, or `onCreate`?

thechenky:
Confirming I can reproduce this with database triggered function with Node 10 runtime. Looking into this now.

thechenky:
[...]

Issue #456, opened 2019-06-03
Storage onFinalize trigger not working in Node 10 runtime


loicmarie:
### Related issues

No related issue

### [REQUIRED] Version info

**node:** 10.15.3

**firebase-functions:** ^2.3.0

**firebase-tools:** 6.10.0

**firebase-admin:** ~7.0.0

### [REQUIRED] Test case

Following the [official example](https://firebase.google.com/docs/functions/gcp-storage-events#trigger_a_function_on_changes) :
```
exports.generateThumbnail = functions.storage.object().onFinalize(async (object) => {
  console.log(object);
});
```

### [REQUIRED] Steps to reproduce

When setting node runtime in `package.json` in `functions` directory :
```
{
  ...
  "engines": {
    "node": "10"
  }
}
```
### [REQUIRED] Expected behavior

Should output `ObjectMetadata` content in function logs, whatever the node runtime is.

### [REQUIRED] Actual behavior

Output is 'undefined' with node 10 runtime, **but is working fine with node 8 runtime** (can see the `ObjectMetadata` properties outputted)

### Were you able to successfully deploy your functions?

`firebase deploy` runs successfully, but `object` is undefined.

thechenky:
Hi @loicmarie this is likely the same issue as https://github.com/firebase/firebase-functions/issues/447. Can you post the error you're seeing when your function runs?

loicmarie:
Hi @thechenky, I have no error, since I am not doing anything ATM, but the problem is that object variable should not be undefined, it should be an instance of ObjectMetadata.
It works with Node 8 and Node 10 with local emulator, but not online with Node 10.

thechenky:
Hm, this looks like it may be caused by a missing `FIREBASE_CONFIG` env var. We had some issues on the backend that were fixed recently, but the changes would require a redeploy to take effect. Can you redeploy and see if this is still occurring?

If it is still occurring even after redeploy, check whether the `FIREBASE_CONFIG` var is set by printing the following and pasting in your response:

```
console.log(process.env.FIREBASE_CONFIG)
```

This env var will not be set if you're deploying with another tool, so please use the Firebase CLI if you're not already.

loicmarie:
It is still occuring after redeploy, and the outputted `FIREBASE_CONFIG` variable seems correct to me in console logs. For this cloud function with Node 10 :
```
export const generateThumbs = functions.storage
  .object()
  .onFinalize(async (object) => {
[...]

PR #460, opened 2019-06-10
Remove LegacyEvent, change all tests to use new Event format

thechenky:
### Description
As part of dropping Node 6 support, we can drop support for `LegacyEvent`, which was the event format used in the Node 6 runtime in Cloud Functions.

`LegacyEvent` refers to the old function signature used in Node 6 runtime where the event comes in as a [single param](https://cloud.google.com/functions/docs/writing/background#functions_background_parameters-node6). `Event` is the new function signature used in Node 8 and up where the function accepts [two parameters](https://cloud.google.com/functions/docs/writing/background#functions_background_parameters-node8-10): `data, context`. Now, `LegacyEventType` refers to `v1beta1` format and is unrelated to `LegacyEvent`. We still need to support `v1beta1` formats while removing Node 6 event format.

Additionally, this PR fixes https://github.com/firebase/firebase-functions/issues/447 since I've now removed the fork in the code that looks for an env var that is only set in Node 8.

### Testing
Integration tests are passing and they test all our providers. Providers are split between using `v1beta1` format and `v1beta2` format with most still on `v1beta1`.

#### `v1beta1`
* Auth
* Analytics
* Crashlytics
* Database (even though they have entries in both, they're still on v1beta1)
* Firestore

#### `v1beta2`
* Pubsub
* RemoteConfig
* Storage

### DIFF
--- a/spec/cloud-functions.spec.ts
+++ b/spec/cloud-functions.spec.ts
@@ -69,31 +69,6 @@ describe('makeCloudFunction', () => {
     });
   });
 
-  it('should construct the right context for legacy event format', () => {
-    let args: any = _.assign({}, cloudFunctionArgs, {
-      handler: (data: any, context: EventContext) => context,
-    });
-    let cf = makeCloudFunction(args);
-    let test: LegacyEvent = {
-      eventId: '00000',
-      timestamp: '2016-11-04T21:29:03.496Z',
-      eventType: 'providers/provider/eventTypes/event',
-      resource: 'resource',
-      data: 'data',
-    };
-
-    return expect(cf(test)).to.eventually.deep.equal({
-      eventId: '00000',
-      timestamp: '2016-11-04T21:29:03.496Z',
-      eventType: 'mock.provider.mock.event',
-      resource: {
-        service: 'service',
-        name: 'resource',
-      },
-      params: {},
-    });
-  });
-
   it('should construct the right context for new event format', () => {
     let args: any = _.assign({}, cloudFunctionArgs, {
       handler: (data: any, context: EventContext) => context,
@@ -112,7 +87,7 @@ describe('makeCloudFunction', () => {
       data: 'data',
     };
 
-    return expect(cf(test)).to.eventually.deep.equal({
+    return expect(cf(test.data, test.context)).to.eventually.deep.equal({
       eventId: '00000',
       timestamp: '2016-11-04T21:29:03.496Z',
       eventType: 'provider.event',
@@ -178,7 +153,7 @@ describe('makeCloudFunction', () => {
       data: 'test data',
     };
 
-    return cf(test).then(result => {
+    return cf(test.data, test.context).then(result => {
       expect(result).to.deep.equal({
         eventId: '00000',
         timestamp: '2016-11-04T21:29:03.496Z',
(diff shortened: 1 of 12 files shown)


thechenky:
@kevinajian apologies for the large PR - most of these changes are just updating the unit tests to use the new Event format where both `data` and `context` are provided. Seems like these tests were not changed for quite a while.

spec/cloud-functions.spec.ts
```diff
@@ -204,19 +178,6 @@ describe('makeParams', () => {
...
-      data: 'data',
-    };
-
-    return expect(cf(testEvent)).to.eventually.deep.equal({
-      foo: 'a',
-      bar: 'b',
-    });
-  });
-
   it('should construct params from the event resource of new format events', () => {
```
kevinajian: should this just be updated to remove `new format` since there's no legacy anymore?

spec/providers/analytics.spec.ts
```diff
@@ -260,23 +311,35 @@ describe('Analytics Functions', () => {
...
         );
 
         // The event data delivered over the wire will be the JSON for an AnalyticsEvent:
         // https://firebase.google.com/docs/auth/admin/manage-users#retrieve_user_data
-        let event: LegacyEvent = {
-          eventId: 'f2e2f0bf-2e47-4d92-b009-e7a375ecbd3e',
-          eventType: 'providers/google.firebase.analytics/eventTypes/event.log',
-          resource: 'projects/myUnitTestProject/events/first_open',
+        let event: Event = {
+          // eventId: 'f2e2f0bf-2e47-4d92-b009-e7a375ecbd3e',
```
kevinajian: are these comments needed?

spec/providers/analytics.spec.ts
```diff
@@ -69,22 +69,33 @@ describe('Analytics Functions', () => {
...
+            (data: analytics.AnalyticsEvent, context: EventContext) => data
+          );
 
         // The event data delivered over the wire will be the JSON for an AnalyticsEvent:
         // https://firebase.google.com/docs/auth/admin/manage-users#retrieve_user_data
-        let event: LegacyEvent = {
-          eventId: 'f2e2f0bf-2e47-4d92-b009-e7a375ecbd3e',
-          eventType: 'providers/google.firebase.analytics/eventTypes/event.log',
-          resource: 'projects/myUnitTestProject/events/first_open',
+        let event: Event = {
```
kevinajian: don't need to handle it in this pr but can these lets be consts? In other tests as well.

src/cloud-functions.ts
```diff
@@ -314,23 +301,7 @@ export function makeCloudFunction<EventData>({
...
-        // new eventflow v1beta2 format
-        context = _.cloneDeep(raw.context);
-      } else {
-        // eventflow v1beta1 format
-        context = _.omit(raw, 'data');
-      }
-      return cloudFunctionNewSignature(data, context);
-    };
-  }
+  cloudFunction = cloudFunctionNewSignature;
```
kevinajian: can remove the let cloudFunction above and do const here, or just use cloudFunctionNewSignature below

thechenky: Good catch - changed to use cloudFunction everywhere.

spec/providers/analytics.spec.ts
thechenky: Ah good catch, removed.

spec/providers/analytics.spec.ts
thechenky: Yeah, good idea. I'll do it in a separate PR.

thechenky:
Made suggested changed, going to do some more testing before merging. Thanks!

liezl200:
Setting `process.env.X_GOOGLE_NEW_FUNCTION_SIGNATURE = 'true'; `in my index.js is not working. I also tried `process.env.X_GOOGLE_NEW_FUNCTION_SIGNATURE = true;`
  I still get "TypeError: functions.storage.bucket(...).onFinalize is not a function"

https://github.com/firebase/firebase-functions/issues/447
https://github.com/firebase/firebase-functions/issues/456

Issue #437, opened 2019-05-10
GCLOUD_PROJECT environment variable missing from the Node 10 runtime


peterpeterparker:
### [REQUIRED] Version info

**node:**
Node v10

### [REQUIRED] Steps to reproduce

I have upgraded my functions from Node v8 to Node v10. Compilation as upload in the cloud are ok but I face an error at runtime, respectively `GCLOUD_PROJECT` is not found/set

Here the stack trace of the error I found in the Firebase console:

```
Error: process.env.GCLOUD_PROJECT is not set.
    at DocumentBuilder (/srv/functions/node_modules/firebase-functions/lib/providers/firestore.js:99:23)
    at cloudFunctionNewSignature (/srv/functions/node_modules/firebase-functions/lib/cloud-functions.js:102:13)
    at cloudFunction (/srv/functions/node_modules/firebase-functions/lib/cloud-functions.js:151:20)
    at Promise.resolve.then (/srv/node_modules/@google-cloud/functions-framework/build/src/invoker.js:330:28)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

When I downgrade back to Node v8, I do not face that error.

### Were you able to successfully deploy your functions?

Yes

thechenky:
Hi @peterpeterparker thanks for opening up the new issue. Can you provide a snippet of your functions code?

peterpeterparker:
sure @thechenky here you go:

```
import {CollectionReference, DocumentSnapshot, QuerySnapshot, DocumentReference} from '@google-cloud/firestore';

import * as admin from 'firebase-admin';

import {Change, EventContext} from 'firebase-functions';

export async function applyUsersUpdate(change: Change<DocumentSnapshot>, context: EventContext) {
    const newValue: any = change.after.data() as any;

    const previousValue: any = change.before.data() as any;

    if (!newValue || !previousValue || newValue.my_field === previousValue.my_field) {
        return;
    }

    try {
        await myUpdateMethod(change, newValue, CirclesStateField.REQUEST);
    } catch (err) {
        console.error(err);
    }
}
```

thechenky:
Hm, I would expect to see something calling Firestore to set the value (since the error is coming from here: firebase-functions/lib/providers/firestore.js:99:23). Is that in the `myUpdateMethod`? Would you mind pasting that too?

thechenky:
[...]

PR #462, opened 2019-06-11
Set GCLOUD_PROJECT from FIREBASE_CONFIG when missing

kevinajian:
### Description

Addresses https://github.com/firebase/firebase-functions/issues/437

Set GCLOUD_PROJECT from FIREBASE_CONFIG when not set.

### DIFF
--- a/spec/index.spec.ts
+++ b/spec/index.spec.ts
@@ -28,6 +28,7 @@ import * as nock from 'nock';
 nock.disableNetConnect();
 
 import 'mocha';
+import { setup } from '../src/index';
 import './utils.spec';
 import './apps.spec';
 import './cloud-functions.spec';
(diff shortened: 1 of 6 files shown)


spec/index.spec.ts
```diff
@@ -43,3 +44,28 @@ import './providers/pubsub.spec';
 import './providers/remoteConfig.spec';
 import './providers/storage.spec';
 import './providers/crashlytics.spec';
+
```
thechenky: I think this file is for consolidating all the test files so that they run together when you do `npm run test`. Can we have a separate file for testing setup? That way we don't need all those imports at the top which at first confused me.

spec/index.spec.ts
```diff
@@ -43,3 +44,28 @@ import './providers/pubsub.spec';
...
+describe('setup()', () => {
+  afterEach(() => {
+    delete process.env.FIREBASE_CONFIG;
+    delete process.env.CLOUD_RUNTIME_CONFIG;
```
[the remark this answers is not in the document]
kevinajian: changed cloud runtime to gcloud project

src/index.ts
```diff
@@ -51,34 +51,50 @@ export {
...
+          `${process.env.DATABASE_URL}` ||
+          `https://${process.env.GCLOUD_PROJECT}.firebaseio.com`,
+        storageBucket:
+          `${process.env.STORAGE_BUCKET_URL}` ||
+          `${process.env.GCLOUD_PROJECT}.appspot.com`,
+        projectId: process.env.GCLOUD_PROJECT,
+      });
+    } else {
+      console.warn(
+        'Warning, FIREBASE_CONFIG environment variable is missing. Initializing firebase-admin will fail'
```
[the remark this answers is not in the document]
kevinajian: updated messaging

spec/index.spec.ts
kevinajian: moved to `setup.spec`

spec/index.spec.ts
```diff
@@ -28,11 +28,11 @@ import * as nock from 'nock';
 nock.disableNetConnect();
 
 import 'mocha';
-import { setup } from '../src/index';
 import './utils.spec';
 import './apps.spec';
 import './cloud-functions.spec';
 import './config.spec';
+import './setup.spec';
```
thechenky: I think since the test is in `setup.spec` it would be good to move the code to be there too instead of being a function in`index.ts` so that everything is consistent and the reader knows where to find the test code for the source file. What do you think?

kevinajian: ok made a `setup.ts`

PR #467, opened 2019-06-13
Adds npm lint command and updates lint options

thechenky:
### Description
**This PR does not add new functionality**

These changes to the linter are setting up for adding the lint command into our tests just like in the CLI - the options are consistent with what we have in [firebase-tools](https://github.com/firebase/firebase-tools/blob/master/tslint.json). But before we can add this into our test flow we need to fix all the linter issues that currently being thrown, which I will be doing in following PRs. 

### Code sample
`npm run lint` will now print the warnings (not error yet, we can put that in later). There are a lot of them right now.

### DIFF
--- a/integration_test/functions/src/index.ts
+++ b/integration_test/functions/src/index.ts
@@ -52,7 +52,9 @@ function callScheduleTrigger(functionName: string, region: string) {
       {
         method: 'POST',
         host: 'cloudscheduler.googleapis.com',
-        path: `projects/${firebaseConfig.projectId}/locations/us-central1/jobs/firebase-schedule-${functionName}-${region}:run`,
+        path: `projects/${
+          firebaseConfig.projectId
+        }/locations/us-central1/jobs/firebase-schedule-${functionName}-${region}:run`,
         headers: {
           'Content-Type': 'application/json',
         },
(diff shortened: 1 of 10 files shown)


integration_test/functions/src/index.ts
```diff
@@ -52,7 +52,9 @@ function callScheduleTrigger(functionName: string, region: string) {
       {
         method: 'POST',
         host: 'cloudscheduler.googleapis.com',
-        path: `projects/${firebaseConfig.projectId}/locations/us-central1/jobs/firebase-schedule-${functionName}-${region}:run`,
+        path: `projects/${
```
thechenky: FYI, this file only contains prettier changes but I decided to bundle this into this PR since they're small.

package.json
```diff
@@ -56,7 +57,10 @@
     "prettier": "^1.17.1",
     "sinon": "^7.3.2",
     "ts-node": "^8.2.0",
-    "typescript": "^3.5.1"
+    "typescript": "^3.5.1",
+    "tslint": "^5.16.0",
+    "tslint-no-unused-expression-chai": "^0.1.4",
+    "tslint-plugin-prettier": "^2.0.0"
```
bkendall: Why are these out of order? Were they added with `npm install --save-dev`?

thechenky: They were not, I just added them manually and slipped on the alphabetizing, will fix :)

thechenky: Fixed.

PR #469, opened 2019-06-13
Fix lint errors for auth, analytics, crashlytics, remoteconfig tests

thechenky:
### Description
Continuation of lint fixes now in `spec/providers` directory.

### DIFF
--- a/spec/providers/analytics.spec.ts
+++ b/spec/providers/analytics.spec.ts
@@ -20,11 +20,12 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-import * as analytics from '../../src/providers/analytics';
 import { expect } from 'chai';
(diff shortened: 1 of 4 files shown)


spec/providers/analytics.spec.ts
```diff
@@ -289,10 +291,11 @@ describe('Analytics Functions', () => {
       it('should recognize all the fields the payload can contain', () => {
         const cloudFunction = analytics
           .event('first_open')
-          .onLog((data: analytics.AnalyticsEvent) => data);
+          .onLog((event: analytics.AnalyticsEvent) => event);
```
kevinajian: should the other ones also be changed to event?

thechenky: Good point. I was renaming this specific instance because of a shadowed variable down lower that is also called data. Event providers should all take in `(data, context)`, here it looks we're just cutting some corners by not passing two arguments, which I think makes things more confusing. Since this is this is the first argument I'm actually reconsidering naming this data and changing the other var to something else.

PR #470, opened 2019-06-13
Fix lint errors for database, firestore, https tests

thechenky:
### Description
Continuation of the lint fixes. This PR fixes lint errors for database, firestore, https tests.

### DIFF
--- a/spec/providers/firestore.spec.ts
+++ b/spec/providers/firestore.spec.ts
@@ -20,17 +20,17 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-import * as admin from 'firebase-admin';
-
-import * as firestore from '../../src/providers/firestore';
-import * as _ from 'lodash';
 import { expect } from 'chai';
+import * as admin from 'firebase-admin';
+import * as _ from 'lodash';
+
 import * as functions from '../../src/index';
+import * as firestore from '../../src/providers/firestore';
 
 describe('Firestore Functions', () => {
   function constructValue(fields: any) {
     return {
-      fields: fields,
+      fields,
       name: 'projects/pid/databases/(default)/documents/collection/123',
       createTime: '2017-06-02T18:48:58.920638Z',
       updateTime: '2017-07-02T18:48:58.920638Z',
@@ -40,7 +40,7 @@ describe('Firestore Functions', () => {
   function makeEvent(data: any, context?: { [key: string]: any }) {
     context = context || {};
     return {
-      data: data,
+      data,
       context: _.merge(
         {
           eventId: '123',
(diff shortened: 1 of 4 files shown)


spec/providers/firestore.spec.ts
```diff
@@ -252,19 +254,19 @@ describe('Firestore Functions', () => {
...
+      const testFunction = firestore
         .document('path')
         .onCreate((data, context) => {
           expect(data.data()).to.deep.equal({ key1: true, key2: 123 });
           expect(data.get('key1')).to.equal(true);
           return true; // otherwise will get warning about returning undefined
         });
-      let data = constructEvent({}, createValue(), 'document.create');
-      return testFunction(data.data, data.context);
+      const event = constructEvent({}, createValue(), 'document.create');
```
kevinajian: if we change this to event should we change the other ones as well?

spec/providers/https.spec.ts
```diff
@@ -164,6 +164,7 @@ async function runTest(test: CallTest): Promise<any> {
 }
 
 // MockRequest mocks an https.Request.
+// tslint:disable-next-line
```
kevinajian: what's the reason for this?

spec/providers/https.spec.ts
```diff
@@ -394,10 +395,10 @@ describe('callable.FunctionBuilder', () => {
 
     it('should handle unhandled error', () => {
       return runTest({
-        httpRequest: request(null),
+        httpRequest: mockRequest(null),
         expectedData: null,
         callableFunction: (data, context) => {
-          throw 'ceci n\'est pas une error';
+          throw new Error("ceci n'est pas une error");
```
kevinajian: can we use a template string instead?

spec/providers/https.spec.ts
```diff
@@ -559,20 +560,22 @@ describe('callable.FunctionBuilder', () => {
 describe('callable CORS', () => {
   it('handles OPTIONS preflight', async () => {
     const func = https.onCall((data, context) => {
-      throw "This shouldn't have gotten called for an OPTIONS preflight.";
+      throw new Error(
+        "This shouldn't have gotten called for an OPTIONS preflight."
```
kevinajian: same here, template string.

spec/providers/firestore.spec.ts
thechenky: Changing it back to data here as well.

spec/providers/https.spec.ts
thechenky: Sure

spec/providers/https.spec.ts
thechenky: Rule that allows only one class per file - I didn't want to change this here yet, but perhaps I should leave the `tslint:disable` off so that we can fix that in the future. We are not currently running linter on tests so it won't fail.

spec/providers/firestore.spec.ts
thechenky: Actually, this makes sense to be event, so changed the others to event instead of data.

spec/providers/https.spec.ts
thechenky: Actually I just moved this out into its own file in `fixtures` and now we don't need that disable line.

PR #491, opened 2019-06-27
Add automatic TypeDoc generation to Functions SDK

thechenky:
### Description
Make it possible to autogenerate documentation by using TypeDoc on the Functions SDK repo. This version does not depend on the `index.d.ts` file, but instead uses `/src` as the source of truth. Note that the links are currently broken - it looks like the files are getting written out as just the object name instead of the full namespace. So instead of `functions.analytics.AnalyticsEvent.html` it's writing it out as `AnalyticsEvent.html`. If this is acceptable, I can change the paths in the `toc` file to unnest. I think this is probably from the differences between admin generating their doc from an index.d.ts file that preserves the hierarchies vs us generating from src/ files. 

### Code sample
To generate docs, run `node docgen/generate-docs.js` from the root directory.


docgen/content-sources/toc.yaml
```diff
@@ -0,0 +1,75 @@
+toc:
+- title: "functions"
+  path: /docs/reference/functions/functions
+
+- title: "functions.analytics"
+  path: /docs/reference/functions/functions.analytics
+  section:
+  - title: "AnalyticsEvent"
+    path: /docs/reference/functions/AnalyticsEvent
```
thechenky: @hiranya911 this seems to work when generating from `/src` in the functions SDK, but it's not the nice hierarchy that I saw in admin. I'd love for these paths to be namespaced better but I suppose it doesn't matter since these paths just make the links work, right?

thechenky:
@egilmorez @hiranya911 this is a work in progress but just wanted to get your take on this and how far away it is from looking ready. Mind taking a look and regenerating the docs on your machine to see if they look right? There's an issue with some links being broken that I can fix.

egilmorez:
I have been pulling and trying this out as we go, and right now everything looks fine!

thechenky:
@kevinajian @egilmorez this is now ready for review. I fixed the toc checks, so should now be in final state.

PR #493, opened 2019-07-03
Run Prettier and TSLint in the CI

merlinnot:
### Description

This PR introduces two new scripts: `format:fix` and `lint:fix`, which runs Prettier and TSLint with autofixers. It also changes a behaviour of `format` to report issues. Both are now run in the CI, in parallel with tests.

I fixed all linter errors, so the CI passes. All changes in source files are autofixes made by TSLint.

### Code sample

Not relevant.

PR #492, opened 2019-07-03
Update dependencies to latest versions

merlinnot:
### Description

This PR updates all dependencies to latest versions, there are no breaking changes - some files are reformatted by Prettier.

### Code sample

Not relevant.

PR #494, opened 2019-07-03
Extract configuration to break dependency cycle

merlinnot:
### Description

This PR moves definitions related to configuration options to `functions-configuration`, as it causes an import cycle.

This PR contains some changes made automatically by Prettier and TSLint, it will be gone once #493 is merged.

### Code sample

Not relevant.


merlinnot:
@thechenky Unfortunately I don't have permissions to assign you as a reviewer, as requested in the previous PR.

thechenky:
This still has formatting and lint changes ya? Let's remove those - I approved the lint/format PR and can merge that in now.

merlinnot:
I thought you'll just merge the other PR straight away, so I can rebase this one. I have my editor set up to autofix and autoformat on save, hence these changes were applied to all of the files I touched. I reverted it manually. What's the process of merging PRs? Do we need more reviews first?

thechenky:
I merged the other two PRs, can you update with master? It should remove the extra formatting changes too.

thechenky:
Ah sorry missed your last comment - we should be good to go, I forgot I'm supposed to do the merging since you can't push that button :)

merlinnot:
Only you have the power to do that ;) I've merged master, Travis takes a long time to run today, after yesterdays incident they still have a large backlog of jobs.

thechenky:
Just merged https://github.com/firebase/firebase-functions/pull/492 also

thechenky:
Nice, this now looks like only relevant changes. Reviewing now :)

src/function-configuration.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+export const VALID_MEMORY_OPTIONS = [
+  '128MB',
+  '256MB',
+  '512MB',
+  '1GB',
+  '2GB',
+] as const;
+
+export interface ScheduleRetryConfig {
```
thechenky: Let's add a comment here, something like:

```
/**
* Scheduler retry options. Applies only to scheduled functions.
*/
```

src/function-configuration.ts
```diff
@@ -0,0 +1,63 @@
...
+  retryCount?: number;
+  maxRetryDuration?: string;
+  minBackoffDuration?: string;
+  maxBackoffDuration?: string;
+  maxDoublings?: number;
+}
+
+export interface Schedule {
```
thechenky: Same here: 

```
/**
* Configuration options for scheduled functions
*/
```

src/function-configuration.ts
```diff
@@ -0,0 +1,63 @@
...
+   * Amount of memory to allocate to the function.
+   */
+  memory?: typeof VALID_MEMORY_OPTIONS[number];
+  /**
+   * Timeout for the function in seconds, possible values are 0 to 540.
+   */
+  timeoutSeconds?: number;
+}
+
+export interface DeploymentOptions extends RuntimeOptions {
```
thechenky: Love this, thank you for making this extend `RuntimeOptions`, much cleaner now!

src/index.ts
```diff
@@ -23,16 +23,16 @@
...
-import { handler } from './handler-builder';
 import * as crashlytics from './providers/crashlytics';
 import * as database from './providers/database';
 import * as firestore from './providers/firestore';
 import * as https from './providers/https';
 import * as pubsub from './providers/pubsub';
 import * as remoteConfig from './providers/remoteConfig';
 import * as storage from './providers/storage';
+
+import * as apps from './apps';
```
thechenky: Think this formatting change slipped through?

merlinnot: I moved it on purpose. I added one export to this file. I noticed the comment above the block where I moved this line from says “providers”, and this is not a provider. I can extract it to a separate PR if you feel like it.

thechenky: Got it. Would tslint complain about this because it's now out of order?

merlinnot: No, TSLint works on blocks of imports - and we have a newline above.

src/function-configuration.ts
merlinnot: f28e5bfc5eb67b9ee9a3c2b788121ae2b7d4ef1d

src/function-configuration.ts
merlinnot: f28e5bfc5eb67b9ee9a3c2b788121ae2b7d4ef1d

src/function-configuration.ts
merlinnot: I'm happy you like it :)

merlinnot:
@thechenky When I was working on the comments, I made a small change to stricten type definitions for regions, hope that's fine -> 22b6d0b674ddcfac969f8eb337a2afcad1975897.

spec/function-builder.spec.ts
```diff
@@ -165,21 +165,21 @@ describe('FunctionBuilder', () => {
...
     }).to.throw(Error, 'region');
 
     expect(() => {
-      return functions.region('unsupported', 'us-east1');
+      return functions.region('unsupported' as any, 'us-east1');
     }).to.throw(Error, 'region');
 
     expect(() => {
-      return functions.region('unsupported', 'us-east1').runWith({
+      return functions.region('unsupported' as any, 'us-east1').runWith({
```
merlinnot: All of these can be changed to avoid `any`, I can do that if that's your preference. `any` is much simpler here.

```suggestion
      return functions
        .region(
          ('unsupported' as unknown) as Parameters<
            typeof functions.region
          >[number],
          'us-east1'
        )
        .runWith({
```

thechenky: I feel it looks too verbose with using `unknown`, let's keep using `any`. Thanks!

src/index.ts
thechenky: Ack.

thechenky:
@merlinnot hm, I just ran the integration tests on master after merging this in, and they are failing on deploying:

image

thechenky:
It works if I remove the `as const` for SUPPORTED_REGIONS and VALID_MEMORY_OPTIONS.

merlinnot:
I'll check that.

thechenky:
Unless there's something we can do to make this work with keeping `as const` I have a fix in https://github.com/firebase/firebase-functions/pull/497.

thechenky:
Actually, master integration tests being broken will block releases for GCF, so I'm going to submit the fix now and if we figure out a way to make it work with original version, we can add that in later.

PR #499, opened 2019-07-03
Support running integration tests in directories with special characters

merlinnot:
### Description

This PR makes the bash script used to run integration tests more strict, by surrounding variables with `""`, so each is treated as a string. It also makes references to variables more consistent by always surrounding them with `${}`.

### Code sample

Not relevant.


merlinnot:
@thechenky

PR #502, opened 2019-07-09
Remove unused files

merlinnot:
### Description

This PR removes an unused file, which was edited for the last time two years ago and has no meaningful content. This might have been an idea that resulted in creation of `@firebase/testing`, but I can't tell for sure. What's more, these files are actually excluded from the package (see `.npmignore`). Bottom line, it seems like it can be removed.

### Code sample

Not relevant.


merlinnot:
@thechenky

PR #504, opened 2019-07-09
Simplify TypeScript configuration

merlinnot:
### Description

Since I'd like to make some changes to TypeScript configuration, e.g. ban unused variables, I simplified the configuration so it can be changed in one place, instead of manually keeping two files in sync. The only relevant change in behaviour is that the output directory was specified as `.tmp`, but there's no script that actually produces output, so I removed it.

### Code sample

Not relevant.


tsconfig.release.json
```diff
@@ -7,7 +7,7 @@
     "outDir": "lib",
     "stripInternal": true,
     "target": "es2017",
-    "typeRoots": ["node_modules/@types"]
+    "typeRoots": ["./node_modules/@types"]
```
merlinnot: `./` are added to make all references consistent. There are fields in TS config files where paths without `./` are not allowed (such as `extends`), so adding it is the only way to achieve consistency.

merlinnot:
@thechenky

PR #505, opened 2019-07-09
Remove redundant fields from ScheduleBuilder

merlinnot:
### Description

While working on removing unused locals, I noticed redundancy in ScheduleBuilder class. I refactored it so it's simpler and avoids redundancy.

### Code sample

Not relevant.


src/providers/pubsub.ts
```diff
@@ -107,7 +102,7 @@ export function _scheduleWithOptions(
   schedule: string,
   options: DeploymentOptions
 ): ScheduleBuilder {
-  return new ScheduleBuilder({ schedule }, options);
+  return new ScheduleBuilder({ ...options, schedule: { schedule } });
```
merlinnot: Apparently `schedule` of type `object` has a property `schedule` of type `string`. Whereas it might look wrong at a first glance, that's how it's defined.

merlinnot:
@thechenky

src/providers/pubsub.ts
```diff
@@ -65,20 +64,16 @@ export function schedule(schedule: string): ScheduleBuilder {
 }
 
 export class ScheduleBuilder {
-  private _options: DeploymentOptions;
-
   /** @internal */
-  constructor(private schedule: Schedule, private options: DeploymentOptions) {
-    this._options = { schedule, ...options };
-  }
+  constructor(private options: DeploymentOptions) {}
```
thechenky: Does this @internal tag make sure that `options` are also not exposed?

src/providers/pubsub.ts
thechenky: Good catch, yeah it's very nested and we should probably just rename this instead of having schedule => Schedule => schedule (but don't worry about that for now). Also missed in the code review that DeploymentOptions already had schedule in it. You have very good attention to detail!

PR #506, opened 2019-07-09
Enable and apply `noUnusedLocals` compiler option

merlinnot:
### Description

This PR removes all unused locals from the codebase and enables `noUnusedLocals` TypeScript configuration option.

It builds on #504 and #505, so these two should be merged first.

### Code sample

Not relevant.


merlinnot:
@thechenky

thechenky:
Both #504 and #505 have been merged.

PR #507, opened 2019-07-09
Specify a type of DeploymentOptions.regions more precisely

merlinnot:
### Description

This PR provides a more correct type definition for the property.

### Code sample

Not relevant.


merlinnot:
@thechenky

PR #508, opened 2019-07-09
Change Prettier configuration to print all commas

merlinnot:
### Description

This PR is split in two commits: one with a configuration change and one with a reformat, which was performed automatically.

This changes makes PR changes smaller whenever there's a property, parameter, array entry etc. added after any existing ones.

### Code sample

Not relevant.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,5 @@
 {
   "arrowParens": "always",
   "singleQuote": true,
-  "trailingComma": "es5"
+  "trailingComma": "all"
 }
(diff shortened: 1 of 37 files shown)


merlinnot:
@thechenky

thechenky:
I'm not a fan of the trailing comma, can we remove that?

merlinnot:
Sure, it’s just a proposal. I don’t really mind either style, it’s just that with trailing commas you have cleaner diffs, as explained in the PR description. But if it’s not valuable enough for you, let’s just close this one.

PR #510, opened 2019-07-09
Move unexported dependencies on @types to devDependencies

merlinnot:
### Description

Whereas `@types/express` is actually a dependency, as it's types are re-exported, the other three are only used internally - they are therefore not included and used in built package - an easy find in `lib` can confirm that. I moved it to `devDependencies`, so it's not installed together with the package, making installation more performant - which is important especially in FaaS environments.

### Code sample

Not relevant.


merlinnot:
@thechenky

merlinnot:
Are you sure all of these are required, not only `@types/express`? It would make perfect sense, as I described in the PR description.

I just tested it locally, doing `npm pack`, initialising an empty project, linking it locally, compiling - all worked just fine.

thechenky:
Ah yes, you're right, I didn't notice that `@types/express` was still under dependencies. I think that's the only one we need. I just cloned your fork to confirm this. When I remove `@types/express` I get compile errors, same as what we saw when releasing v3.0.0, but with it back in, no errors. Going to run integration tests now.

thechenky:
Oh actually one nit - the ordering is a bit off in the package.json - `chai` should be above `chai-as-promised` and there's a couple others.

thechenky:
Integration tests passed. This is good to go in!

PR #511, opened 2019-07-09
Change @internal => @hidden

thechenky:
### Description
TypeDoc does not respect `@internal` tags and only respects `@hidden`, so changing all occurrences to `@hidden` to aid doc generation.

This complements #491.


src/providers/database.ts
```diff
@@ -87,7 +87,7 @@ export function _instanceWithOptions(
 }
 
 export class InstanceBuilder {
-  /* @internal */
+  /* @hidden */
```
kevinajian: `/** @hidden */`

src/cloud-functions.ts
```diff
@@ -149,7 +149,7 @@ export namespace Change {
   }
 
   /**
-   * @internal
+   * @hidden
```
kevinajian: should we make these single lined?

thechenky: Yep: https://github.com/firebase/firebase-functions/pull/511/commits/53c3a1f88882c9bbbf6e0ef4c769d31701ca6854

src/providers/database.ts
thechenky: Good catch, fixed in https://github.com/firebase/firebase-functions/pull/511/commits/e69011bd9b8e445720d4457d3cecf760b02a600e.

PR #512, opened 2019-07-10
Include all TypeScript files in development configuration

merlinnot:
### Description

There are two config files for TypeScript:  `tsconfig.release.json` and `tsconfig.json`. The latter is always used for local development, as some editors (like VS Code) and tools don't handle non-default locations. Since `tsconfig.json` is used by all of these tools, it's convenient to include all TypeScript files in it. Some of the files in the project were not included, namely `integration_test` directory.

This PR changes the configuration so it includes all files, causing both TypeScript and TSLint to take `integration_test` into account. The PR is split in two commits: one that changes the configuration and fixes minor compilation errors and one with the output of TSLint autofixer.

### Code sample

Not relevant.


merlinnot:
@thechenky

PR #513, opened 2019-07-10
Remove unused valAt utility

merlinnot:
### Description

This PR removes an unused utility function.

### Code sample

Not relevant.

PR #514, opened 2019-07-10
Extract and document path manipulation utilities

merlinnot:
### Description

While working on a small fix I wanted to add a utility function that could be reused across the codebase. I noticed that all utilities live in one file, which makes it too easy to throw everything in there. I started by removing an unused function in #513. This PR extracts path manipulation utilities into its own file. I created `utilities` directory, so more well-named, well-scoped files and methods can be added there.

### Code sample

Not relevant.


merlinnot:
@thechenky

PR #515, opened 2019-07-10
Introduce Mocha configuration file

merlinnot:
### Description

While adding some tests to the repository, I noticed it's required to register these in `spec/index.spec.ts`. I find that inconvenient, most often all test files are picked up automatically.

This PR introduces a Mocha configuration file (introduced in Mocha v6, supersedes now deprecated `mocha.opts`). It applies a sane configuration, with a few extras:
- Since Realtime Database is left in a "dirty" state by some tests, I added an `exit` option which forces Mocha to force the process exit after all tests are finished. I'd treat it as a temporary fix, until the root cause is fixed. It will greatly speed up the CI too. For more information, see https://mochajs.org/#-exit
- A setup file is extracted to `mocha/setup.ts`. I'm not sure if that's the most fortunate naming, but I didn't want to mix it with actual test files in the `setup` directory. I'm open to suggestions!
- It specifies extensions, so a watch mode can be easily enabled.

### Code sample

Not relevant.


merlinnot:
@thechenky

thechenky:
This is very nice, I like the exit, those database logs have been pretty annoying that keep printing after the tests finish. Can we change the config file to be json format instead of yaml? Since other files are using json, it would be good to be consistent.

.mocharc.yaml
```diff
@@ -0,0 +1,10 @@
+exit: true
```
thechenky: Let's make this json :)

merlinnot: Not all of the configuration files use JSON, e.g. Travis configuration is YAML.

Yaml has this advantage over JSON, that it supports comments, which is quite important when you want to have everything well documented. AFAIK Mocha does not support JSON5, so the most reasonable format that supports comments seems to be yaml.

Given the above, can we stick to YAML?

thechenky: Okay, sure that sounds reasonable.

thechenky:
@merlinnot can you resolve the conflicts so I can merge?

thechenky:
@patilharshal16 did you have some comments on this PR? I see you reviewed it....?

PR #517, opened 2019-07-11
Add code comments for storage events

esmacik:
Add code comments for storage events mostly from existing storage-externs.js


thechenky:
@esmacik can you run `npm run format:fix` and commit the changes? We check formatting in the continuous tests in travis - the tests are currently failing because formatting is off. I would suggest running `npm run format:fix` before every commit to ensure proper formatting 😄

PR #521, opened 2019-07-12
Increase type strictness of HttpsError class

merlinnot:
### Description

While working on changes to `providers/https.ts`, I got a little confused as there are three statuses we're dealing with:
- provided as an argument to HttpsError
- actual HTTP status code
- canonical error code for gRPC

This PR aims to minimise confusion by improving type strictness, so it's well defined what the return types of methods should be and how it's represented. It also moves the mapping into one place, instead of having a map outside of the class and a switch inside of it.

### Code sample

Not relevant.

### DIFF
--- a/src/providers/https.ts
+++ b/src/providers/https.ts
@@ -27,7 +27,6 @@ import * as _ from 'lodash';
 import { apps } from '../apps';
 import { HttpsFunction, optionsToTrigger, Runnable } from '../cloud-functions';
 import { DeploymentOptions } from '../function-configuration';
-import { assertNever } from '../utilities/assertions';
 
 export interface Request extends express.Request {
   rawBody: Buffer;
--- a/src/providers/https.ts
+++ b/src/providers/https.ts
@@ -127,32 +127,64 @@ export type FunctionsErrorCode =
   | 'data-loss'
   | 'unauthenticated';
 
+export type CanonicalErrorCodeName =
+  | 'OK'
+  | 'CANCELLED'
+  | 'UNKNOWN'
+  | 'INVALID_ARGUMENT'
+  | 'DEADLINE_EXCEEDED'
+  | 'NOT_FOUND'
+  | 'ALREADY_EXISTS'
+  | 'PERMISSION_DENIED'
+  | 'UNAUTHENTICATED'
+  | 'RESOURCE_EXHAUSTED'
+  | 'FAILED_PRECONDITION'
+  | 'ABORTED'
+  | 'OUT_OF_RANGE'
+  | 'UNIMPLEMENTED'
+  | 'INTERNAL'
+  | 'UNAVAILABLE'
+  | 'DATA_LOSS';
+
+interface HttpErrorCode {
+  canonicalName: CanonicalErrorCodeName;
+  status: number;
+}
+
 /**
- * Standard error codes for different ways a request can fail, as defined by:
+ * Standard error codes and HTTP statuses for different ways a request can fail,
+ * as defined by:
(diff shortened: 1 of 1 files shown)


src/providers/https.ts
```diff
@@ -163,88 +195,44 @@ export class HttpsError extends Error {
    * A standard error code that will be returned to the client. This also
    * determines the HTTP status code of the response, as defined in code.proto.
    */
-  readonly code: FunctionsErrorCode;
+  public readonly code: FunctionsErrorCode;
```
merlinnot: This is to make it more explicit that this property is public.

src/providers/https.ts
```diff
@@ -163,88 +195,44 @@ export class HttpsError extends Error {
...
   /**
    * Extra data to be converted to JSON and included in the error response.
    */
-  readonly details?: unknown;
+  public readonly details: unknown;
+
+  /**
+   * A wire format representation of a provided error code.
+   */
+  public readonly httpErrorCode: HttpErrorCode;
```
merlinnot: Not this is not `@hidden`. I'm not sure if we're allowed to make this change without consultation with the API council, but I decided to give it a shot as I found it very useful in some of the projects to have this mapping and I ended up duplicating the code. It would be very convenient to have it available.

src/providers/https.ts
```diff
@@ -163,88 +195,44 @@ export class HttpsError extends Error {
...
-    }
+    this.httpErrorCode = errorCodeMap[code];
   }
 
   /** @hidden */
-  public toJSON() {
-    const json: any = {
-      status: this.status,
-      message: this.message,
+  public toJSON(): HttpErrorWireFormat {
```
thechenky: Hm, just noticing that having public and @hidden together seems a little odd. @egilmorez do we expose toJSON() method on https provider? I see we do in other providers but I didn't find it for https.

src/providers/https.ts
thechenky: I think that makes sense - we can expose this. In fact this is great timing as we've been doing some docs changes and doing a full pass through to see whether methods/variables should be exposed or not. Let me know if you find others that you would find helpful to expose.

src/providers/https.ts
merlinnot: I applied this change in fe67bd7ab994ceb4dfd43416ccd8e35fb26be2b9 so we can merge if @egilmorez agrees. I'll revert otherwise.

src/providers/https.ts
thechenky: Ah, I missed a couple things here. On second pass, I see this is part of the `HttpsError` class, which we cannot change without API review. Let's take this out for now, and I can follow up on this later after discussing with the API council. Apologies for the confusion!

src/providers/https.ts
thechenky: Let's hide this for now.

src/providers/https.ts
```diff
@@ -204,6 +204,8 @@ export class HttpsError extends Error {
 
   /**
    * A wire format representation of a provided error code.
+   *
+   * @hidden
```
thechenky: Let's move @hidden to the top of the comment to be consistent with the rest.

merlinnot: `@hidden` is not used in combination with actual documentation anywhere in the codebase yet. When you hover over the use of it in any IDE, the thing you see is the content of JSDoc, that's why I'd prefer to have human-readable things above machine-readable. WDYT?

thechenky: Sure, I don't have strong feelings here, if it makes it easier to see in the IDE then let's keep it as is.

PR #531, opened 2019-07-18
Adding code comments and @hidden tags for src/providers/database.ts

esmacik:
* Adding code comments to database.ts for `database.ref()` method, `database.RefBuilder` and `database.DataSnapshot` from existing database-externs.js file
* Adding `@hidden` tags to any variable or method with an underscore before it in `database.DataSnapshot`


esmacik:
Here's also a link to see the staged changes :smile: https://firebase.devsite.corp.google.com/docs/reference/functions/providers_database_.html

PR #535, opened 2019-07-22
Adding code comments to analytics.ts

esmacik:
* Adding code comments to src/providers/analytics.ts, mostly from the existing analytics-externs.js file for `event` function, `AnalyticsEventBuilder`, `AnalyticsEvent`, and `UserPropertyValue`.
* The rest of the documentation changes are just formatting tweaks so the generated docs will look nicer.
* Adding hidden tags to the functions and a constant at the bottom of the file that need to be hidden.
* [Staged docs](https://firebase.devsite.corp.google.com/docs/reference/functions/providers_analytics_.html)

### DIFF
--- a/src/providers/analytics.ts
+++ b/src/providers/analytics.ts
@@ -36,8 +36,12 @@ export const provider = 'google.analytics';
 export const service = 'app-measurement.com';
 
 /**
- * Select analytics events to listen to for events.
- * @param analyticsEventType Name of the analytics event type.
+ * Registers a Cloud Function to handle analytics events.
+ *
+ * @param analyticsEventType Name of the analytics event type to which
+ *   this Cloud Function is scoped.
+ *
+ * @return Analytics event builder interface.
  */
 export function event(analyticsEventType: string) {
   return _eventWithOptions(analyticsEventType, {});
@@ -61,7 +65,7 @@ export function _eventWithOptions(
 /**
  * The Firebase Analytics event builder interface.
  *
- * Access via [`functions.analytics.event()`](functions.analytics#event).
+ * Access via [`functions.analytics.event()`](functions.analytics#.event).
  */
 export class AnalyticsEventBuilder {
   /** @hidden */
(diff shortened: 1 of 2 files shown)


src/providers/analytics.ts
```diff
@@ -36,8 +36,12 @@ export const provider = 'google.analytics';
 export const service = 'app-measurement.com';
 
 /**
- * Select analytics events to listen to for events.
- * @param analyticsEventType Name of the analytics event type.
+ * Registers a Cloud Function to handle analytics events.
```
egilmorez: We're inconsistent on this unfortunately, but wherever you see "Cloud Function" used to refer to an actual single, discrete function, replace it with "function."

src/providers/analytics.ts
```diff
@@ -73,12 +77,10 @@ export class AnalyticsEventBuilder {
...
    * Event handler that fires every time a Firebase Analytics event occurs.
    *
-   * @param {!function(!functions.Event<!functions.analytics.AnalyticsEvent>)}
-   *   handler Event handler that fires every time a Firebase Analytics event
+   * @param handler Event handler that fires every time a Firebase Analytics event
    *   occurs.
    *
-   * @return {!functions.CloudFunction<!functions.analytics.AnalyticsEvent>} A
-   *   Cloud Function you can export.
+   * @return A Cloud Function that you can export and deploy.
```
egilmorez: Here too.  Later on I can do a cleanup for instances that we have missed.  Thanks!

src/providers/analytics.ts
```diff
@@ -313,69 +319,86 @@ export interface DeviceInfo {
...
+   * The app's version name.
+   *
+   * Examples: "1.0", "4.3.1.1.213361", "2.3 (1824253)", "v1.8b22p6".
    */
   appVersion?: string;
 
   /**
-   *  Unique id for this instance of the app.
-   *  Example: "71683BF9FA3B4B0D9535A1F05188BAF3".
+   * Unique id for this instance of the app.
```
egilmorez: Suggest "ID" or "identifier."

src/providers/analytics.ts
```diff
@@ -102,13 +104,11 @@ export class AnalyticsEventBuilder {
   }
 }
 
-/**
- * Interface representing a Firebase Analytics event that was logged for a specific user.
- */
+/** Interface representing a Firebase Analytics event that was logged for a specific user. */
```
thechenky: Does this fit on one line? Can't tell

Issue #476, opened 2019-06-17
3.0.1 not using .runtimeconfig.json


zanona:
I am not sure it's just me but after upgrading to 3.0.1 with `.runtimeconfig.json` file under the functions directory no longer populates values from the file with:

```js
const functions = require('firebase-functions');
console.log(functions.config())
```

```
cd functions-dir
firebase functions:config:get > .runtimeconfig
firebase serve
```

the output is `{}`

but then if I downgrade to `firebase-functions` `3.0.0`, it works without problem.

thechenky:
Hi @zanona I'm assuming you found the resolution to your issue - would you mind sharing it here in case other developers run into the same problem?

zanona:
Hey Diana, unfortunately I haven’t found it. However it seems the issue was
not really the version number. But it is still a mystery to me how
.runtimeconfig.json works since it is still bot populating vars on firebase
serve even with the file in the functions directory.
Is there any specific documentation on how that works? At the moment I take
it’s magic :)

Il giorno lun 17 giu 2019 alle 13:41 Diana Tkachenko <
notifications@github.com> ha scritto:

thechenky:
@zanona Just to clarify - are you having issues on `firebase serve` only? As in, are the config values populated properly in your deployed functions? If it's `firebase serve` that is the problem, then make sure you're using the latest `firebase-tools` version (I recommend 6.12.0 or 7.0.0 if you're not using Node 6), since we've had some emulation issues that have been resolved recently. You can see more info and possibly related issues in this repo: https://github.com/firebase/firebase-tools/issues. Let me know if that resolves your issue.

To answer your general question, we use the `.runtimeconfig.json` file behind the scenes during deployment of functions (and during emulation as well), and this file actually gets uploaded to the GCF runtime and read on start to populate your configs so that you can do `functions.config().mykey` and get the value you set it to using `firebase functions:config:set mykey="some-value"`.

zanona:
Thanks @thechenky. Ok, I believe I have found the issue:

https://github.com/firebase/firebase-functions/blob/860dd00fb81148bd600ffe6b44d9bb6657fd3586/src/config.ts#L85

[...]

Issue #438, opened 2019-05-10
Not loading .runtime.config.js when using pnpm


ChrisBrownie55:
### [REQUIRED] Version info

**node:**

v8.16.0

**firebase-functions:**
^2.3.0

**firebase-tools:**

6.8.0

I'm using 6.8.0 because `firebase functions:shell` doesn't work past this version

**firebase-admin:**
^7.0.0

### [REQUIRED] Test case

Here is a repro of this problem. The README file explains how to reproduce it yourself and includes screenshots of the issue failing and it sort of not failing when a fix is introduced.
https://github.com/ChrisBrownie55/firebase-functions-repro

### [REQUIRED] Expected behavior

I expect the following json value when I set the environment value of `"test"` to `"value"`. And the extra config variables:
```json
{
  "test": "value",
  "databaseURL": "https://databaseName.firebaseio.com",
  "storageBucket": "projectId.appspot.com",
  "projectId": "projectId"
}
```

### [REQUIRED] Actual behavior

```json
{}
```

### Were you able to successfully deploy your functions?

No error messages at deploy time but I always get `FIREBASE_CONFIG environment variable is missing` at run time.

thechenky:
Hi @ChrisBrownie55 to set config variables for your function, please use `firebase functions:config:set KEY=VALUE` as explained in the [Set Environment Configuration docs](https://firebase.google.com/docs/functions/config-env#set_environment_configuration_for_your_project). Using `FIREBASE_CONFIG` variable and modifying `.runtimeconfig.json` file is not a supported way to configure your function. Closing this out.

Issue #328, opened 2018-10-03
Can not read "functions.config()" with`firebase serve`


BrentDietrich:
OS: macOS Sierra v10.12.6

### Version info
- firebase-tools: v4.2.1
```
"firebase-admin": "6.0.0",
"firebase-functions": "2.0.5",
```

**firebase-functions:**

### Test case

- I set the environment configs
`firebase functions:config:set mockservice.value="something"`
- I check that they were set
`firebase functions:config:get`
  - results
```
{
  "mockservice": {
    "value": "something"
  }
}
```

### Actual behavior

- When I run `firebase serve` with " functions.config()" in my code they always come up empty.

I checked both past issues ([264](https://github.com/firebase/firebase-functions/issues/264), [282](https://github.com/firebase/firebase-functions/issues/282)) and stackoverflow nobody seems to be having problems getting env configs locally.

I've deleted and re-installed node_modules, uninstalled and reinstalled firebase-tools and now I've resorted to a hacky workaround. Any help would be greatly appreciated. Thanks in advance.

laurenzlong:
Hi, you'll need to run `firebase functions:config:get > .runtimeconfig.json` in the functions directory, see https://firebase.google.com/docs/functions/local-emulator#serve_http_functions_from_the_command_line. (You can also manually create a .runtimeconfig.json file and put in test config values) This is to prevent you from accidentally using real API keys during testing without realizing it and having unintended consequences.

BrentDietrich:
@laurenzlong, I have been able to set and get config values from the command line using `config:set` and `config.get > .someconfig.json`. Everything I've tried has been conducted within the 'functions' folder. But, as mentioned in the "Test case" section above, I have only received an empty object in code when using `firebase serve`. Nothing I've tried has worked yet.

laurenzlong:
The file has to be named `.runtimeconfig.json`, anything else won't work. As well, if you are on Windows Powershell, you need to run: `firebase functions:config:get | ac .runtimeconfig.json` (This is documented in https://firebase.google.com/docs/functions/local-emulator)

alexandervlpl:
I had a similar problem when running the firebase emulator. Turns out `functions:config:get` doesn't output valid JSON. There are a bunch of wierd bytes containing "Firebase CLI" at the beginning:

$ firebase functions:config:get | hexdump -C
00000000  1b 5d 30 3b 46 69 72 65  62 61 73 65 20 43 4c 49  |.]0;Firebase CLI|
[...]

Issue #264, opened 2018-06-17
functions.config() returns empty JSON object in localhost.


ivanhdzd:
### Version info

```
"firebase-admin": "~5.12.0",
"firebase-functions": "^1.0.1",
```

**firebase-functions:**

### Steps to reproduce

Running firebase functions in localhost, using Typescript, functions.config() returns an empty JSON object:

```
import * as functions from 'firebase-functions';

export const hi = functions.https.onRequest((req, res) => res.send(functions.config().test));
```

And when test it cloud function in localhost it returns an empty JSON object.

### Expected behavior

I already set environment variables by the command:

```
firebase functions:config:set test.value="Test value" test.id="1"
```

And I asure that variables executing:

```
firebase functions:config:get
```

And it shows test variable value.

If I deploy my firebase functions it works, but to can develop my cloud functions I need to run it in localhost,

### Actual behavior

functions.get() returns an empty JSON object.

laurenzlong:
You need to run `firebase functions:config:get > .runtimeconfig.json` or manually create a `.runtimeconfig.json` with config values if you don't want to use your production config values. Make sure  the command is run in the functions folder. See https://firebase.google.com/docs/functions/local-emulator (the documentation isn't perfectly clear, we're in the middle of fixing it!)

CabuxaMapache:
It's happening to me. I have the ".runtimeconfig.json" in the "functions" dir generated with "functions:config:get > .runtimeconfig.json". But when I am in localhost and execute "functions.config()" in my project it returns an empty object.

nbgoodall:
Same situation as OP on macOS.

`.runtimeconfig.json` exists in the functions folder and `function.config()` returns an empty JSON object when run locally with `firebase serve`.

Have tried creating `.runtimeconfig.json` manually and same issue^^.

Thank you for the help 🙂

### Version info
```
"firebase-admin": "5.13.1",
"firebase-functions": "^2.0.2"
```

judoole:
Heads up to others arriving here, and getting an empty config from `functions.config()` with `firebase serve`; the `.runtimeconfig.json` has to be located in the `functions` dir, and will not adhere to the source-folder you have specified in **firebase.json**. For example, if this is your **firebase.json**
```
...
"functions": {
    "source": "functions/src/",
....
```
Your `.runtimeconfig.json` has to be placed in the functions-folder.

mbleigh:
@judoole that looks like a bug, thanks for the info.

[...]

PR #542, opened 2019-07-30
Change lookup of '.runtimeconfig.json' as relative to cwd

loklaan:
👋 Elllooooo, this is my first time contributing to firebase stuff.

### Description

The lookup for `.runtimeconfig.json` has an unexpected behaviour, which makes it only ever work if the `.runtimeconfig.json` file is at the root of your repo/project (next to `node_modules`).

This breaks local-emulation and deployed projects that are setup with _monorepos_, _yarn's pnp_, or _bundled fn's_.

Directly addresses: #476, #264, #438, #328, #172

Relates to: firebase/firebase-tools#653

### Notes

> _Is this a breaking change?_  

I'm inclined to say *no*, and it is just a minor change. But one could argue that folks now rely on this static `../../../.runtimeconfig.json` lookup. Counter-counter argument would be to say that these kinds of consumers are liking relying on `CLOUD_RUNTIME_CONFIG` anyway. 🤷‍♂️


mbleigh:
As written this is definitely behavior-changing, and there are instances where `cwd()` is not going to be the same as the package root directory.

I think the behavior probably ought to be to make your lookup additive, if `CLOUD_RUNTIME_CONFIG` is not present and the static lookup path does not exist. As-is, I worry that this would break people in unexpected ways.

loklaan:
You're right @mbleigh, and good approach. I've implemented it in the recent push.

It is now supporting the existing behaviour, then if those locations don't exist it will then traverse up from the `cwd` until it finds a `.runtimeconfig.json` file (or it doesn't, that's fine too).

mbleigh:
Hey @loklaan this issue is...well, really old now. But if you're interested in updating the PR to resolve merge conflicts and keep the "don't break existing behavior", I'm open to merging.

Sorry for letting this one slip through the cracks!

loklaan:
@mbleigh Oh brilliant I forgot about this - thanks for coming back to it!

Since I opened this PR, it looks like the behaviour has changed from:
1. Use path from ENV variable
2. Or, use a fixed path (`../../../.runtimeconfig.json`)

To:
1. _Use path from ENV variable_ (same)
2. **Or, use a fixed path (`${process.env.PWD}/.runtimeconfig.json`)** (changed)

This is great, although even after this long I still think that there should be a 3rd lookup attempt on parent directories above the working directory. So, I've updated the PR to support the current behaviours, and fallback to parent lookup using [`sindresorhus/find-up`](https://github.com/sindresorhus/find-up)

loklaan:
Ah looks like the node ^8 test fails.. For everyone? Let's ignore that I suppose.

Pinging you @mbleigh so, fingers crossed, this won't get buried again haha

mbleigh:
Thanks for the update. I hate to be a pain when you already were willing to overlook a year-plus gap, but we are trying to maintain or reduce the number of dependencies in the Firebase CLI. Do you think you could implement the same logic without `find-up`?

loklaan:
Haha all good, I get it @mbleigh .

Ahh I'm sure we could inline it, plucking the code from find-up itself?

I want to point out that `find-up` is quite light though, so including it might ok?
```
📦  find-up@5.0.0
=== Tarball Contents ===
1.1kB license
1.9kB index.js
917B  package.json
4.0kB readme.md
3.8kB index.d.ts
=== Tarball Details ===
name:          find-up
version:       5.0.0
filename:      find-up-5.0.0.tgz
package size:  3.8 kB
unpacked size: 11.8 kB
shasum:        4c92819ecb7083561e4f4a240a86be5198f536fc
integrity:     sha512-78/PXT1wlLLDg[...]vNtrJRiW6nGng==
total files:   5
```

mbleigh:
It's less about the size of the implementation and more about exposing us to code that changes without us knowing it (there are scary enough examples of that in NPM's history). FWIW [here's some similar code](https://github.com/firebase/firebase-tools/blob/master/src/detectProjectRoot.ts) we wrote for the Firebase CLI to find `firebase.json` in an ancestor directory tree. That could be a good base.

loklaan:
@mbleigh I've removed the find-up dep, replacing it with a similar function. 👍

mbleigh (changes requested):
Oh goodness, I had this review queued up months ago and forgot to hit submit 😦

loklaan:
Hi! Thanks for getting around to it. Last couple (er, many) months have been rough for many all round.

I'll be dropping this PR though 😞 as I don't have the involvement in Firebase projects anymore, and am spending time elsewhere. I encourage anyone that has subscribed to pick it up, if you want!

Thanks @mbleigh for the reviews, and sorry if this feels at all like I've wasted your time. I've definitely appreciated your thoughtfulness here, as well as your long term commitment to the Firebase team. All the best!

mbleigh:
👋  Love having folks in the community contribute, sorry this got lost in the shuffle. Have a good one!

PR #549, opened 2019-08-12
Linking to external types - docgen

esmacik:
This is the first iteration of my solution to linking to external library types for the Firebase Functions Reference documentation.
1. Map external types to their links.
1. Create a DOM of each HTML file.
1. When fixing links, intersect HTML data of generated doc and query tags of type `tsd-signature-type`.
1. Add link to matching types

In these staged files, there are now links to `UserInfo`, `UserRecord`, and `DocumentSnapshot` types:
* [functions.auth](https://firebase.devsite.corp.google.com/docs/reference/functions/providers_auth_.html)
* [functions.auth.UserBuilder](https://firebase.devsite.corp.google.com/docs/reference/functions/providers_auth_.userbuilder.html)
* [functions.firestore](https://firebase.devsite.corp.google.com/docs/reference/functions/providers_firestore_.html)
* [functions.firestore.DocumentBuilder](https://firebase.devsite.corp.google.com/docs/reference/functions/providers_firestore_.documentbuilder.html)

### DIFF
--- a/docgen/generate-docs.js
+++ b/docgen/generate-docs.js
@@ -41,6 +41,8 @@ const devsitePath = `/docs/reference/functions/`;
 
 const { JSDOM } = require("jsdom");
 
+const typeMap = require('./type-aliases.json');
+
 /**
  * Strips path prefix and returns only filename.
  * @param {string} path
@@ -129,22 +131,18 @@ function addTypeAliasLinks(data) {
   const htmlDom = new JSDOM(data);
   const fileTags = htmlDom.window.document.querySelectorAll(".tsd-signature-type");
   fileTags.forEach(tag => {
-    for (const type of typeMap.keys()) {
-      const lastTypeOccurrence = tag.textContent.lastIndexOf(type);
-
-      // Helps to prevent matching similar types, like `UserRecord` and `UserRecordMetadata`.
-      if (lastTypeOccurrence >= 0 && lastTypeOccurrence + type.length == tag.textContent.length) {
-        console.log('Adding link to '+type);
+    const mapping = typeMap[tag.textContent];
+      if (mapping) {
+        console.log('Adding link to '+tag.textContent+" documentation.");
         
         // Add the corresponding document link to this type
         const linkChild = htmlDom.window.document.createElement('a');
-        linkChild.setAttribute('href', typeMap.get(type));
+        linkChild.setAttribute('href', mapping);
         linkChild.textContent = tag.textContent;
         tag.textContent = null;
         tag.appendChild(linkChild);
       }
-    }
-  });
+    });
   return htmlDom.serialize();
 }
 
@@ -303,18 +301,6 @@ function fixAllLinks(htmlFiles) {
  *    links as needed.
  * 5) Check for mismatches between TOC list and generated file list.
  */
-var typeMap = new Map();
-fs.readFile('./type-aliases.json', (err, jsonString) => {
-  if (err) {
-    console.log('Error: type-aliases.json not found in this directory. Unable to link to external library documentation.');
-    return;
-  }
-
-  typeJson = JSON.parse(jsonString);
-  typeJson.forEach(alias => {
-    typeMap.set(alias.type, alias.link);
-  });
-});
 Promise.all([
   fs.readFile(`${contentPath}/toc.yaml`, 'utf8'),
   fs.readFile(`${contentPath}/HOME.md`, 'utf8')
--- a/docgen/generate-docs.js
+++ b/docgen/generate-docs.js
@@ -39,6 +39,9 @@ const contentPath = path.resolve(`${__dirname}/content-sources`);
 const tempHomePath = path.resolve(`${contentPath}/HOME_TEMP.md`);
 const devsitePath = `/docs/reference/functions/`;
 
+const jsdom = require("jsdom");
+const { JSDOM } = jsdom;
+
 /**
  * Strips path prefix and returns only filename.
  * @param {string} path
@@ -103,6 +106,8 @@ function renameFiles() {
  */
 function fixLinks(file) {
   return fs.readFile(file, 'utf8').then(data => {
+    console.log(file);
+    data = addTypeAliasLinks(data);
     const flattenedLinks = data
       .replace(/\.\.\//g, '')
       .replace(/(modules|interfaces|classes)\//g, '')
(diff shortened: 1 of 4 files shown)


docgen/generate-docs.js
```diff
@@ -103,6 +106,8 @@ function renameFiles() {
  */
 function fixLinks(file) {
   return fs.readFile(file, 'utf8').then(data => {
+    console.log(file);
```
hiranya911: Probably not necessary

docgen/generate-docs.js
```diff
@@ -271,6 +305,12 @@ function fixAllLinks(htmlFiles) {
  *    links as needed.
  * 5) Check for mismatches between TOC list and generated file list.
  */
+
+var typeMap = new Map();
```
hiranya911: Load this from a separate JSON file.

docgen/generate-docs.js
```diff
@@ -116,6 +121,35 @@ function fixLinks(file) {
...
+ * Adds links to external documentation for type aliases that
+ * reference an external library.
+ * 
+ * @param data File data to add external library links to.
+ */
+function addTypeAliasLinks(data) {
+  const htmlDom = new JSDOM(data);
+  const fileTags = htmlDom.window.document.querySelectorAll(".tsd-signature-type");
+  fileTags.forEach(tag => {
+    for (const type of typeMap.keys()) {
```
hiranya911: Why not something simpler and more efficient like:

```
const mapping = typeMap[tag.textContent];
if (mapping) {
   // Add <a> tag
}
```

docgen/generate-docs.js
```diff
@@ -271,6 +305,12 @@ function fixAllLinks(htmlFiles) {
  *    links as needed.
  * 5) Check for mismatches between TOC list and generated file list.
  */
+
+var typeMap = new Map();
+typeMap.set('DocumentSnapshot', `https://googleapis.dev/nodejs/firestore/latest/DocumentSnapshot.html`);
+typeMap.set('UserRecord', 'https://firebase.google.com/docs/reference/admin/node/admin.auth.UserRecord.html');
```
hiranya911: I think you should only rewrite fully qualified names like `firebase.auth.UserRecord`. Short names like `UserRecord` refers to the type alias defined in the the functions code itself.

docgen/generate-docs.js
```diff
@@ -39,6 +39,9 @@ const contentPath = path.resolve(`${__dirname}/content-sources`);
 const tempHomePath = path.resolve(`${contentPath}/HOME_TEMP.md`);
 const devsitePath = `/docs/reference/functions/`;
 
+const jsdom = require("jsdom");
+const { JSDOM } = jsdom;
```
hiranya911: Can this be just one line:

```
const { JSDOM } = require('jsdom');
```

docgen/generate-docs.js
esmacik: So if there's a reference to just `DocumentSnapshot`, like [here](https://firebase.devsite.corp.google.com/docs/reference/functions/providers_firestore_.html#change-constructor), should I replace the text with `firebase.firestore.DocumentSnapshot` or just not add a link at all?

docgen/generate-docs.js
```diff
@@ -271,6 +303,18 @@ function fixAllLinks(htmlFiles) {
  *    links as needed.
  * 5) Check for mismatches between TOC list and generated file list.
  */
+var typeMap = new Map();
+fs.readFile('./type-aliases.json', (err, jsonString) => {
```
hiranya911: You can do `const typeMap = require('./type-aliases.json');`

docgen/type-aliases.json
```diff
@@ -0,0 +1,14 @@
+[
```
hiranya911: Just make it the format your code expects:

```
{
  "DocumentSnapshot": "https://googleapis.dev/nodejs/firestore/latest/DocumentSnapshot.html",
  ...
}
```

docgen/generate-docs.js
hiranya911: I still think you should only replace full qualified external names like `firebase.auth.UserRecord`.

docgen/generate-docs.js
```diff
@@ -103,6 +107,7 @@ function renameFiles() {
  */
 function fixLinks(file) {
   return fs.readFile(file, 'utf8').then(data => {
+    data = addTypeAliasLinks(data);
```
thechenky: I see we are doing this for all files, but not all the files have one of the three links we'd like to make external so we're doing a lot of extra work per each file. If it's easy to know which files need this adjustment, I would suggest adding this check to only those files. However, if you feel this will make the code harder to read, or if you don't have a clear idea of which files will need to be modified, feel free to leave as is.

docgen/generate-docs.js
```diff
@@ -116,6 +121,31 @@ function fixLinks(file) {
...
 
+/** 
+ * Adds links to external documentation for type aliases that
+ * reference an external library.
+ * 
+ * @param data File data to add external library links to.
+ */
+function addTypeAliasLinks(data) {
+  const htmlDom = new JSDOM(data);
+  const fileTags = htmlDom.window.document.querySelectorAll(".tsd-signature-type");
```
thechenky: Not immediately clear why we're selecting `.tsd-signature-type`?

docgen/type-aliases.json
```diff
@@ -0,0 +1,5 @@
+{
```
thechenky: Is this the complete list of all external links?

docgen/generate-docs.js
thechenky: I would add a comment here explaining why this tag, so that the next contributor knows the context.

docgen/generate-docs.js
```diff
@@ -116,6 +121,31 @@ function fixLinks(file) {
...
+  fileTags.forEach(tag => {
+    const mapping = typeMap[tag.textContent];
+      if (mapping) {
+        console.log('Adding link to '+tag.textContent+" documentation.");
```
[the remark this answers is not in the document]
egilmorez: Looks like Erik fixed that.

docgen/generate-docs.js
egilmorez: Adding what I *think* Erik was doing -- just picking an ID to iterate through that would always be associated with potential external types in TypeDoc.

docgen/type-aliases.json
egilmorez: It is definitely a good question how we'll keep this file up to date as things change.  I'm giving myself a KR to write up the various TypeDoc procedures, and could add this there.

But for now, I think this is the full list of external types.

joehan:
@googlebot I fixed it.

joehan:
Overriding the CLA bot here, as we have a CLA from the commit author from a different email, and the commits causing the  check to fail are from github.

PR #575, opened 2019-10-29
Fixing some link breakage in reference files

egilmorez:
A bit of link breakage found by our "doc tracker."

Probably not all of them . . . but it's a start :)

Thanks!


egilmorez:
Looks like I need to clean out these previous commits first.

PR #617, opened 2020-01-23
Changelog for v3.3.1

tinaliang:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,2 +1,2 @@
-feature - Allow specifying retry policies for event triggered functions.
-feature - Add support for writing scheduled functions under handler namespace.
+fixed - updated documentation
+fixed - update types for integration tests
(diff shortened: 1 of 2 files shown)


merlinnot:
FYI: retry policies are not yet supported by `firebase-tools`, so I'm not sure if this should be released: https://github.com/firebase/firebase-tools/pull/1858. Even if we remove it from the Changelog, it will still be present in the API, so users who discover it might expect it to work.

tinaliang:
> FYI: retry policies are not yet supported by `firebase-tools`, so I'm not sure if this should be released: [firebase/firebase-tools#1858](https://github.com/firebase/firebase-tools/pull/1858). Even if we remove it from the Changelog, it will still be present in the API, so users who discover it might expect it to work.

We are currently looking into the work required in `firebase-tools` and this PR will remain unmerged until that has been resolved.

PR #624, opened 2020-02-27
add support for maxInstances in RuntimeOptions

efiShtain:
### Description

This is a matching change on firebase-tools to allow passing maxInstances parameter when creating or editing a function
https://github.com/firebase/firebase-tools/pull/1998

### Code sample

```javascript
import * as functions from 'firebase-functions'
functions
    .runWith({ maxInstances: 1 })
    .pubsub.schedule('every 1 minutes')
```


samtstern:
@joehan can you either take a look at this or assign to someone else?  It would need to go through API review but first we need to know if we want to do this.

There is a paired PR on firebase-tools to make it actually work.

efiShtain:
@samtstern 
just a friendly ping. anything I can do to speed up the process?

joehan:
Hey @efiShtain, we just added this proposal to our 3/19 API review meeting. Thanks so much for working on this - I'll update here again after the meeting.

joehan:
This was approved! Thanks again for working on this!

bolds07:
How does google server take in consideration the maxInstances value?
imho this is just a placebo parameter... there isnt any side effects of changing it...

I've a function with maxInstances set to 1... this function does simple return the currentTimeMillis so takes maximum 0.3 seconds to run (worst case).

**AND STILL GCF HAS 200 INSTANCES OF IT**

Clipboard01

so what is the point of that?
Documentation says that infrastructure will "try the best" to reach the `maxInstances `set but with no garantee and SOMETIMES the current instances value will exced it...

well as my graph shows THERE WASNT ANY TIME wihtout exceeding the `maxInstances ` set... and most of the time it was 15.000% over the desired value...
15.000% in what plannet this is an acceptable deviation?

im sorry if i'm being rude but it is clear that this whole parameter is just a waste of time for either you engineer team as for any developer trying to use your infrastructure

mbleigh:
@bolds07 I'd recommend writing in to [Firebase support](https://firebase.google.com/support/troubleshooter/contact) with details of how you configured your function and what you tested to try out `maxInstances`. This repo is for dealing with issues with the `firebase-functions` SDK, and since I see "Maximum function instances" seems to be set correctly I don't think there's any action to take on the SDK side.

PR #628, opened 2020-03-14
Adding reference pages for HandlerBuilder

egilmorez:
Adding this namespace per an internal bug/feature request.

Staged internally at https://firebase.devsite.corp.google.com/docs/reference/functions/handler_builder_.handlerbuilder?skip_cache=true


docgen/content-sources/toc.yaml
```diff
@@ -15,11 +15,19 @@ toc:
       - title: 'ChangeJson'
         path: /docs/reference/functions/cloud_functions_.changejson.html
 
+  - title: 'functions.HandlerBuilder'
```
laurenzlong: the title should be `functions.handler`, since that's how a developer actually writes it in code. 
e.g.
```exports.myFunction = functions.handler.database.ref.onWrite(...)```

File: src/handler-builder.ts [hunk 2]
```diff
@@ -35,6 +35,13 @@ import * as remoteConfig from './providers/remoteConfig';
 import * as storage from './providers/storage';
 import * as testLab from './providers/testLab';
 
+/**
+ * The `HandlerBuilder` class facilitates the writing of functions by producers
+ * of Firebase Extensions (currenly in limited beta).
```
laurenzlong: Revise to:
The `HandlerBuilder` class facilitates the writing of functions by producers of Firebase Extensions and developers that wish to use gcloud CLI or Cloud Console to deploy their functions. 

(I would actually remove the part of limited beta, it is not true, it's actually in EAP, and we also will definitely forget to update this comment later)

docgen/content-sources/toc.yaml
laurenzlong: This section can also be lower in the list (it can be the last item actually, since regular CF3 developers won't need it, and extensions docs will link directly to the handlers page.)

src/handler-builder.ts
```diff
@@ -35,6 +35,13 @@ import * as remoteConfig from './providers/remoteConfig';
...
+ * 
+ * **Do not use `HandlerBuilder` when writing normal functions for deployment via
+ * Cloud Functions for Firebase.** For normal purposes, use `FunctionBuilder`.
+ */
 export class HandlerBuilder {
   constructor() {}
```
laurenzlong: Can you add 
`/** @hidden */`
above the constructor? 

It should not show up in the docs.

PR #632, opened 2020-03-18
Eg handler toplevel

egilmorez:
Adding basic functions.handler docs for us to build out incrementally.

Thanks!


egilmorez:
Starting from clean slate to isolate format issues.

Issue #630, opened 2020-03-16
Deployed functions fails to read runtime config on Node 10


FredyC:
### Related issues

https://github.com/firebase/firebase-functions/issues/433

### [REQUIRED] Version info

**node:** 10.16.0

**firebase-functions:** 3.3.0

**firebase-tools:** 7.15.0

**firebase-admin:** 8.10.0

### [REQUIRED] Test case

I tried to create reproduction in repo https://github.com/FredyC/firebase-functions-repro
Unfortunately, it behaves correctly I did not found any difference from my _broken_ project.

### [REQUIRED] Steps to reproduce

See above

### [REQUIRED] Expected behavior

To have a deployed function that can read a runtime config.

### [REQUIRED] Actual behavior

We have like 15 functions deployed in production and they are running smoothly for several months now. They are all Node 10 functions.

Today, I needed to tweak one function a little, I've deployed it (with `--only` flag) and suddenly that one function is unable to read runtime config, it's always undefined.

In the console log, I've noticed the warning 

> Warning, FIREBASE_CONFIG and GCLOUD_PROJECT environment variables are missing. Initializing firebase-admin will fail

I assume it's major culprit if this issue, however considering that I have dependencies of the exact same version in a project and in _reproduction_ repo then I don't follow what this warning depends on.

### Were you able to successfully deploy your functions?

Functions do deploy, but fail in runtime when reading `functions.config()`.

tilgovi:
`firebase-admin` will check for `GOOGLE_CLOUD_PROJECT || GCLOUD_PROJECT`. I think `firebase-functions` should probably do the same. Some guidance from GCP about what the most modern option here is would be welcome.

the0rem:
This is pretty frustrating as we had the staging CI deploy and run fine however the deploy to production doesn't have the config data (because it was deployed after the bug was released).

There's also no way to rollback and fortunately we realised (after a couple hours of debugging) that the issue was specific to using the `region()` command which meant we could at least re-deploy our functions in favour of `https` and `firebase` firebase-function declarations.

The fact that this issue wasn't caught by your internal CI systems, as well as the fact that there's no way for the end user to recover and rollback from the issue is embarrassing.

Please rectify this with high priority.

FredyC:
@tilgovi You mean like either of those should be set locally? Why in that case it's working with fresh barebone project? I certainly haven't set any env variable.

@the0rem The workaround to deploy to default (US) region seems to be working indeed, thank you for that at least.

[...]

PR #634, opened 2020-03-19
Use process.env.PWD for finding .runtimeconfig.json

laurenzlong:
### Description

Fixes https://github.com/firebase/firebase-functions/issues/630

I've verified this with a project that's white listed for build packs, and I was able to successfully deploy to both Node 8 and Node 10, and make use of `functions.config()` and `admin.initializeApp()`

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,2 +1,2 @@
-**IMPORTANT: Please update to this version of firebase-functions if you are using Node.js 10 functions, otherwise your functions will break in production.** 
+**IMPORTANT: Please update to this version of `firebase-functions` if you are using Node.js 10 functions, otherwise your functions will break in production.** 
 Prevents deployment and runtime issue for Node.js 10 functions, due to `process.env.FIREBASE_CONFIG` and `functions.config()` being empty. (Issue #630)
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -68,9 +69,10 @@ export function firebaseConfig(): firebase.AppOptions | null {
 
   // Could have Runtime Config with Firebase in it as an ENV location or default.
   try {
-    const path =
-      process.env.CLOUD_RUNTIME_CONFIG || '../../../.runtimeconfig.json';
-    const config = require(path);
+    const configPath =
+      process.env.CLOUD_RUNTIME_CONFIG ||
+      path.join(process.env.PWD, '.runtimeconfig.json');
```
mbleigh: As written this is technically a breaking change. To make it not a breaking change we'd have to try the relative require first (ideally by checking to see if the file exists instead of relying on a caught exception) and then fall through to `$PWD` based if that doesn't work.

laurenzlong: I'm not sure that it is a breaking change. For trigger parsing, `PWD` is always the temp folder which is a copy of the functions folder. For in the GCF runtime, `PWD` is the root of the uploaded source code. We have always required that `package.json` be at the root of the functions directory, otherwise the function will not deploy. Therefore the location of `node_modules/firebase-functions` is always the same. So this change doesn't change any behavior.

changelog.txt
```diff
@@ -0,0 +1,2 @@
+**IMPORTANT: Please update to this version of `firebase-functions` if you are using Node.js 10 functions, otherwise your functions will break in production.** 
+Prevents deployment and runtime issue for Node.js 10 functions, due to `process.env.FIREBASE_CONFIG` and `functions.config()` being empty. (Issue #630)
```
mbleigh:
```suggestion
**Fixed:** Prevents deployment and runtime issues caused by upcoming changes to Cloud Functions infrastructure for Node.js 10 functions. (Issue #630)
```

mbleigh:
Oh also the changelog lines need to be a Markdown bullet list for the new release process I believe (need leading `*`)

PR #642, opened 2020-03-25
Updating Testlab portion of TOC

egilmorez:
Somehow, probably due to my own error, the TOC source fell behind the actual published TOC, omitting all the pages for Testlab.

This quick update fixes that for generating the docs going forward.

Thanks!


egilmorez:
Hm, even though toc-fix is a new branch based off master, some other changes have crept in . . . I think this may be because I did not close or refresh VScode.  

Might be best to close this PR and try anew.

PR #652, opened 2020-03-31
Add docstrings for handler.firestore, handler.database, and handler.https

laurenzlong:

### DIFF
--- a/src/handler-builder.ts
+++ b/src/handler-builder.ts
@@ -47,13 +47,22 @@ import * as testLab from './providers/testLab';
 export class HandlerBuilder {
   constructor() {}
 
+  /**
+   * Create a handler for HTTPS events.
+  
+   * `onRequest` handles an HTTPS request and has the same signature as an Express app.
+   * `exports.myFunction = functions.handler.https.onRequest((req, res) => { ... })`
+   *  This is identical to writing:
+   * `exports.myFunction = functions.https.onRequest((req, res) => { ... })`
+
+   * `onCall` declares a callable function for clients to call using a Firebase SDK.
+   * `exports.myFunction = functions.handler.https.onCall((data, context) => { ... })`
+   *  This is identical to writing:
+   * `exports.myFunction = functions.https.onCall((data, context) => { ... })`
+   * 
+   */
   get https() {
     return {
-      /**
-       * Handle HTTP requests.
-       * @param handler A function that takes a request and response object,
-       * same signature as an Express app.
-       */
       onRequest: (
         handler: (req: express.Request, resp: express.Response) => void
       ): HttpsFunction => {
@@ -74,8 +83,24 @@ export class HandlerBuilder {
     };
   }
 
+  /**
+   * Create a handler for Realtime Database events.
+   * 
+   * `ref.onCreate` handles new data creation.
+   * `exports.myFunction = functions.handler.database.ref.onCreate((snap, context) => { ... })`
+   
+   * `ref.onUpdate` handles data updates.
+   * `exports.myFunction = functions.handler.database.ref.onUpdate((change, context) => { ... })`
+   
+   * `ref.onDelete` handles data deletion.
+   * `exports.myFunction = functions.handler.database.ref.onDelete((snap, context) => { ... })`
+   
+   * `ref.onWrite` handles data creation, update, or deletion.
+   * `exports.myFunction = functions.handler.database.ref.onWrite((change, context) => { ... })`
+   */
   get database() {
     return {
+      /** @hidden */
       get instance() {
         return {
           get ref() {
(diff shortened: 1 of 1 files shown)


src/handler-builder.ts
```diff
@@ -47,13 +47,22 @@ import * as testLab from './providers/testLab';
 export class HandlerBuilder {
   constructor() {}
 
+  /**
+   * Create a handler for HTTPS events.
+  
+   * `onRequest` handles an HTTPS request and has the same signature as an Express app.
+   * `exports.myFunction = functions.handler.https.onRequest((req, res) => { ... })`
+   *  This is identical to writing:
```
egilmorez: We may want to really explicitly clarify that we're comparing extensions handlers to "regular" function handlers, or maybe avoid the "identical" phrasing.

Even just changing "identical" to "similar" would make me feel better.  WDYT?

src/handler-builder.ts
```diff
@@ -47,13 +47,22 @@ import * as testLab from './providers/testLab';
...
+   * `exports.myFunction = functions.https.onRequest((req, res) => { ... })`
+
+   * `onCall` declares a callable function for clients to call using a Firebase SDK.
+   * `exports.myFunction = functions.handler.https.onCall((data, context) => { ... })`
+   *  This is identical to writing:
```
egilmorez: We may want to really explicitly clarify that we're comparing extensions handlers to "regular" function handlers, or maybe avoid the "identical" phrasing.

Even just changing "identical" to "similar" would make me feel better.  WDYT?

src/handler-builder.ts
```diff
@@ -74,8 +83,24 @@ export class HandlerBuilder {
...
+   
+   * `ref.onDelete` handles data deletion.
+   * `exports.myFunction = functions.handler.database.ref.onDelete((snap, context) => { ... })`
+   
+   * `ref.onWrite` handles data creation, update, or deletion.
+   * `exports.myFunction = functions.handler.database.ref.onWrite((change, context) => { ... })`
+   */
   get database() {
     return {
+      /** @hidden */
```
egilmorez: Can you explain why this is hidden, but the return.getDocument for Firestore isn't?  I'm not clear on the pattern.

src/handler-builder.ts
```diff
@@ -74,8 +83,24 @@ export class HandlerBuilder {
     };
   }
 
+  /**
+   * Create a handler for Realtime Database events.
+   * 
+   * `ref.onCreate` handles new data creation.
+   * `exports.myFunction = functions.handler.database.ref.onCreate((snap, context) => { ... })`
+
```
egilmorez: Do you think we'd want to add statements like the ones in the HTTP handler comments pointing out the similar-but-different relationship with parallel "regular"functions code?  

I don't feel sure we need to clarify that anywhere but the top level comment, but will respect your opinion :)

laurenzlong: HTTP functions was the only ones where the behavior was identical. (Since the trigger definition for an HTTP function normally doesn't include other metadata anyways, whereas for the background event types they do include things like database path, etc.)

To put it another, with the background functions, the handler namespace is "similar by different". With HTTP functions, the handler namespace is "identical" to the original functions. 

Do you think there's a way we can communicate that better?

src/handler-builder.ts
laurenzlong: because we don't actually want people to use "functions.handler.database.instance.ref" since "instance" is redundant, it behaves the same as "functions.handler.database.ref". So we don't want to document it, but we do want to still leave the code there so that if someone does write "functions.handler.database.instance.ref" it will still work.

src/handler-builder.ts
laurenzlong: For HTTP functions, they are actually identical, haha. We created "handler.https" for consistency purposes (so that the handler namespace would have all the same event providers), not because it's actually useful. So I'm not sure what to do here. Maybe we can just remove these sentences altogether, in case they cause more confusion?

laurenzlong:
Thanks for the review Eric! Is there any chance you can try to stage this, so we can see what it looks like?

src/handler-builder.ts
egilmorez: Ah OK, I didn't pick that up about HTTP functions.  I do think that just removing the "this is identical" phrasing is probably best.

src/handler-builder.ts
egilmorez: Got it, thanks!

src/handler-builder.ts
egilmorez: Got it, didn't know that.

I think it's probably best to just remove these sentences, if you're OK with that.

laurenzlong: Done!

egilmorez:
Here as requested is a stage of these changes:

https://firebase.devsite.corp.google.com/docs/reference/functions/handler_builder_.handlerbuilder#https

One global suggestion: let's take each sample string and enclose it in this markup:

     * @example
     * ```javascript
     * exports.myFunction = functions.handler.https.onRequest((req, res) => { ... })
     * ```
That will create some whitespace and make everything pretty for the output.  

Thanks!

egilmorez:
Or actually, why not merge this as/is and I can apply the formatting like tech writers are supposed to do. 

Thanks!

PR #656, opened 2020-04-07
Crashlytics handler details and example formatting

egilmorez:
Formatting existing examples and adding Crashlytics.

Stage internally at https://firebase.devsite.corp.google.com/docs/reference/functions/handler_builder_.handlerbuilder#crashlytics

Thanks!

### DIFF
--- a/src/handler-builder.ts
+++ b/src/handler-builder.ts
@@ -51,10 +51,18 @@ export class HandlerBuilder {
    * Create a handler for HTTPS events.
   
    * `onRequest` handles an HTTPS request and has the same signature as an Express app.
-   * `exports.myFunction = functions.handler.https.onRequest((req, res) => { ... })`
-
+   *
+   * @example
+   * ```javascript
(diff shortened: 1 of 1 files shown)


src/handler-builder.ts
```diff
@@ -140,6 +182,36 @@ export class HandlerBuilder {
...
   }
 
+  /**
+   * Create a handler for Crashlytics events.
+  
+   * `issue.onNew` handles events where the app experiences an issue for the first time.
+
+   * @example
+   * ```javascript
+   * exports.myFunction = functions.handler.crashlytics.issue().onNew(async
```
laurenzlong, line 192: This should actually be `exports.myFunction = functions.handler.crashlytics.issue.onNew(async`. No brackets after issue. Everything in the handler namespace is formatted like this, no parameters. Same applies for the other examples for crashlytics.

egilmorez: Got it -- removed the parens.  Hope these are correct now, thanks!

PR #657, opened 2020-04-08
Eg moar handlers

egilmorez:
Adding the rest of the extensions handlers for review.

Thanks!

### DIFF
--- a/src/handler-builder.ts
+++ b/src/handler-builder.ts
@@ -223,7 +223,7 @@ export class HandlerBuilder {
   /**
    * Create a handler for Remote Config events.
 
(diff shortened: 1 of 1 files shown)


src/handler-builder.ts
```diff
@@ -233,6 +243,16 @@ export class HandlerBuilder {
...
   }
 
+  /**
+   * Create a handler for Anlytics events.
+   
+   * `event.onLog` handles the logging of Analytics conversion events.
+ 
+   * @example
+   * ```javascript
+   * exports.myFunction = functions.handler.analtics.event.onLog((event) => { ... })
```
laurenzlong, line 253: type: analytics

PR #665, opened 2020-04-21
Adds structured logging API.

mbleigh:
Node 10 runtime will soon support structured logging by emitting single-line JSON-encoded data to `STDOUT` / `STDERR`. This adds a new `firebase-functions/logger` export that provides structured logging in easy-to-use methods as well as `firebase-functions/logger/compat` that patches system `console` methods to output structured logs, solving the "multi-line log" problem for Node 10.

### TODO

- [x] Write reference docs
- [x] Write tests
- [x] Get API finalized / approved internally


mbleigh:
@laurenzlong for code review
@egilmorez for reference doc review

spec/logger.spec.ts
```diff
@@ -0,0 +1,105 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import * as logger from '../src/logger';
+
+const SUPPORTS_STRUCTURED_LOGS =
+  parseInt(process.versions?.node?.split('.')?.[0] || '8', 10) >= 10;
```
laurenzlong, line 6: This test is essentially relying on the runtime that the test is running in. This is a bit odd to be since the test behaves differently depending on which runtime it's running in. It would be better to stub `process.versions.node` so you can write separate tests for Node versions <10 and >=10

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+- Adds `functions.logger` SDK to enabled structured logging in the Node.js 10 runtime. Example:
```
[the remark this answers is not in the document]
mbleigh: Done.

spec/logger.spec.ts
mbleigh: There's no easy way to stub the behavior since it happens before anything is actually defined or exported (I tried). This was what I landed on and I think to some extent it makes sense -- it properly tests behaviors in each runtime and our CI tests do run the Node 8 tests.

src/logger.ts
```diff
@@ -0,0 +1,135 @@
+// Determine if structured logs are supported (node >= 10). If something goes wrong,
+import { format } from 'util';
```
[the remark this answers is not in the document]
mbleigh: Oops, fixed.

src/logger.ts
```diff
@@ -0,0 +1,135 @@
...
+
+// safely preserve unpatched console.* methods in case of compat require
+const unpatchedConsole = {
```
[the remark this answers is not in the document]
mbleigh: It does work, because the first thing `compat` does is require `logger` which preserves the unpatched console methods. Added a comment to clarify.

This wasn't in my original proposal but Kato asked if it was possible and it turned out to be easier than expected. I do worry about the chance for bugs here but this is just one way to use the SDK and provides a convenience. Anyone who encounters bugs we can just tell to use the other ways of using the SDK.

src/logger.ts
```diff
@@ -0,0 +1,135 @@
...
+  error: console.error,
+};
+
+// assume no since unstructured is safer.
+const SUPPORTS_STRUCTURED_LOGS =
+  parseInt(process.versions?.node?.split('.')?.[0] || '8', 10) >= 10;
+
+// Map LogSeverity types to their equivalent `console.*` method.
+const CONSOLE_SEVERITY: {
+  [severity: string]: 'debug' | 'info' | 'warn' | 'error';
```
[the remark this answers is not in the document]
mbleigh: TS doesn't let you do this 😢

spec/logger.spec.ts
laurenzlong: Bummer. Ok this works then.

src/logger.ts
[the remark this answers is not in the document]
laurenzlong: Ah yes, compat does import logger first, this is really subtle behavior, it's good you added a comment to clarify. 

I'm still not a fan of `compat` and I think it's just making this SDK hard to maintain (it's already way too complicated and has too many subtle behaviors) But if you feel like the risk is worth the user experience benefit, we can keep it in.

src/logger.ts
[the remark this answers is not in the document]
laurenzlong: 😢

CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
```
egilmorez, line 1: Typo, should be "to enable"

CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
+
```
egilmorez, line 2: A nit, but in a text flow we prefer "For example:"

CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
+  ```js
```
egilmorez, line 3: Wow, does this work?  In the d.ts files we use ```javascript, but a shorter form would be nice.

CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
...
+  });
+  ```
+
+  The logger can also override default behavior of `console.*` methods through a special require:
+
+  ```js
+  require('firebase-functions/logger/compat');
+  ```
+
+  In older runtimes, the logger will print to console, no structured data is saved.
```
egilmorez, line 18: Suggest "logger prints to the console, and no..."

src/logger.ts
```diff
@@ -0,0 +1,135 @@
+import { format } from 'util';
+
+// safely preserve unpatched console.* methods in case of compat require
```
egilmorez, line 3: Will these be parsed into the reference by our TypeDoc script?  If so, suggest "//Safely . . . require." with init cap and period.

src/logger.ts
```diff
@@ -0,0 +1,135 @@
...
+  warn: console.warn,
+  error: console.error,
+};
+
+// Determine if structured logs are supported (node >= 10). If something goes wrong,
+// assume no since unstructured is safer.
+const SUPPORTS_STRUCTURED_LOGS =
+  parseInt(process.versions?.node?.split('.')?.[0] || '8', 10) >= 10;
+
+// Map LogSeverity types to their equivalent `console.*` method.
```
egilmorez, line 17: If this goes to public docs, let's backtick `LogSeverity` or just say "log severity" in the text flow.

src/logger.ts
```diff
@@ -0,0 +1,135 @@
...
+  NOTICE: 'info',
+  WARNING: 'warn',
+  ERROR: 'error',
+  CRITICAL: 'error',
+  ALERT: 'error',
+  EMERGENCY: 'error',
+};
+
+/**
+ * LogSeverity indicates the detailed severity of the log entry. See [Cloud Logging docs](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity) for more.
```
egilmorez, line 32: Suggest backticks for `LogSeverity`

egilmorez, line 32: Suggest the actual doc section title, [`LogSeverity`](...)

src/logger.ts
```diff
@@ -0,0 +1,135 @@
...
+  | 'INFO'
+  | 'NOTICE'
+  | 'WARNING'
+  | 'ERROR'
+  | 'CRITICAL'
+  | 'ALERT'
+  | 'EMERGENCY';
+
+/**
+ * LogEntry represents a structured Cloud Logging entry. All keys aside from `severity` and `message` are
```
egilmorez, line 45: Suggest backticks for `LogEntry`

src/logger.ts
```diff
@@ -0,0 +1,135 @@
...
+ * included in the `jsonPayload` of the logged entry.
+ */
+export interface LogEntry {
+  severity: LogSeverity;
+  message?: string;
+  [key: string]: any;
+}
+
+/**
+ * Writes a LogEntry to `stdout`/`stderr` (depending on severity).
```
egilmorez, line 55: Suggest backticks for `LogEntry` (and every other literal that appears in the HTML output for the reference docs).

CHANGELOG.md
mbleigh: Done.

CHANGELOG.md
mbleigh: Done.

CHANGELOG.md
mbleigh: Yup 😄

CHANGELOG.md
mbleigh: Done.

src/logger.ts
mbleigh: Nope.

src/logger.ts
mbleigh: Won't be in public docs.

src/logger.ts
mbleigh: Done.

mbleigh: Done.

src/logger.ts
mbleigh: Done.

src/logger.ts
mbleigh: Done.

rhodgkins:
Any ideas when this is due to be merged/released?

sk-:
@mbleigh Will this PR correctly display the log severity when using `firebase-functions/logger/compat`? See https://issuetracker.google.com/issues/124403972 for context.

sk-:
@mbleigh I can confirm that the severity is respected, at least on Cloud Logging (I don't see it on Firebase yet).

However there are some issues:
1) The log entry in Cloud Logging is double quoted.
Screen Shot 2020-06-03 at 11 32 05 AM
2) The `message` field does not populate the `textPayload` log field, so one needs to change the queries to use `jsonPayload.message` instead. Is there any way to keep using `textPayload` for the queries? As it is now, using the compat module would be a breaking change, as it would break any metrics one may have.

mbleigh:
Hmm...interesting. Will try to dig into this a bit more before release. The way this works is using the [container logs contract](https://cloud.google.com/run/docs/logging#container-logs) from Cloud Run (newer Cloud Functions runtimes are built on the same infra).

It seems like it *should* be populating `textPayload` so long as there is no structured data attached to the log.

sk-:
@mbleigh You are right, if there is no structured data (meaning no extra fields other than `message` and `severity`) the logs do populate `textPayload` and are displayed without the double quotes.

By reading the logs contract docs, it only says about the message field that it is used as the display text, but mentions nothing of the special case when only these two fields are populated.

It'd be great if `textPayload `was always populated, and no double quotes were added to the displayed message.

mbleigh:
Let's continue this conversation on #697 since this PR is closed and merged.

PR #666, opened 2020-04-22
types: Allow no return value for action from DataSnapshot.forEach

rhodgkins:
This is to match the Admin SDK:

https://github.com/firebase/firebase-js-sdk/blob/37b98e9271c494a0fb58ca1960f8fcfaec49ade9/packages/database/src/api/DataSnapshot.ts#L129-L137


joehan:
LGTM

PR #683, opened 2020-05-08
Use `domain` in context to override RTDB domain

fredzqm:
See go/rtdb-multi-region-function-sdk

Multi-region realtime database will be served under different domains. In order to make sure the `DataSnapshot` returned by the function handler can connect to the appropriate RTDB, the backend will provide `domain` field in the pubsub payload.


spec/providers/database.spec.ts
```diff
@@ -457,8 +511,15 @@ describe('Database Functions', () => {
...
     const populate = (data: any) => {
-      const [instance, path] = database.resourceToInstanceAndPath(
-        'projects/_/instances/other-subdomain/refs/foo'
+      const [instance, path] = database.extractInstanceAndPath(
+        {
+          context: {
+            resource: {
+              name: 'projects/_/instances/other-subdomain/refs/foo'
+            },
+            domain: 'firebaseio-staging.com'
```
fredzqm, line 520: I was trying with the multi-region path. However, it didn't work due to a bug in the admin SDK.
I will fix that separately. For now, this will unblock testing in staging at least.

fredzqm:
@laurenzlong 
After talking with Ana, it looks like the `data` and `resource` are  processed in some special ways that I didn't understand much.
However, any other fields in the pubsub payload will be visible as part of the context (similar to `auth`). The easiest and cleanest way, IMO, is to supply a `domain` field.

The JS SDK right now has a bug parsing the multi-region domain, but it has no problem understanding the staging domain.
I will work on a fix in the JS SDK and update the dependencies here in a follow up CL.

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+- Adds support for multi-region RTDB.
```
fredzqm, line 1: Emm, not sure if we should include this in the CHANGE LOG.

After all the backend work is done, this will be the earliest version to support multi-region, but multi-region isn't a thing yet now.

fredzqm:
Move to private repos.

PR #684, opened 2020-05-08
Add RTDB service provider override

fredzqm:
This will allow us to override the service provider name to `firebaseio-staging.com` so that we can use it test environments.

FWIW, only two service names make sense: `firebaseio.com` or `firebaseio-staging.com`
They will work even for multi-region databases after we migrate FlowUpdate to integrate with metadata service instead: go/multi-region-function.

### DIFF
--- a/src/providers/database.ts
+++ b/src/providers/database.ts
@@ -38,7 +38,12 @@ import { applyChange } from '../utils';
 /** @hidden */
 export const provider = 'google.firebase.database';
 /** @hidden */
-export const service = 'firebaseio.com';
+function service(): string {
+ if (process.env.REALTIME_DATABASE_EVEENTFLOW_SERVICE) {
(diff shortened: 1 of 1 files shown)


fredzqm:
@laurenzlong 
I need some help understanding where function SDK is running in.
How would we populate the `process.env. REALTIME_DATABASE_PROVIDER_SERVICE`?

laurenzlong:
@fredzqm We would have to populate it in the CLI when deploying the function similar to https://github.com/firebase/firebase-tools/pull/2186. Note: this means that only customers that have updated their CLI will be able to use this feature, which I don't think is desirable.

fredzqm:
Okay, this is mostly for our own testing purpose. Do you see a simpler way?
Or is updating CLI + using env here is best option?

fredzqm:
Will move to use a customized zip of the build for staging.

Issue #685, opened 2020-05-12
TypeScript: @types/express version ambiguity breaks build


montasaurus:
<!-- DO NOT DELETE
validate_template=true
template_path=.github/ISSUE_TEMPLATE/---report-a-bug.md
-->

### Related issues
https://github.com/firebase/firebase-functions/issues/610
<!-- Are there any related firebase-functions issues that you found on this topic before deciding to open a new issue? Please link them here-->

### [REQUIRED] Version info

<!-- What versions of the following libraries are you using? Note that your issue may already
be fixed in the latest versions. -->

**node:** v13.12.0

<!-- Run node --version and print the output here-->

**firebase-functions:** 3.6.1

**firebase-tools:** 8.2.0

<!-- Run firebase --version and print the output here-->

**firebase-admin:** ^8.11.0

### [REQUIRED] Test case
```
import * as functions from "firebase-functions";
functions.https.onCall((data, context) => {}); // error TS2707: Generic type 'Request<P>' requires between 0 and 1 type arguments.
```
<!-- Provide a minimal, complete, and verifiable example (http://stackoverflow.com/help/mcve) -->

### [REQUIRED] Steps to reproduce
Upgrade the `firebase-functions` package to 3.6.1 with `@types/express` already installed and resolved to version 4.17.3 (which matches the spec for this package). Run `tsc` and watch the build fail.
<!-- Provide the steps needed to reproduce the issue given the above test case. -->

### [REQUIRED] Expected behavior
The build should succeed.
<!-- What is the expected behavior? -->

### [REQUIRED] Actual behavior
The build fails because the type signatures have changed in `@types/express` since the spec version in `package.json`. Breaking change here: https://github.com/DefinitelyTyped/DefinitelyTyped/commit/bb2673643a8292a565378ffe73b01c37aa7e1194

```
- error TS2707: Generic type 'Request<P>' requires between 0 and 1 type arguments.

64         onCall: (handler: (data: any, context: https.CallableContext) => any) => import("./cloud-functions").TriggerAnnotated & ((req: express.Request<import("express-serve-static-core").ParamsDictionary, any, any, import("express-serve-static-core").Query>, resp: express.Response<any>) => void) & import("./cloud-functions").Runnable<any>;
[...]

PR #686, opened 2020-05-12
pin @types/express version

montasaurus:
### Description
This pins the version of `@types/express` to fix the TypeScript build issue described in #685


laurenzlong:
Thank you for contributing! And sorry for the delayed review!

PR #690, opened 2020-05-14
Fixes to reference doc generation

egilmorez:
Attempting a cleanup of our reference doc generation in these areas:

-- Adding home page content.
-- Hiding things to prevent unnecessary pages from being generated.
-- Hiding specific things to exclude them from generated pages.
-- Adding missing things such as a TOC entry for CallableContext.

Staged internally at: https://firebase.devsite.corp.google.com/docs/reference/functions/providers_https_

Thanks!

### DIFF
--- a/src/providers/https.ts
+++ b/src/providers/https.ts
@@ -274,7 +274,7 @@ interface HttpRequest extends Request {
 }
 
 /** @hidden */
-// The format for the http body response to a callable function.
+// The format for an HTTP body response from a Callable function.
 interface HttpResponseBody {
   result?: any;
   error?: HttpsError;
--- a/src/providers/https.ts
+++ b/src/providers/https.ts
@@ -128,6 +128,7 @@ export type FunctionsErrorCode =
   | 'data-loss'
   | 'unauthenticated';
 
+/** @hidden */
 export type CanonicalErrorCodeName =
   | 'OK'
   | 'CANCELLED'
@@ -264,7 +265,7 @@ export interface CallableContext {
   rawRequest: Request;
 }
 
-// The allowed interface for an http request for a callable function.
+// The allowed interface for an HTTP request to a Callable function.
 /** @hidden*/
 interface HttpRequest extends Request {
   body: {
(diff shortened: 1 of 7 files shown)


src/providers/https.ts
```diff
@@ -262,18 +265,21 @@ export interface CallableContext {
 }
 
 // The allowed interface for an http request for a callable function.
```
laurenzlong, line 267: I know we're hiding this, but just to document it better in the source code, can you revise this to:

The allowed interface for an HTTP request to a Callable function.

src/providers/https.ts
```diff
@@ -262,18 +265,21 @@ export interface CallableContext {
...
+/** @hidden*/
 interface HttpRequest extends Request {
   body: {
     data: any;
   };
 }
 
+/** @hidden */
 // The format for the http body response to a callable function.
```
laurenzlong, line 276: I know we're hiding this, but just to document it better in the source code, can you revise this to:

The format for an HTTP body response from a Callable function.

src/utilities/assertions.ts
```diff
@@ -1,3 +1,4 @@
+/** @hidden */
```
laurenzlong, line 1: I think this can be combined with the lines below to make one comment block:

```
/** @hidden
  * @file Provides common assertion...
....
```

src/utilities/path.ts
```diff
@@ -1,3 +1,4 @@
+/** @hidden */
```
laurenzlong, line 1: I think this can be combined with the lines below to make one comment block

src/providers/https.ts
egilmorez: Done.

src/utilities/assertions.ts
egilmorez: Done.

src/utilities/path.ts
egilmorez: Done.

src/providers/https.ts
egilmorez: Done.

PR #705, opened 2020-06-11
Fix onRequest handler return type to allow promises (async)

p-young:
### Description

Fixes the types for onRequest so that it allows for async function handlers (returning a Promise)


sk-:
@p-young This seems to be a duplicate of #696 (which is already merged).

p-young:
@sk- Ah yea I was just making a similar change so that the type is exposed in the function builder (which is where I end up needing it).

Check the files changed, let me know if that makes sense / is the right thing

sk-:
Thanks @p-young for the clarification. That definition was indeed missing.

sk-:
Note that the PR is out of sync with the base branch.

sk-:
@laurenzlong Any chance this could be merged?

PR #711, opened 2020-06-21
Add proper generic types

Ranguna:
### Description
Currently, `firebase-admin` offers support for generics in the `DocumentSnapshot` and `QueryDocumentSnapshot` classes, but `firebase-function` does not make use of these generics. As such this PR attempts to extend generic support for the firestore provider.

### Code sample

Now users can do the following:
```ts
interface Doc {
  prop1: number;
}

functions.firestore.document('/messages/{documentId}').onCreate<Doc>((snap, context) => {
  console.log(snap.data().prop1)
});
```
And get autocomplete on `snap.data()` to the `Doc` interface.

As noted in https://github.com/googleapis/nodejs-firestore/issues/109. With this PR, the user has the option to not provide a type, thus keeping the use case of not having compile time type safety (unrelated with whether or not firestore provides run time type safety).

This follows the principal used in `nodejs-firestore`'s `Transaction` class https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L309, which allows the user to provide a type that describes a Document.

### DIFF
--- a/src/providers/firestore.ts
+++ b/src/providers/firestore.ts
@@ -41,8 +41,9 @@ export const service = 'firestore.googleapis.com';
 /** @hidden */
 export const defaultDatabase = '(default)';
 let firestoreInstance: any;
-export type DocumentSnapshot = firebase.firestore.DocumentSnapshot;
-export type QueryDocumentSnapshot = firebase.firestore.QueryDocumentSnapshot;
+export type DocumentData = firebase.firestore.DocumentData;
+export type DocumentSnapshot<T = DocumentData> = firebase.firestore.DocumentSnapshot<T>;
+export type QueryDocumentSnapshot<T = DocumentData> = firebase.firestore.QueryDocumentSnapshot<T>;
 
 /**
  * Select the Firestore document to listen to for events.
@@ -146,7 +147,7 @@ function _getValueProto(data: any, resource: string, valueFieldName: string) {
 }
 
 /** @hidden */
-export function snapshotConstructor(event: Event): DocumentSnapshot {
+export function snapshotConstructor<T>(event: Event): DocumentSnapshot<T> {
   if (!firestoreInstance) {
     firestoreInstance = firebase.firestore(apps().admin);
   }
@@ -161,7 +162,7 @@ export function snapshotConstructor(event: Event): DocumentSnapshot {
 
 /** @hidden */
 // TODO remove this function when wire format changes to new format
-export function beforeSnapshotConstructor(event: Event): DocumentSnapshot {
+export function beforeSnapshotConstructor<T>(event: Event): DocumentSnapshot<T> {
   if (!firestoreInstance) {
     firestoreInstance = firebase.firestore(apps().admin);
   }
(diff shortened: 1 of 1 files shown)


laurenzlong:
Sorry for the delay in getting to this! @hiranya911 since you're the expert on firebase-admin and have more context on this feature, is this a safe change to make? If so, do we have to increment our `firebase-admin` peer dependency version? (it is currently `^8.0.0`

hiranya911:
Actually we should ask @schmidt-sebastian since the Admin SDK Firestore API is coming from `@google-cloud/firestore`. One thing to consider is whether this constitutes a breaking change or not. I think it should be ok, but somebody should verify.

schmidt-sebastian:
A `DocumentSnapshot` should only be converted to `DocumentSnapshot<T>` if a converter is specified (via `DocumentReference.withConverter()` or similar). Simply marking a DocumentSnapshot as returning `T` will break the invariant that the object returned from `DocumentSnapshot.data()` is an instance of `T`. 
Firebase Functions does not have enough information to reconstitute T and as such, this change breaks our (and TypeScript's) typing guarantee. Unfortunately, that means that you will have to manually convert your documents to your data models.

Ranguna:
@schmidt-sebastian: In that case, a `withConverter` would be best in `functions.firestore.document('/messages/{documentId}')`, something like:
```ts
functions.firestore.document('/messages/{documentId}').withConverter(messageConverter).onCreate<Doc>((snap, context) => {
  console.log(snap.data().prop1)
});
```

schmidt-sebastian:
@Ranguna That would work - but for that change, I have to hand it back to @laurenzlong. Sorry :)

Ranguna:
@schmidt-sebastian no worries, thank you for the help 👍

laurenzlong:
Thanks @schmidt-sebastian for the input!

@Ranguna Is your suggestion in https://github.com/firebase/firebase-functions/pull/711#issuecomment-681822812 that much more convenient than doing the conversion inside of the function body? Can you help me understand your use case better?

Ranguna:
@laurenzlong sorry for the late reply!
I could do the conversion myself but I think it would be a nicer api if we just use the converter like that, the same way we use when fetching documents from firebase. 
It's not a requirement, it would just be nicer and consistent with the rest of the api (in my opinion, of course).
I just feels like firebase should have this built in.

I'm totally fine we you decide it's not needed.

Ranguna:
More info on `functions.firestore.document('/messages/{documentId}').withConverter(messageConverter)`:

I have a file with all the entity related things like:
```ts
\\ entity file

export const userDataCollectionName = "UserData";

export const userDataUsernameBuilder = () => functions.firestore.document(userDataCollectionName + "/{username}");

export default () => admin.firestore().collection(userDataCollectionName).withConverter(userDataConverter);
```

On another file where I want to create a trigger for `UserData`, I just need to do:
```ts
\\ triggers file

userDataUsernameBuilder().onCreate(...)
userDataUsernameBuilder().onDelete(...)
...
userDataUsernameBuilder().onAnythingThatIWantTodo(...)
```

Without the `.document(path).withConverter` syntax, I'll have to convert the document in each one of those triggers, it's essentially duplicate code and it makes me repeat myself instead of centralizing it by doing:
```ts
\\ entity file

...

export const userDataUsernameBuilder = functions.firestore.document(userDataCollectionName + "/{username}").withConverter(userDataConverter);

...
```
And that's it.

One problem that I see with this is if I do:
```ts
export const userDataUsernameBuilder = functions.firestore.document("UserData/{username}/{someOtherSubDocument}").withConverter(userDataConverter);
```
Where `userDataConverter` is only ready to accept documents from `UserData/{username}` and not `UserData/{username}/{someOtherSubDocument}`.
But I think it is easily understandable that the provided converter only works for the provided document in that document path. So this might not even be a problem.

laurenzlong:
Hey @Ranguna sorry for the late reply! I'm not at Firebase anymore, so I'll leave it up to @joehan to decide what to do.

Ranguna:
I'll close this PR since I think the idea now is to do this a little differently.
I won't create another issue or feature request for this because it doesn't look like people are really interested in this.

PR #714, opened 2020-06-23
Adding logger SDK to reference, with some edits.

egilmorez:
Adding logger SDK to reference docs, with some edits.

Staged internally at: https://firebase.devsite.corp.google.com/docs/reference/functions/logger_

Thanks!

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -20,7 +20,8 @@ export type LogSeverity =
   | 'EMERGENCY';
 
 /**
- * `LogEntry` represents a structured Cloud Logging entry. All keys aside from `severity` and `message` are
+ * `LogEntry` represents a [structured Cloud Logging](https://cloud.google.com/logging/docs/structured-logging)
+ * entry. All keys aside from `severity` and `message` are
  * included in the `jsonPayload` of the logged entry.
  */
 export interface LogEntry {
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -7,7 +7,7 @@ import {
 } from './logger/common';
 
 /**
- * `LogSeverity` indicates the detailed severity of the log entry. See [LogSeverity](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity) for more.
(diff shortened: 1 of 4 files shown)


egilmorez:
Correction on internal staging link -- it ends with an underscore: https://firebase.devsite.corp.google.com/docs/reference/functions/logger_

mbleigh:
Why is there an underscore at the end? I saw that in the yaml but don't
know the convention

src/logger.ts
```diff
@@ -31,7 +31,7 @@ export interface LogEntry {
 
 /**
  * Writes a `LogEntry` to `stdout`/`stderr` (depending on severity).
- * @param entry The LogEntry including severity, message, and any additional structured metadata.
+ * @param entry The `LogEntry` including severity, message, and any additional structured metadata.
```
jhuleatt, line 34: Will the `LogEntry` type show up in the reference? Can we link to it? Or maybe that happens automatically since it's an argument for the function?

src/logger.ts
```diff
@@ -56,7 +56,7 @@ export function write(entry: LogEntry) {
 
 /**
  * Writes a `DEBUG` severity log. If the last argument provided is a plain object,
- * it will be added to the `jsonPayload` in the Cloud Logging entry.
+ * it is added to the `jsonPayload` in the Cloud Logging entry.
```
jhuleatt, line 59: Here's a link to an explanation of jsonPayload: https://cloud.google.com/logging/docs/structured-logging

Not sure if that's TMI though

If added, this applies to all comments that mention jsonPayload

src/logger.ts
egilmorez: Yup!  TypeDoc creates a link for that.

src/logger.ts
egilmorez: Hm, that's a good reference for the guide material, as well as for a top-level comment for `logger` -- which I should probably add to the main page here.

egilmorez:
Michael has a really good question about the underscores . . . there is probably some work we can do in the docgen script to clean up the file names, similar to what we've done with the Web SDK TypeDoc script.

PR #722, opened 2020-07-07
Adds newly available GCF regions.

mbleigh:

### DIFF
--- a/src/function-configuration.ts
+++ b/src/function-configuration.ts
@@ -5,11 +5,15 @@ export const SUPPORTED_REGIONS = [
   'us-central1',
   'us-east1',
   'us-east4',
+  'us-west3',
   'europe-west1',
   'europe-west2',
   'europe-west3',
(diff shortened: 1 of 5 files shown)


src/function-configuration.ts
```diff
@@ -5,11 +5,15 @@ export const SUPPORTED_REGIONS = [
   'us-central1',
   'us-east1',
   'us-east4',
+  'us-west3',
```
samtstern, line 8: Now that GCF supports so many region could we consider changing the invalid region check from an error to a warning?  That way we don't have to keep chasing this list.

We can link people to this page from the warning:
https://cloud.google.com/functions/docs/locations

mbleigh: I'm not actually going to have it error or warn anymore. The supported regions will just be a TypeScript hinting convenience, and in a followup I think we should have the CLI do a live check against the locations API instead.

PR #729, opened 2020-07-08
Eg pubsub fixes

egilmorez:
Adding ScheduleBuilder class to TOC and adding minimal commenting.  

Happy to add additional comments for undocumented properties like retryConfig (or hide them if necessary); just LMK the right general verbiage.

Staged internally at: https://firebase.devsite.corp.google.com/docs/reference/functions/providers_pubsub_.schedulebuilder.

Thanks!


src/providers/pubsub.ts
```diff
@@ -98,6 +98,11 @@ export class TopicBuilder {
   }
 }
 
+/**
+ * The Google Cloud Pub/Sub schedule builder.
```
joehan, line 102: I hesitate to say `Google Cloud Pub/Sub schedule builder.` here, because that sounds to me like something that is built into pubsub. However, this is really just the Cloud Scheduler job that we create to run scheduled functions with.

Maybe this could be something like:
```
The schedule builder for scheduled functions. This describes the Cloud Scheduler job that will be deployed to trigger a scheduled function at the provided frequency.
```

src/providers/pubsub.ts
```diff
@@ -137,6 +142,14 @@ export class ScheduleBuilder {
     return this;
   }
 
+  /**
+   * Event handler that fires when a scheduled event occurs for
```
joehan, line 146: Language is a bit weird here - maybe something like "Event handler for scheduled functions. Triggered whenever the associated scheduler job sends a PubSub message"?

src/providers/pubsub.ts
egilmorez: I know . . . I think the reason this was overlooked was exactly that: we thought it was a class outside of CF3.

But here it is -- sourced from providers/pubsub.ts and generated along with the other files.

Joe do you think we should we actually be *hiding/omitting* this class from the reference docs instead of trying to add it and flesh it out?

laurenzlong: We should definitely be documenting it as a part of the firebase-functions reference docs. What about:

"The builder for scheduled functions, which are powered by Pub/Sub and Cloud Scheduler."

src/providers/pubsub.ts
laurenzlong: +1

docgen/content-sources/toc.yaml
```diff
@@ -106,6 +106,8 @@ toc:
         path: /docs/reference/functions/providers_pubsub_.message.html
       - title: 'TopicBuilder'
         path: /docs/reference/functions/providers_pubsub_.topicbuilder.html
+      - title: 'ScheduleBuilder'
+        path: /docs/reference/functions/providers_pubsub_.schedulebuilder
```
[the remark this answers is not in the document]
egilmorez: Good catch; it works, but there's no reason it should stand out.

src/providers/pubsub.ts
egilmorez: See what you think about this amalgam of all feedback.  Snuck in a link to the guide.

egilmorez:
Thanks for the feedback!

Again, feel free to suggest commenting for anything else in pubsub.ts that should be described in the reference docs.

src/providers/pubsub.ts
joehan: This version SGTM!

src/providers/pubsub.ts
```diff
@@ -98,6 +98,12 @@ export class TopicBuilder {
   }
 }
 
+/**
+ * Registers a Cloud Function to run at specified times.
+ *
+ * @param schedule The scheduling logic, in Unix Crontab or AppEngine syntax.
```
laurenzlong, line 104: I would go with the "The schedule" rather than "The scheduler logic", since it's not really a logic. It's just a definition of the schedule.

PR #731, opened 2020-07-09
Update ref docs for handler SDK to include scheduled functions.

laurenzlong:
### Description

Adding the ref docs for a feature that is already supported.

### DIFF
--- a/src/handler-builder.ts
+++ b/src/handler-builder.ts
@@ -306,7 +306,7 @@ export class HandlerBuilder {
   /**
    * Create a handler for Cloud Pub/Sub events.
    * 
-   * `topic.onPublish` handles messages published to a Pub/Sub topic from SDKs, Cloud Console, or Gcloud CLI. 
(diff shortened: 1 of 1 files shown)


src/handler-builder.ts
```diff
@@ -305,13 +305,20 @@ export class HandlerBuilder {
 
   /**
    * Create a handler for Cloud Pub/Sub events.
-
-   * `pubsub.onPublish` handles the publication of messages to a topic.
- 
+   * 
+   * `topic.onPublish` handles messages published to a Pub/Sub topic, from SDKs, Cloud Console or Gcloud CLI.
```
egilmorez, line 309: If those are three sources of publishing events, then I think omitting the comma after topic, and then adding one after "Cloud Console" will be more clear.

Should the "G" be capitalized in gcloud?

laurenzlong: Fixed! And you're right, gcloud seems to be lower case in the docs.

PR #736, opened 2020-07-14
Construct the database instance from the application options for the DataSnapshot on events

abuecker:
### Description

When wiring up to the local emulator suite we discovered that the database instance value for the `DataSnapshot` is constructed from the context's resource name and domain, building a URL that looks like `<projectId>.localhost`, which doesn't exist and surfaces the error:

```
Error: FIREBASE FATAL ERROR: Cannot parse Firebase url. Please use https://<YOUR FIREBASE>.firebaseio.com
```

Instead of constructing the domain from the context, using the `databaseURL` value from the application options, resolves this issue for us.  

Not sure if this is the correct approach, but here is a PR that addresses this and tests pass.

### DIFF
--- a/src/providers/database.ts
+++ b/src/providers/database.ts
@@ -234,7 +234,7 @@ export class RefBuilder {
         raw.data.delta,
         path,
         this.apps.admin,
-        dbInstance
+        this.apps.admin.options?.databaseURL || dbInstance
       );
     };
     return this.onOperation(handler, 'ref.create', dataConstructor);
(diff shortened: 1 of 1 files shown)


samtstern:
@abuecker thanks for sending this PR, I *think* this should already be fixed by this PR from last week:
https://github.com/firebase/firebase-functions/pull/727/files

That was just released in `3.8.0`.  Could you try it and let me know if it works for you?  I am also working on a companion PR in the emulators: https://github.com/firebase/firebase-tools/pull/2434

abuecker:
Ha!  Ships passing in the night.  Thanks for fixing this!

PR #744, opened 2020-07-22
Do not log the full request object on invalid onCall invocation

JrSchild:
### Description

When a malformed JSON body is sent to an `onCall` function, the entire request object will be logged. This is undesirable for the following reasons:

1. The object might contain sensitive data, which could be non-compliant with internal or regulatory privacy rules.
1. Each line in the output console, results in a structure log write which makes it easy to hit logging quota's. There are hundreds of logs per invalid request (picture below for just the first few lines).
1. Hitting logging quota's can be quite easily used for malicious purposes.
1. The output logs can become cluttered.

Screenshot 2020-07-22 at 10 46 10

### DIFF
--- a/src/providers/https.ts
+++ b/src/providers/https.ts
@@ -420,7 +420,7 @@ export function _onCallWithOptions(
   const func = async (req: Request, res: express.Response) => {
     try {
       if (!isValidRequest(req)) {
-        console.error('Invalid request', req);
+        console.error('Invalid request');
         throw new HttpsError('invalid-argument', 'Bad Request');
       }
 


mbleigh:
@JrSchild thanks very much for the pull request! This is definitely an issue, as we hadn't updated internal logging in the SDK to be compatible with newer runtimes (and I agree about not logging the full request). I wanted to do a bigger cleanup, so I've superseded this PR with #745 but thanks very much for posting!

JrSchild:
Works for me @mbleigh! Thanks for picking it up straight away.

Issue #552, opened 2019-08-30
Support VPC connectors


bjon:
It would be great if firebase-functions supported Serverless VPC connectors, like GCP functions do.

merlinnot:
Please keep in mind that there should be a way to specify names of VPC Connectors for each region separately - that is, if a function is deployed to multiple regions, each should be able to use a Connector with a different name.

merlinnot:
For anyone interested, here's a patch that can be used until the feature lands:

_Note it assumes the same name of connectors across regions and validates checksums of a file. Checksums have to be updated whenever there is an update to `firebase-tools`._

```bash
#!/usr/bin/env bash

# cspell:ignore cloudfunctions, shasum

# Patch `firebase-tools` package to support VPC Connector on functions.
# See: https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions/create

VPC_CONNECTOR="functions"

FILE_PATH="node_modules/firebase-tools/lib/gcp/cloudfunctions.js"
PATCHED_SHA1="96e7de86258358c0f5605f3f8dcb9ed76b99f9b7"

DATA_PATTERN="labels: options.labels,"
DATA_REPLACEMENT="labels: options.labels,vpcConnector:'${VPC_CONNECTOR}',"

MASK_PATTERN="\[\"sourceUploadUrl\", \"name\", \"labels\"\];"
MASK_REPLACEMENT="[\"sourceUploadUrl\", \"name\", \"labels\", \"vpcConnector\"];"

echo "Applying patch to firebase-tools with the VPC connector: ${VPC_CONNECTOR}"

set -x

sed \
-i.bak \
-e "s/${DATA_PATTERN}$/${DATA_REPLACEMENT}/g" \
${FILE_PATH}

sed \
-i.bak \
-e "s/${MASK_PATTERN}$/${MASK_REPLACEMENT}/g" \
${FILE_PATH}

sed -i.bak -e '$a\' ${FILE_PATH}

{ set +x; } 2>/dev/null

# The following is to ensure the monkey-patch applied to `firebase-tools` is
# not impacted by an update to the package.
echo "${PATCHED_SHA1}  ${FILE_PATH}" | shasum --check --portable --status -
```

kevinajian:
Hi @bjon,

Thanks for the feedback. Can you please file feature requests with support here: https://firebase.google.com/support/contact/bugs-features/

micksatana:
[...]

PR #752, opened 2020-08-06
Add support for VPC connectors in `functions.runWith`

pcboy:
### Description

This PR adds support for vpcConnectors inside runWith.
Refers to #552  and my corresponding PR in firebase-tools: [#2525](https://github.com/firebase/firebase-tools/pull/2525)

### Code sample

```
functions
      .runWith({
        vpcConnector: 'test-connector',
        vpcConnectorEgressSettings: 'PRIVATE_RANGES_ONLY'
      })
      .auth.user()
      .onCreate((user) => user);
```

### DIFF
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -261,6 +261,8 @@ export interface TriggerAnnotated {
     regions?: string[];
     schedule?: Schedule;
     timeout?: string;
+    vpcConnector?: string;
+    vpcConnectorEgressSettings?: string;
   };
 }
 
@@ -493,5 +495,14 @@ export function optionsToTrigger(options: DeploymentOptions) {
   if (options.maxInstances) {
     trigger.maxInstances = options.maxInstances;
   }
+
+  if (options.vpcConnector) {
+    trigger.vpcConnector = options.vpcConnector;
+  }
+
+  if (options.vpcConnectorEgressSettings) {
+    trigger.vpcConnectorEgressSettings = options.vpcConnectorEgressSettings;
+  }
+
   return trigger;
 }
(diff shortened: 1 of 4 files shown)


src/cloud-functions.ts
```diff
@@ -493,5 +495,12 @@ export function optionsToTrigger(options: DeploymentOptions) {
   if (options.maxInstances) {
     trigger.maxInstances = options.maxInstances;
   }
+
+  if (options.vpcConnector) {
+    trigger.vpcConnector = options.vpcConnector;
+    trigger.vpcConnectorEgressSettings =
+      options.vpcConnectorEgressSettings || 'ALL_TRAFFIC';
```
mbleigh, line 502:
```suggestion
    if (options.vpcConnectorEgressSettings) {
      trigger.vpcConnectorEgressSettings = options.vpcConnectorEgressSettings;
    }
```

The Cloud Functions default for this is `PRIVATE_RANGES_ONLY` and I don't want to change those defaults without a good reason. If we only conditionally set the value it will use the backend default which is what I'd expect.

pcboy: Yeah I wasn't quite sure about setting up a default either. If indeed GCP does it by default it's a better choice I guess.
I made the changes.

joshdifabio:
It looks like the Google Cloud console UI for editing cloud functions is currently broken for Firebase functions*, which makes it impossible to assign a VPC connector to a Firebase function right now, so it'd be great to get this merged and avoid having to use that fragile UI forevermore.

cubissimo:
If someone still stucked into this, i built a gcloud function (who works with gcloud console or gcloud cli to assign the vpc) to act like a "proxy" routing out egress traffic with a vpc connector. 

And i call this proxy function from my firebase functions when need to use the vpc.

I know its very aweful but there no way to do this direct in firebase by now.

edyrkaj:
Hi, 
Which are the available options for:
`vpcConnectorEgressSettings` => ('PRIVATE_RANGES_ONLY' | 'ALL_TRAFFIC')  I am supposing this based on values of two radio buttons that exists on cloud function / edit / network.

Thnx

PR #482, opened 2019-06-21
Allow specifying failure policies

merlinnot:
This PR:
- Makes runtime configuration validation more precise.
- Adds a `FailurePolicy`, which can be specified for all functions.
- Adds a warning message when `FailurePolicy` is used for `https` functions. Originally I planned to express it using more correct typings, it's doable, but not very clean. It's tricky because the runtime options are specified before the function trigger is chosen, so specifying a retry policy would result in limiting the number of options for triggers (removing `https`). We'd either need two classes returned conditionally based on the config (`EventTriggeredFunctionBuilder` when retry policy is specified, current `FunctionBuilder` otherwise) or we can just have this warning and ignore this option for `https` functions. Let me know which approach do you like better.
- Exposes `FailurePolicy` in `__trigger`, so it can be consumed by `firebase-tools`.
- Adds tests for setting and exposing `failurePolicy`.

I'm wondering how to approach releasing this functionality. Should we first implement support for it in `firebase-tools`, release a new version of it and start requiring it in `peerDependencies`? It would be weird if it's allowed to specify it here, but the setting is ignored by the CLI.

### Description

Resolves #425.

### Code sample

The API change is described in the issue linked above.

PR #760, opened 2020-08-13
Restore failurePolicy

samtstern:
### Description

See: https://github.com/firebase/firebase-functions/pull/482
See: https://github.com/firebase/firebase-functions/pull/623
See: https://github.com/firebase/firebase-tools/pull/1858

This is ready to go now that https://github.com/firebase/firebase-tools/pull/1858 is no longer blocked.

### Code sample

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Adds support for functions failure policies (#482)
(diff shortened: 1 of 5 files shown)


samtstern:
@merlinnot I am un-reverting your work here!

merlinnot:
That so awesome! I'll be super happy if we manage to land this feature!

PR #780, opened 2020-09-14
Update tests to change region based on env variable FIREBASE_FUNCTIONS_TEST_REGION

joehan:
### Description
Updates integration tests to use the region specified in the environment variable FIREBASE_FUNCTIONS_TEST_REGION. If not set, defaults to 'us-central1'.

### Code sample
export FIREBASE_FUNCTIONS_TEST_REGION=europe-west1
./integration-test/run_tests.sh <project-name>

### DIFF
--- a/integration_test/functions/src/auth-tests.ts
+++ b/integration_test/functions/src/auth-tests.ts
@@ -3,7 +3,9 @@ import * as functions from 'firebase-functions';
 import { expectEq, TestSuite } from './testing';
 import UserMetadata = admin.auth.UserRecord;
 
-export const createUserTests: any = functions.auth.user().onCreate((u, c) => {
+const REGION = process.env.FIREBASE_FUNCTIONS_TEST_REGION || "us-central1";
+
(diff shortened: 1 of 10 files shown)


integration_test/functions/src/auth-tests.ts
```diff
@@ -3,68 +3,82 @@ import * as functions from 'firebase-functions';
 import { expectEq, TestSuite } from './testing';
 import UserMetadata = admin.auth.UserRecord;
 
-export const createUserTests: any = functions.auth.user().onCreate((u, c) => {
-  const testId: string = u.displayName;
-  console.log(`testId is ${testId}`);
+const REGION = process.env.FIREBASE_FUNCTIONS_TEST_REGION || 'us-central1';
```
joehan, line 6: This (and the `.region(REGION)` calls) are the only real changes in this file - the formatter just happened to move everything around slightly.

mbleigh (approved):

PR #781, opened 2020-09-15
introduce `package-lock.json`

bkendall:
### Description

There are a bunch of `npm audit` issues that cannot be dealt with without a `package-lock.json` file. Additionally, `istanbul` is required but it is (a) unused, as far as I see, and (b) [deprecated](https://www.npmjs.com/package/istanbul), so I've removed it.

Any reason why `package-lock.json` shouldn't be included in this repository?

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-package-lock=false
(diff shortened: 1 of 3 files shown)


joehan (approved):
This looks good to me - I searched thru a bunch of old github issues and chat threads, and I couldn't find any reason why we shouldn't have a package-lock.json in this repo

staadecker:
Any update on this?

PR #842, opened 2021-01-08
Add support for 4GB memory option

### LINKED ISSUE #834: deploying a function with memory: '4GB' leads to getting deployed with the default 256MB
### Related issues

We have some cpu and memory intensive image resizes in a cloud function.
We have been using the max 2GB in production and everything has worked well.
I now wanted to make it even faster and so I changed to the new 4GB option.

But when I deploy it defaults back to 256MB

### [REQUIRED] Version info

**node:**
node for development: v14.15.2
node set for functions: 12

**firebase-functions:**
3.13.0

**firebase-tools:**
9.1.0

**firebase-admin:**
9.4.2

### [REQUIRED] Test case

```ts
export const storage = functions
  .region('europe-west1')
  .runWith({
    memory: '4GB',
    timeoutSeconds: 540,
  })
  .storage.bucket('my-special-bucket')
  .object()
  .onFinalize(projectStorageRouter)
```

```bash
firebase deploy --only functions:storage
```

function should deploy with 4GB memory

function deploys with 256MB memory

yes it deploys as usual but with wrong memory setting

leon:
### Description
The 4GB memory option was not added everywhere.

Fixes #834

Tests are added verifying that the 4GB option works

### DIFF
--- a/spec/function-builder.spec.ts
+++ b/spec/function-builder.spec.ts
@@ -286,4 +286,16 @@ describe('FunctionBuilder', () => {
       });
     }).to.throw();
   });
+
+  it('should allow setting 4GB memory option', () => {
+    const fn = functions
+      .runWith({
+        memory: '4GB',
+      })
+      .region('europe-west1')
+      .auth.user()
+      .onCreate((user) => user);
+
+    expect(fn.__trigger.availableMemoryMb).to.deep.equal(4096);
+  });
 });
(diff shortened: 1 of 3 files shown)


joehan (approved):
LGTM, sorry for the oversight!

joehan:
https://github.com/firebase/firebase-tools/pull/3026 filed this PR in firebase-tools to support this option in the functions emulator as well.

leon:
@joehan When do you think there will be a new version released?

PR #843, opened 2021-01-13
Fall back to updateTime if no readTime exists

### LINKED ISSUE #599: Firestore: ReadTime not populated
See https://github.com/firebase/firebase-js-sdk/issues/2459

It looks like `data.value.readTime` does not exist for a Firestore Event. When testing, this causes `snapshotConstructor` [to set the readTime argument](https://github.com/firebase/firebase-functions/blob/master/src/providers/firestore.ts#L157) to `undefined`. If the read time is not available, we should use the update or create time of a document.

iwikal:
### Description

Fixes #599. It seemed to work when I tested it in the emulator.


iwikal:
@googlebot I signed it!

inlined:
Per comment in this bug, should it be the `updateTime` or should it be the event timestamp?

inlined (approved):

iwikal:
All up to date now! I don't understand why the CI isn't completing though.

iwikal:
@inlined Is there something else I need to do before this can be merged?

inlined:
No, there was nothing left to do. Did you not have permission to merge yourself after I approved? Either way, I committed the change.

iwikal:
Nope, no merge button on my end.

MauriceArikoglu:
This is not working still @iwikal - tested on 3.22.0

Issue #863, opened 2021-02-23
Remove deprecated Crashlytics function triggers.


mbleigh:
The Crashlytics triggers are no longer functional and should be removed.

vn01971:
Why were the crashlytics triggers removed? Is there a plan to add them back at some point? Or is there an alternative? I'd like to make a request whenever a Crashlytics issue is created.

inlined:
Unfortunately the backend for the Crashlytics events integration was decommissioned. Not all products make it out of beta and this was one of them. You can still export Crashlytics data to BigQuery or set up velocity alerts that go to email, Slack, Jira, or PagerDuty. You can read more about the supported integrations at https://firebase.google.com/docs/crashlytics/velocity-alerts

yacut:
@inlined @mbleigh  Is there a plan to add a generic webhook integration or some way to submit custom integration?
I'm pretty sure that firebase crashlytics users want more integration than is currently available.

PR #867, opened 2021-03-01
Add relnotes for removing crashlytics

inlined:
Should have included this in the #863 fix.


CHANGELOG.md
```diff
@@ -0,0 +1 @@
+Remove no-longer-functional Crashlytics handlers
```
joehan, line 1: Nit:
```suggestion
- Removes non-functional Crashlytics handlers.
```
Reworded it to sound slightly better to me, but either way is fine. We should start each line in this with a`-`, however, so that they show up as bullet points on our release notes.

joehan (approved):
LGTM if you add a -

PR #868, opened 2021-03-01
Cleaning up Crashlytics trigger

joehan:
### Description
This trigger no longer works, so this PR cleans up all references to this (outside of the generated docs). As a follow up, we should clean this up from the reference docs as well.


joehan:
Ahhh, @inlined beat me to it with https://github.com/firebase/firebase-functions/pull/866

joehan:
Looks like there's one reference you missed @inlined - may as well merge this too to get it cleaned up

inlined (approved):

PR #871, opened 2021-03-05
Cleaning up docgen

joehan:
### Description
Cleaning up our docgen script:
- Removes crashlytics from toc.yaml
- Stops trying to delete css.map files if they don't exist

### DIFF
--- a/docgen/generate-docs.js
+++ b/docgen/generate-docs.js
@@ -323,7 +323,9 @@ Promise.all([
     // Clean up temp home markdown file. (Nothing needs to wait for this.)
     fs.unlink(tempHomePath);
     // Devsite doesn't like css.map files.
-    return fs.unlink(`${docPath}/assets/css/main.css.map`);
+    if (fs.existsSync(`${docPath}/assets/css/main.css.map`)) { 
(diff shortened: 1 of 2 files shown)


docgen/generate-docs.js
```diff
@@ -323,7 +323,9 @@ Promise.all([
     // Clean up temp home markdown file. (Nothing needs to wait for this.)
     fs.unlink(tempHomePath);
     // Devsite doesn't like css.map files.
-    return fs.unlink(`${docPath}/assets/css/main.css.map`);
+    if (fs.existsSync(`${docPath}/assets/css/main.css.map`)) { 
+      return fs.unlink(`${docPath}/assets/css/main.css.map`);
```
joehan, line 327: When I run this script, I don't get any css.map file generated by Typedoc, and so the script errors out. It's possible that the correct solution here is just to get rid of this unlink entirely, but I figured this was slightly safer in case I'm missing some reason that css.map would get generated .

inlined: I did some digging and I can't find out why either. LGTM.

inlined (approved):

PR #874, opened 2021-03-15
Update generate-docs to modern JS

inlined:
While reviewing `docgen/generate-docs.js` I had a little trouble getting my mind to think in such old JS 🙃

Updated to use async/await. In the meantime I caught a case where we weren't awaiting a promise.

### DIFF
--- a/docgen/generate-docs.js
+++ b/docgen/generate-docs.js
@@ -215,18 +215,18 @@ async function checkForMissingFilesAndFixFilenameCase() {
  * @param {boolean} shouldRemove Should just remove the file
  */
 async function checkForUnlistedFiles(filenamesFromToc, shouldRemove) {
-  const files = fs.readdir(docPath);
+  const files = await fs.readdir(docPath);
   const htmlFiles = files
     .filter(filename => filename.slice(-4) === 'html');
   const removePromises = [];
   const filesToRemove = htmlFiles
-    .select(filename => !filenamesFromToc.includes(filename))
-    .select(filename => filename !== 'index' && filename != 'globals');
+    .filter(filename => !filenamesFromToc.includes(filename))
+    .filter(filename => filename !== 'index' && filename != 'globals');
   if (filesToRemove.length && !shouldRemove) {
     // This is just a warning, it doesn't need to finish before
     // the process continues.
     console.warn(
-      `Unlisted file: ${filename} generated ` +
+      `Unlisted files: ${filesToRemove.join(" ,")} generated ` +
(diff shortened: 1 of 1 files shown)


docgen/generate-docs.js
```diff
@@ -72,59 +72,54 @@ function runTypedoc() {
...
-          if (err) console.log(err)
-        });
-      }
-    })
-  })
+async function renameFiles() {
+  const files = await fs.readdir(docPath);
+  console.log(files);
+  const renames = [];
+  files.forEach(file => {
```
joehan, line 90: Nit since it was already like this: https://google.github.io/styleguide/tsguide.html#iterating-containers suggests that we should avoid .foreach and use for loops instead.

docgen/generate-docs.js
```diff
@@ -305,69 +287,75 @@ function fixAllLinks(htmlFiles) {
...
+    const localVariablesIncludeString = `{% include "docs/web/_local_variables.html" %}\n`;
+    await fs.writeFile(`${docPath}/index.html`, localVariablesIncludeString + data);
+  } catch (err) {
+  if (err.stdout) {
+    console.error(err.stdout);
+  } else {
+    console.error(err);
+  }
+}
+})();
```
joehan, line 361: Add newline at end of file.

docgen/generate-docs.js
```diff
@@ -72,59 +72,54 @@ function runTypedoc() {
...
-      }
-    })
-  })
+async function renameFiles() {
+  const files = await fs.readdir(docPath);
+  console.log(files);
+  const renames = [];
+  files.forEach(file => {
+    let newFileName = file;
+    if (_.startsWith(file, "_") && _.endsWith(file, "html")) {
```
joehan, line 92: Nit - this is the only usage of lodash, and it can be done in native JS pretty simply.

joehan (approved):
LGTM with a few suggestion to modernize further if we want.

inlined:
NVM I found the bug. The old code used a global variable and I declared a local with the same name. I've removed the global.

egilmorez:
Glad to see you folks keeping this fresh and up to date!  Assuming this doesn't change anything about how to run the script to generate output.  If it does, or if it requires a node update, please LMK.  Thanks!

inlined:
@egilmorez The script returns warnings when it's run, though I've made sure there aren't any regressions. Do you want to verify that things are working correctly?

egilmorez:
You bet, I'll put that on my list.

PR #876, opened 2021-03-29
Correcting UserRecord falsey values to match types in admin sdk (#875)

### LINKED ISSUE #875: Wrong Types for UserRecord
* Operating System version: macOS Big Sur
  * Firebase SDK version: 9.5.0 (firebase-admin)
  * Firebase Product: Auth
  * Node.js version: v14.16.0
  * NPM version: 1.22.10

### [REQUIRED] Step 3: Describe the problem

Wrong types in `firebase-admin/lib/auth/index.d.ts` for `UserRecord`.

Attributes like `phoneNumber` or `photoURL` are marked as `string | undefined` while in reality they are `string | null`.
I thought I was going crazy 😅

#### Steps to reproduce:

Create a dummy on auth hook and console log some attributes that do not exist.

#### Relevant Code:

```
import { fns } from '../lib/admin'

export const onUserCreate = fns.auth.user().onCreate(async (user) => {
  console.log(user.photoURL) // Should be undefined but is null
})

```

joehan:
### Description
Stop setting these fields in UserRecord to null when they are not present - they should be undefined in that case. Fixes #875, though I am a little worried that this breaks users who were specifically checking `if (userRecord.field == null)`.

### DIFF
--- a/spec/providers/auth.spec.ts
+++ b/spec/providers/auth.spec.ts
@@ -131,17 +131,10 @@ describe('Auth Functions', () => {
       const record = auth.userRecordConstructor({ uid: '123' });
       expect(record.toJSON()).to.deep.equal({
         uid: '123',
-        email: null,
         emailVerified: false,
-        displayName: null,
-        photoURL: null,
-        phoneNumber: null,
         disabled: false,
         providerData: [],
         customClaims: {},
-        passwordSalt: null,
-        passwordHash: null,
-        tokensValidAfterTime: null,
         metadata: {
           creationTime: null,
           lastSignInTime: null,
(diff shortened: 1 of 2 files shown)


i14h:
@karayu what do you think about this issue? the change will be possibly breaking for some users

inlined:
1. IIRC the behavior was explicitly to provide exists-but-falsey behavior, so we should just replace the `null` with `undefined` so that iteration still includes those keys (we can revisit if we think this behavior is wrong)
2. Per offline triage, we cannot fix this until we make a major version bump. Will consider in vNext.

PR #880, opened 2021-04-16
SDK version that works with firebase-functions@exp

inlined:
Modified the SDK and integration tests to use the new (non compat-mode) version of `firebase-admin@exp`. This removes support for Node.js 10, so we won't merge into master until @exp is the new release.


inlined:
Dropping for now. We can reimplement if/when we raise the min version of the SDK.

PR #889, opened 2021-05-12
Update release script to build packages using Node 14

taeold:
https://github.com/firebase/firebase-functions/pull/885 changes typescript target to ES2018, and running `npm test` in Node 8 no longer works.

We update the release script to update to an environment that uses Node 14. As a side-effect, we have to update the publish.sh script where we use `npm version` - npm v7 requires the local git directory to be clean by default.

### DIFF
--- a/scripts/publish-container/Dockerfile
+++ b/scripts/publish-container/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:14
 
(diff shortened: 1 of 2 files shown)


joehan (approved):

taeold:
I built and published a new version of publisher-container that includes changes made here.

PR #896, opened 2021-05-17
Fix publish script to commit changes to version info in package{-lock}.json.

taeold:
While migrating publish script to run on Node.js 14 in https://github.com/firebase/firebase-functions/pull/889, we accidentally  removed mechanism to commit and push changes to the `version` attribute in `package.json` and `package-lock.json`. This broke the publishing pipeline by pushing out wrong versions of the release.

We revert the changes made in the faulty PR. I still don't know why the release script failed with error messages like below when we migrated the publisher container to use nodejs14:

```
Step #11: Making a minor version...
Step #11: npm ERR! Git working directory not clean.
Step #11:
Step #11: npm ERR! A complete log of this run can be found in:
Step #11: npm ERR!     /builder/home/.npm/_logs/2021-05-12T18_47_59_428Z-debug.log
Finished Step #11
```

I'm adding a `git status` command before `npm version` command to help us more easily diagnose the issue if it happens again on our next release.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -89,6 +89,7 @@ echo "Ran publish build."
 echo "Making a $VERSION version..."
 # TODO: Remove the following command.
 #   npm version command had previously failed claiming unclean git repo, and we don't know why.
+echo "DEBUG: Running git status to show dirty files..."
 git status
 npm version $VERSION
 NEW_VERSION=$(jq -r ".version" package.json)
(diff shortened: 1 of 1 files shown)


inlined (commented):
Why is `git status` necessary?

taeold:
> Why is `git status` necessary?

I couldn't figure out why `npm version` started to fail when I upgraded the publisher container image to use nodejs14 (not reproducible on my machine). Hoping that adding `git status` will help us debug the issue when we have the issue again in the future.

inlined (approved):
LGTM. I might extend the comment to make it clear that this is a debug print (Could even be a header you print first). I thought git status was some weird hack to trigger a subtle workaround.

joehan (approved):

PR #897, opened 2021-05-18
Optionally defer the loading of the Firestore SDK further

inlined:
Serializing an idea that's been mulling around for a while. Firestore functions are slow because it takes a long time to load the Firestore SDK. This is generally believed to be because the protobuff and gRPC libraries cause too much I/O to load. This PR replaces the DocumentSnapshot parameter (and its contained DocumentReference) with proxy objects that implement the API surface. The most commonly used methods are reimplemented without relying on the Firestore SDK. More complex functions then load the SDK to create a real DocumentSnapshot or DocumentReference inside the proxy object and delegate implementations to them. This PR depends on the fact that `import` statements in TypeScript only import the type definition and tell the TypeScript compiler to `require` the file in the first scope where an actual _value_ is referenced.

This change improves the performance of:
* DocumentSnapshot::createTime
* DocumentSnapshot::updateTime
* DocumentSnapshot::readTime
* DocumentSnapshot::data()
* DocumentSnapshot::get()
* DocumentSnapshot::protoField()
* DocumentSnapshot::toWriteProto
* DocumentSnapshot::toDocumentProto
* DocumentReference::path
* DocumentReference::id
* DocumentReference::formattedName

This change will still load the Firestore SDK on:
* DocumentSnapshot::isEqual()
* DocumentReference::parent
* DocumentReference::firestore
* DocumentReference::get()
* DocumentReference::collection()
* DocumentReference::listCollections()
* DocumentReference::create()
* DocumentReference::delete()
* DocumentReference::set()
* DocumentReference::update()
* DocumentReference::onSnapshot()
* DocumentReference::isEqual()
* DocumentReference::withConverter()

A test project has been created at github.com/inlined/firestore-perf. In a test of 600 SDK loads with the old code and 600 SDK loads with the new code, this resulted in the following savings:

* mean: 21% savings
* p50: 17% savings
* p90: 21% savings
* p95: 68% savings
* p99: 21% savings

This change has the downside that event.data no longer subclasses DocumentSnapshot and event.data.ref no longer subclasses DocumentReference. Though it is unlikely that (m)any customers depend on this, I've guarded the code by `functions.firestore.useExperimentalProxyObjects()` to protect against errors until a major version bump. This PR cannot be submitted until `useExperimentalProxyObjects()` is marked hidden or is approved in an API review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,0 @@
-- Fixes a bug where typescript would fail to compile with old (but supported) versions of firebase-admin
-- Replaces 3.13.3 which was an inappropriately numbered version
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,2 @@
+- Fixes a bug where typescript would fail to compile with old (but supported) versions of firebase-admin
+- Replaces 3.13.3 which was an inappropriately numbered version
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "firebase-functions",
-  "version": "3.13.2",
+  "version": "3.14.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "firebase-functions",
-  "version": "3.14.0",
+  "version": "3.14.1",
   "description": "Firebase SDK for Cloud Functions",
   "keywords": [
     "firebase",
--- a/scripts/publish/CHANGELOG.md
+++ b/scripts/publish/CHANGELOG.md
@@ -1,2 +1,0 @@
-- Fixes a bug where typescript would fail to compile with old (but supported) versions of firebase-admin
-- Replaces 3.13.3 which was an inappropriately numbered version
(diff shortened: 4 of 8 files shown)


inlined:
Added more tests to see the raw values of how many files were loaded by requiring the SDK and running trampoline code of a Firestore Function (minimum Firebase-induced coldstart latency). This led me to rewrite the change a bit. Now we have:

| version |  files loaded by require  | files loaded after trampoline |
|--------|-----------------------|-----------------------------|
|  original |  39 | 338 |
| lazy | 40 | 40 |

I reran the performance tests and got the following stats for time taken:

|       version |    mean | p50 | p90 | p95 | p99 |
|------------|--------|------|-----|------|----|
|       original |   298   |  229 | 306 | 912 | 1545 |
|       lazy      |   200   |   202 | 303 | 391 | 806 |
|       delta   |    32%  |   12% |  1% |  57% |  48% |

I'm starting to feel a bit nervous that these numbers are too random (even if they always trend for lazy being faster).

inlined (commented):
I strongly suspect that https://github.com/googleapis/nodejs-firestore/pull/1517 will be sufficient and won't have the downside of proxy objects

inlined:
Will fix the issue in the Firestore SDK itself with googleapis/nodejs-firestore#1517

PR #899, opened 2021-05-26
Adds support for setting user labels on functions

joehan:
### Description
Adds 'labels' to runtimeOptions, per go/firebase-functions-labels. This lets you set user labels on functions!

Note: I did make one minor change to the behavior described in go/firebase-functions-labels - if the labels option is not set in runWith(), we still overwrite labels. Originally, I proposed this behavior to support users who already manually set labels. However, the existing behavior of Firebase Functions is to overwrite labels. We also avoid the awkward case where removing 'labels' from 'runtimeOptions' doesn't cause labels to be removed. 

### Code sample
```
functions.runWith({
  labels: {
    "my-label": "my-value",
  }
})...
```

### Testing
I tested the following cases and confirmed that labels were set correctly via the Equivalent REST button in GCP console.
- Setting labels on normal functions.
- Setting labels on scheduled functions.
- Updating a function to change the labels.
- Setting too many labels throws an error before deployment
- Setting invalid labels throws an error before deployment

### DIFF
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -417,7 +417,7 @@ export function makeCloudFunction<EventData>({
         },
       });
       if (!_.isEmpty(labels)) {
-        trigger.labels = labels;
+        trigger.labels = Object.assign(labels, trigger.labels);
       }
       return trigger;
     },
@@ -551,6 +551,10 @@ export function optionsToTrigger(options: DeploymentOptions) {
         `Invalid option for serviceAccount: '${options.serviceAccount}'. Valid options are 'default', a service account email, or '{serviceAccountName}@'`
       );
     }
+  }
+
+  if (options.labels) {
+    trigger.labels = options.labels;
   }
 
   return trigger;
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -417,7 +417,7 @@ export function makeCloudFunction<EventData>({
         },
       });
       if (!_.isEmpty(labels)) {
-        trigger.labels = Object.assign(labels, trigger.labels);
+        trigger.labels = { ...labels, ...trigger.labels };
       }
       return trigger;
     },
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -417,7 +417,7 @@ export function makeCloudFunction<EventData>({
         },
       });
       if (!_.isEmpty(labels)) {
-        trigger.labels = { ...labels, ...trigger.labels };
+        trigger.labels = { ...trigger.labels, ...labels };
       }
       return trigger;
     },
--- a/src/function-builder.ts
+++ b/src/function-builder.ts
@@ -27,6 +27,7 @@ import { CloudFunction, EventContext } from './cloud-functions';
 import {
   DeploymentOptions,
   INGRESS_SETTINGS_OPTIONS,
+  MAX_NUMBER_USER_LABELS,
   MAX_TIMEOUT_SECONDS,
   RuntimeOptions,
   SUPPORTED_REGIONS,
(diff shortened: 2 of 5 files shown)


src/function-builder.ts
```diff
@@ -120,6 +121,61 @@ function assertRuntimeOptionsValid(runtimeOptions: RuntimeOptions): boolean {
...
+    );
+    if (invalidLengthValues.length > 0) {
+      throw new Error(
+        `Invalid labels: ${invalidLengthValues.join(
+          ', '
+        )}. Label values must be less than 64 charcters.`
+      );
+    }
+
+    // Keys can contain lowercase letters, foreign characters, numbers, _ or -. They must start with a letter.
```
taeold, line 154: nit: `They must start with a lowercase letter`

Also, I'd appreciate if we link https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements somewhere!

src/function-configuration.ts
```diff
@@ -147,6 +149,11 @@ export interface RuntimeOptions {
    * Ingress settings
    */
   ingressSettings?: typeof INGRESS_SETTINGS_OPTIONS[number];
+
+  /**
+   *
```
taeold, line 154: missing documentation?

src/function-builder.ts
```diff
@@ -120,6 +121,61 @@ function assertRuntimeOptionsValid(runtimeOptions: RuntimeOptions): boolean {
       `serviceAccount must be set to 'default', a service account email, or '{serviceAccountName}@'`
     );
   }
+
```
taeold, line 124: Should we explicitly disallow implicitly reserved labels? e.g. I think we reserve the use of `deployment-tools` (as in, we are going to overwrite whatever you put as `deployment-tools` with whatever the CLI wants it to be.

src/cloud-functions.ts
```diff
@@ -417,7 +417,7 @@ export function makeCloudFunction<EventData>({
         },
       });
       if (!_.isEmpty(labels)) {
-        trigger.labels = labels;
+        trigger.labels = { ...labels, ...trigger.labels };
```
inlined, line 420: I think I'd actually flip this. I'd rather bugs show up in the form of "Why can't I set label X" as opposed to "Why am I seeing arbitrary mysterious behavior that depended on a reserved label"

src/function-builder.ts
inlined: I like this. I'd also possibly disable any label that starts with `firebase-`

src/function-configuration.ts
```diff
@@ -94,6 +94,8 @@ export const DEFAULT_FAILURE_POLICY: FailurePolicy = {
   retry: {},
 };
 
+export const MAX_NUMBER_USER_LABELS = 58;
```
inlined, line 97: Does this actually need to be exported here (and thus in index.ts)? We could just put this const in cloudfunctions.ts

inlined (approved):

src/function-builder.ts
joehan: Good call, I'll add that link. However, they can start with an international character as well -technically, anything in the Unicode groups 'letters lowercase' or 'letters other'

src/function-configuration.ts
joehan: We don't technically need to. However, we export other similar constants here, like MAX_TIMEOUT_SECONDS, so I figured it was appropriate to do so for this one too

src/function-configuration.ts
joehan: Good catch. I also went through and cleaned up some of these other, less descriptive comments.

PR #901, opened 2021-05-30
Make DocumentBuilder handlers into generic types

morgs32:
So that we can type our functions better!

### DIFF
--- a/src/providers/firestore.ts
+++ b/src/providers/firestore.ts
@@ -43,6 +43,7 @@ export const defaultDatabase = '(default)';
 let firestoreInstance: any;
 export type DocumentSnapshot = firebase.firestore.DocumentSnapshot;
 export type QueryDocumentSnapshot = firebase.firestore.QueryDocumentSnapshot;
+export type DocumentData = firebase.firestore.DocumentSnapshot;
 
 /**
  * Select the Firestore document to listen to for events.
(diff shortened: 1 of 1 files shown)


morgs32:
@googlebot I signed it!

taeold:
Hey @morgs32. Thanks for taking the time to contribute.

While reviewing the code, I noticed that both `DocumentSnapshot` and `QuerySnapshot` types already default to `DocumentData`:

https://github.com/googleapis/nodejs-firestore/blob/28d645bd3e368abde592bfa2611de3378ca175a6/types/firestore.d.ts#L1121

https://github.com/googleapis/nodejs-firestore/blob/28d645bd3e368abde592bfa2611de3378ca175a6/types/firestore.d.ts#L1488

At least in VSCode, the typescript language server seems to make use of this type annotation:

Screen Shot 2021-06-01 at 8 08 47 PM

I'm pretty new to Typescript, and I'm unsure what features are unlocked when we make the proposed changes. Do you mind helping me understand by sharing an example or two?

morgs32:
This would enable something like this - in which I can easily type the snapshot of the onWrite. Let me know if this makes sense @taeold 

```
export const onWriteUserFirestoreHandler: onWrite<IUser> = async (props) => {

  const {
    after,
    // before
  } = props

  const userAfter = after.data()

  if (!userAfter) {
    return
  }

  // const userId = userAfter.id;

}

```

PR #905, opened 2021-06-23
Support documented case where FIREBASE_CONFIG is a json file.

### LINKED ISSUE #830: Error reading config file set in the FIREBASE_CONFIG environment variable as mentioned in the documentation
### [REQUIRED] Version info

**node:**

**firebase-functions:** 3.13.0

**firebase-tools:**

**firebase-admin:**

### [REQUIRED] Test case

Setup firebase config from a file path set in FIREBASE_CONFIG env variable

In the documentation [here ](https://firebase.google.com/docs/admin/setup#initialize-without-parameters) mentioned 
>  If the content of the FIREBASE_CONFIG variable begins with a { it will be parsed as a JSON object. Otherwise the SDK assumes that the string is the name of a JSON file containing the options.

But actual code is always treat the value as json object regardless of string starts with `{` or not

https://github.com/firebase/firebase-functions/blob/c69c9bc083a2ef55a3ea152839d6027b4bb20352/src/setup.ts#L41

Read and parse the file in the FIREBASE_CONFIG environment variable

Parse the path as json and throw invalid character at the position 0

inlined:
Per https://firebase.google.com/docs/admin/setup#initialize-without-parameters
the FIREBASE_CONFIG environment variable can be a name of a JSON file.

Because this now requires fs.readFileSync to keep API compliance,
firebaseConfg() now uses caching.

Closes #830


src/config.ts
```diff
@@ -20,9 +20,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-import * as firebase from 'firebase-admin';
+import * as fs from 'fs';
 import * as path from 'path';
```
joehan, line 25: remove newline

joehan (approved):
This LGTM, tho it looks like it needs npm format'ting

PR #906, opened 2021-06-24
Create package exports

inlined:
Update library to have proper package exports, which have a few advantages:

1. We now can import files without the "lib" prefix
2. We can now expose cleanly separated package versions irrespective of the underlying file layout

This change exports all previously documented entrypoints to the SDK: the root, `lib/logger` and `lib/logger/compat`. In addition, the logger entrypoints no longer need 'lib': `firebase-functions/logger` and `firebase-functions/logger/compat` now work as well. Finally, the entire existing SDK (aside from logger, which is versionless) can be accessed at `firebase-functions/v1`. We will start implementing v2 shortly.

This PR will break undefined behavior. Customers who previously imported sub-files like `firebase-functions/lib/providers/https` will no longer be able to do so. They can instead say `import {https} from 'firebase-functions'`. This is an already approved change and is not considered breaking, so there will be no major version bump.


joehan (approved):

package.json
```diff
@@ -23,6 +23,14 @@
   ],
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
+  "exports": {
```
taeold, line 26: I think we are missing `.js` extensions here, i.e.

```
"./logger": "./lib/logger.js",
"./lib/logger": "./lib/logger.js",
"./lib/logger/compat": "./lib/logger/compat.js"
```

inlined: I'll add them, though they're clearly not necessary since tests pass. AFAICT it just shortens the lookup process since it doesn't guess  lib/logger.js before the (new) lib/logger/index.js.

Issue #907, opened 2021-06-25
logger crashes on moment.js object ( TypeError: config._d.getTime is not a function ) 


carchrae:
### Related issues

arguably, this is a bug with moment.js - but i'm not certain who should fix : 
https://github.com/moment/moment/issues/5011

### [REQUIRED] Version info

**node:**  12 and 14

**firebase-functions:**  3.14.1

**firebase-tools:**  9.14.0

**firebase-admin:**  9.9.0
 

**moment:**  2.29.1

### [REQUIRED] Test case

```js
const functions = require('firebase-functions').logger;
const moment = require('moment');

console.log('this works', { m : moment() } );
console.log('this works too!', JSON.stringify({ m: moment() } ));

logger.info('wat!  this blows up!', { m : moment() } );
logger.info('but this works!', moment());
```

### [REQUIRED] Steps to reproduce

simply run the test case

### [REQUIRED] Expected behavior

logger should allow logging of moment object.  

### [REQUIRED] Actual behavior

note: this happens locally as well as when deployed

![image](https://user-images.githubusercontent.com/959825/123354474-932c0880-d518-11eb-8ca7-4529ee4912e0.png)

```js	
TypeError: config._d.getTime is not a function
    at new Moment (/workspace/node_modules/moment/moment.js:258:58)
    at Object.clone (/workspace/node_modules/moment/moment.js:3756:16)
    at Object.toISOString (/workspace/node_modules/moment/moment.js:3916:28)
    at Object.toJSON (/workspace/node_modules/moment/moment.js:4247:38)
    at JSON.stringify (<anonymous>)
    at write (/workspace/node_modules/firebase-functions/lib/logger.js:39:84)
    at Object.info (/workspace/node_modules/firebase-functions/lib/logger.js:81:5) 
```

### Were you able to successfully deploy your functions?

yes

carchrae:
fwiw, i suspect the problem is with `removeCircular` somehow destroying the moment object:  https://github.com/firebase/firebase-functions/blob/4028acc07d11cd81cd56ee816bf16f025c0d10df/src/logger.ts#L33

carchrae:
btw a workaround is to simply never try to log a moment object.  eg, always call `moment.format()` or `moment.toDate()`

i14h:
thanks @carchrae for reporting this. @joehan can you please take a look into this?

joehan:
[...]

PR #908, opened 2021-06-28
Defensively JSON.pase+JSON.stringify inputs into removeCircular to prevent NPEs

joehan:
### Description
Some complex objects that overwrote toJSON were breaking the structured logger. Specifically, calling removeCircular on some complex objects (for example, #907 ) could cause JSON.stringify to throw null pointer errors when called on those objects. This fix should ensure that any object can be safely logged.

### Testing
I tested out the case in #907, and I was able to successfully log `moment()` without errors:
``` {"structuredData":"2021-06-28T22:31:17.125Z","severity":"INFO","message":"Hello logs!"}```

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -30,7 +30,7 @@ export interface LogEntry {
   [key: string]: any;
 }
 
-function removeCircular(obj: any, refs:any[] = []): any {
+function removeCircular(obj: any, refs: any[] = []): any {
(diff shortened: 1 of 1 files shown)


joehan:
Seems like this broke some unit tests too, taking a look now.

joehan:
Welp, this approach won't work... For simple circular objects, JSON.stringify throws an error

PR #914, opened 2021-07-13
Fixes docgen after file renames.

mbleigh:
Recent file renames caused docgen to create lots of broken links. This PR does a few small things:

1. Fixes the links for docgen.
2. Changes the default for docgen to explicitly build the `v1` and `logger` folders since that's where all the things we want to document reside.
3. Improves the logic of the broken link warnings for docgen.
4. Marks `removeCircular` as internal because it was showing up in docs despite not being exported.

### DIFF
--- a/docgen/content-sources/toc.yaml
+++ b/docgen/content-sources/toc.yaml
@@ -1,148 +1,148 @@
 toc:
   - title: 'functions'
-    path: /docs/reference/functions/cloud_functions_.html
+    path: /docs/reference/functions/v1_cloud_functions_.html
     section:
       - title: 'CloudFunction'
-        path: /docs/reference/functions/cloud_functions_.html#cloudfunction
+        path: /docs/reference/functions/v1_cloud_functions_.html#cloudfunction
       - title: 'HttpsFunction'
-        path: /docs/reference/functions/cloud_functions_.html#httpsfunction
+        path: /docs/reference/functions/v1_cloud_functions_.html#httpsfunction
       - title: 'EventContext'
-        path: /docs/reference/functions/cloud_functions_.eventcontext.html
+        path: /docs/reference/functions/v1_cloud_functions_.eventcontext.html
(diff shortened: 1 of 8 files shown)


docgen/generate-docs.js
```diff
@@ -125,11 +127,13 @@ function addTypeAliasLinks(data) {
    * Select .tsd-signature-type because all potential external
    * links will have this identifier.
    */
-  const fileTags = htmlDom.window.document.querySelectorAll(".tsd-signature-type");
-  fileTags.forEach(tag => {
+  const fileTags = htmlDom.window.document.querySelectorAll(
+    '.tsd-signature-type'
+  );
+  fileTags.forEach((tag) => {
```
inlined, line 133: Nit: Google style guide prohibits `forEach` since it has no benefits over `for of`

docgen/content-sources/toc.yaml
```diff
@@ -1,148 +1,148 @@
 toc:
   - title: 'functions'
-    path: /docs/reference/functions/cloud_functions_.html
+    path: /docs/reference/functions/v1_cloud_functions_.html
```
inlined, line 3: These paths make me realize that we have a lot of issues here. Every namespace's "overview" page lists paths that are no longer importable. Will talk more offline.

inlined (approved):
An fix to the status quo though we might need more on top of this. LGTM.

docgen/generate-docs.js
mbleigh: Ah, this wasn't a line I touched just Prettier fixing up style, but easy enough to fix. Done for all instances in the file.

docgen/content-sources/toc.yaml
mbleigh: I've done a crude fix to this that changes the header to a simpler "last segment" instead. I think it functions better than the status quo.

PR #915, opened 2021-07-13
Moves most of https pacakge into /common/providers/https.

inlined:
The actual encoding/decoding/error logic etc. of callable functions
will stay the same between v1 and v2, so they should be in a /common
file.

The rest of /v1 is about trigger options, cors, etc.

I tried to get GitHub to pick up that most of this change is a cut + paste, but it doesn't seem to want to take the hint. We may just need to trust that much of the code can be glossed over.

onCall's CORS options are part of the v1 namespace because v2 adds
a "cors" option to HttpsOptions. In the v2 API, we will default
to `"cors": true`, but a customer could say `"cors": "my.firebase.app"`
and we will respect it.

I started this change while writing the v2 namespace as a whole and pulled in a few diffs from that. One notable change in behavior is that we set `__trigger.serviceAccount` to `null` when the option `default` was provided. This is a bugfix because previously setting `default` would do nothing to `__trigger.serviceAccount` and we wouldn't ever update the function accordingly on deploy. `firebase-tools` now has meaningful nulls and setting `serviceAccount` to null will cause it to be included in the update field mask, which will cause the value to be reset to the server-side default.


src/common/providers/https.ts
```diff
@@ -0,0 +1,581 @@
...
+}
+
+// The format for an HTTP body response from a Callable function.
+interface HttpResponseBody {
+  result?: any;
+  error?: HttpsError;
+}
+
+// Returns true if req is a properly formatted callable request.
+function isValidRequest(req: Request): req is HttpRequest {
```
mbleigh, line 309: Please re-add `@hidden` on all unexported functions, since typedoc is bad.

src/common/providers/https.ts
```diff
@@ -0,0 +1,581 @@
...
+      `Callable request verification failed: ${errs.join(' ')}`,
+      logPayload
+    );
+  }
+
+  return verifications;
+}
+
+/** @internal */
+export let onCallHandler = (
```
mbleigh, line 546: Is there a reason for `export let` vs. `export const`? Also why not just `export function`?

Lastly, I wonder if this should be a factory instead `onCallHandler({cors: Corsoptions})`, that way a critical piece of functionality (making the request pass cors) can't be accidentally forgotten in a future version.

src/common/proto.ts
```diff
@@ -0,0 +1,66 @@
...
+    return from;
+  }
+) {
+  if (!Object.prototype.hasOwnProperty.call(src, srcField)) {
+    return;
+  }
+  dest[destField] = converter(src[srcField]);
+}
+
+export function serviceAccountFromShorthand(
```
mbleigh, line 47: This doesn't really have anything to do with protos, so I wonder if this should be elsewhere or the file should be named something else.

mbleigh (approved):
There are some changes needed, but nothing that I have to see the result of before submitting.

PR #916, opened 2021-07-14
Fix logging for objects that define `toJSON`

### LINKED ISSUE #907: logger crashes on moment.js object ( TypeError: config._d.getTime is not a function )

joehan:
### Description
Fixes an issue where objects containing members that define their own `toJSON` function could not be logged correctly. Fixes #907 

### Code sample
The logger should now work as expected for cases like:
```logger.log("something!", {a: new Date()}```
```logger.log("something!", {a: moment()}```


inlined (approved):
🛳   it

PR #917, opened 2021-07-15
Finish refactor of HTTPS code

inlined:
Builds upon #915
    
Creates the first bits of v2/
    
`setGlobalOptions` allows customers to set options that will be preserved across any function written in the v2 API.

`v2/https` has webhook and callable function support. In addition to previous features, customers can now specifiy `cors` as a shorthand for using CORS middleware.
    
We will probably eventually support scheduled functions in the HTTPS namespace but that will have to wait for another day.


package.json
```diff
@@ -29,7 +29,29 @@
     "./logger": "./lib/logger/index.js",
     "./logger/compat": "./lib/logger/compat.js",
     "./lib/logger": "./lib/logger/index.js",
-    "./lib/logger/compat": "./lib/logger/compat.js"
+    "./lib/logger/compat": "./lib/logger/compat.js",
+    "./v2": "./lib/v2/index.js",
+    "./v2/options": "./lib/v2/options.js",
+    "./v2/https": "./lib/v2/providers/https.js"
+  },
+  "typesVersions": {
```
inlined, line 37: This fixes the submodule exports for typescript. TS imports don't actually support submodule exports yet, but typesVersion allows us to do a rewrite of the lookup paths for .d.ts files. The Admin and Web SDK use this hack.

src/v2/options.ts
```diff
@@ -0,0 +1,270 @@
...
+  serviceAccountFromShorthand,
+} from '../common/encoding';
+import * as logger from '../logger';
+import { TriggerAnnotation } from './base';
+import { copyIfPresent, convertIfPresent } from '../common/encoding';
+
+/**
+ * List of all regions supported by Cloud Functions v2
+ */
+export const SUPPORTED_REGIONS = ['us-west1'] as const;
```
joehan, line 34: TIL about const assertions

src/v2/options.ts
```diff
@@ -0,0 +1,270 @@
...
+
+let globalOptions: GlobalOptions | undefined;
+
+/**
+ * Sets default options for all functions written using the v2 SDK.
+ * @param options Options to set as default
+ */
+export function setGlobalOptions(options: GlobalOptions) {
+  if (globalOptions) {
+    logger.warn('Calling setGlobalOptions twice leads to undefined behavior');
```
joehan, line 193: Is this actually true? Looks pretty clear that it just overwrite all previous global options.

joehan (commented):
Looks mostly good to me - but why did you get rid of the entire `packages` section of package-lock.json?

src/v2/options.ts
inlined: I assume we can blame lauren for this one. I just copied the existing code from v1

src/v2/options.ts
inlined: Undefined doesn't mean unpredictable. I just want to set expectations that this is not an OK thing to do. The warning might also be interesting if someone is surprised that it's being set twice. E.g. If they have multiple files they're importing which each set global options. Or if some malicious codebase decided it would be funny to call `setGlobalOptions({minInstances: 100})`

inlined:
RE removing the packages section of `pacakge-lock.json`, this just happened because I ran `npm install` and it overwrote the `package-lock.json` with a v2 format. I can revert for now, but we'll have to face the v2 upgrade once we add new deps.

joehan (approved):

joehan:
Ahhh, I didn't realize - as long as it is still correct, I think its totally fine to include those changes in this PR. I didnt approve just to be sure it was something we wanted to do

PR #921, opened 2021-07-28
Set Access Control for HTTP Functions (v1)

colerogers:
### Description

Creating a way for customers to explicitly set the access control on HTTP triggered functions. The access control is set at the function level, so different functions can have different access policies. The customer can choose to set access control to `public`, `private`, or a custom service account. Access control is set in the `runWith()` function by setting the property `invoker`. To handle multiple service accounts, `invoker` can accept an array of strings.

The corresponding firebase-tools PR is [here](https://github.com/firebase/firebase-tools/pull/3619)

### Code sample

`export const myHttpFunc = runWith({invoker: "private"}).onRequest((req, res) => {
  res.send({hello: "private"});
});
`


src/v1/function-configuration.ts
```diff
@@ -98,6 +98,11 @@ export const DEFAULT_FAILURE_POLICY: FailurePolicy = {
 
 export const MAX_NUMBER_USER_LABELS = 58;
 
+/**
+ * Invoker access control type for https functions.
+ */
+export type Invoker = 'public' | 'private' | string;
```
joehan, line 104: A bit of a nit, but I prefer typing this as `'public' | 'private' | string | string[]`, and then making `RuntimeOptions.invoker: Invoker`
To me, this clarifies that something like`['private', 'public', 'my-service-acount@google.blah']` is not a valid option here.

src/v1/function-builder.ts
```diff
@@ -192,6 +192,34 @@ function assertRuntimeOptionsValid(runtimeOptions: RuntimeOptions): boolean {
       );
     }
   }
+
+  if (
+    runtimeOptions.invoker !== undefined &&
```
joehan, line 197: I think this is a bit of a redundant check - if runtimeOptions.invoker is undefined, then typeof runtimeOptions will equal undefined as well.

I think you can just get rid of this undefined check.

src/v1/function-builder.ts
```diff
@@ -192,6 +192,34 @@ function assertRuntimeOptionsValid(runtimeOptions: RuntimeOptions): boolean {
...
+    Array.isArray(runtimeOptions.invoker)
+  ) {
+    if (runtimeOptions.invoker.length == 0) {
+      throw new Error(
+        'Invalid invoker array, must contain at least 1 service account entry'
+      );
+    }
+    for (const serviceAccount of runtimeOptions.invoker) {
+      if (serviceAccount.length == 0) {
```
joehan, line 215: Consider also checking that serviceAccount != 'public' or 'private'  here, since those don't really make sense in the array context.

joehan (approved):
Some minor feedback, but this mostly LGTM

src/v1/function-configuration.ts
inlined: I (weakly) feel the opposite. One of the shorthands that was proposed was `'hosting'` and that one could make sense to be in the array.

src/v1/function-builder.ts
colerogers: Removed

src/v1/function-builder.ts
colerogers: Added checks and updated unit tests

src/v1/function-configuration.ts
colerogers: If we want to leave it as is for easier implementation of future work, I can do that. Alternatively, we can change it to Joe's suggestion and put a TODO comment when we add in the `hosting` feature. @joehan @inlined

inlined (approved):

src/v1/function-configuration.ts
joehan: Sounds like its 2 votes to 1 here, and I don't think anyone feels super strongly, so probably leave it as is

PR #924, opened 2021-07-29
Adds multiple api version support to docgen.

mbleigh:
This changes the `--source` flag that was previously used for docgen into a `--api` flag that takes either `v1` or `v2` as a value (defaulting to `v1`).

This allows us to have dual `content-sources` directories for each version and will hopefully give us what we need for the short term at least.

Also fixes a bug where since there are no classes in the v2 namespace it was erroring out.


egilmorez (approved):
Thanks Michael!

inlined (commented):
Surprisingly straightforward. Thanks!

PR #925, opened 2021-07-30
Add Pub/Sub support + HTTPS testing

inlined:
While adding Pub/Sub support I noticed that there were now v2/ tests.
Added those tests + fixes that they uncovered.

Also fixed issue where 'firebase-functions/logger' wasn't working as
an import in my test project.


spec/v2/providers/pubsub.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+      apiVersion: 2,
+      platform: 'gcfv2',
+      eventTrigger: EVENT_TRIGGER,
+      labels: {},
+    });
+  });
+
+  it('should create a complex trigger with appropraite values', () => {
+    const result = pubsub.onMessagePublished(
+      {
```
mbleigh, line 34: I wonder if we should extract out a common "kitchen sink" trigger that can be used across tests...

```ts
import { FULL_OPTIONS, FULL_TRIGGER } from "./helpers";

pubsub.onMessagePublished({ ...FULL_OPTIONS, topic: 'topic' });
expect(result.__trigger).to.deep.equal({ ...FULL_TRIGGER, eventTrigger: EVENT_TRIGGER });
```

src/v2/providers/pubsub.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/** Handle a message being published to a Pub/Sub topic. */
+export function onMessagePublished(
+  options: PubSubOptions,
+  callback: (event: CloudEvent<MessagePublishedData>) => any | Promise<any>
+): CloudFunction<MessagePublishedData>;
+
+export function onMessagePublished(
+  topicOrOptions: string | PubSubOptions,
+  callback: (event: CloudEvent<MessagePublishedData>) => any | Promise<any>
```
mbleigh, line 105: I'd prefer "handler" to "callback" I think

src/v2/providers/https.ts
```diff
@@ -44,12 +44,16 @@ export type HttpsHandler = (
...
-  data: any,
+export type CallableHandler<T> = (
+  data: T,
   context: CallableContext
 ) => any | Promise<any>;
 
 export type HttpsFunction = HttpsHandler & { __trigger: unknown };
+export interface CallableFunction<T> extends HttpsHandler {
+  __trigger: unknown;
+  run(data: T, context: CallableContext): any | Promise<any>;
```
mbleigh, line 55: I probably knew this at one point, but what's with `.run`?

inlined (commented):
Finally someone willing to stand up to me in code reviews! 😄

src/v2/providers/pubsub.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /** @hidden */
+  private _json: any;
+
+  constructor(data: any) {
+    this.messageId = data.messageId;
+    this.data = data.data;
+    (this.attributes = data.attributes || {}), (this._json = data.json);
+    this.orderingKey = data.orderingKey;
+    this.publishTime = data.publishTime || new Date().toISOString();
```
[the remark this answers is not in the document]
mbleigh: If we're confident it will never be missing in prod, this seems fine.

src/v2/providers/pubsub.ts
```diff
@@ -0,0 +1,158 @@
...
+): CloudFunction<MessagePublishedData> {
+  let topic: string;
+  let opts: options.EventHandlerOptions;
+  if (typeof topicOrOptions === 'string') {
+    topic = topicOrOptions;
+    opts = {};
+  } else {
+    topic = topicOrOptions.topic;
+    opts = { ...topicOrOptions };
+    delete (opts as any).topic;
```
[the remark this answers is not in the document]
mbleigh: I think we want to establish scalable patterns, and I'd rather be fully confident that unwanted fields are discarded by `optionsToTriggerAnnotations` than start a pattern of unnecessarily removing these fields. It's simple in this case, but when we do audit logs there may be 3-4 different event-specific options and it could become a bit much.

Not going to block approval on it, but that's my 2c

mbleigh (approved):

PR #926, opened 2021-07-31
Align DataSnapshot methods to @firebase/database DataSnapshot

rhodgkins:
### Description

I've noticed[1] a difference between `DataSnapshot`s methods in this SDK compared to the [one used when fetching data](https://github.com/firebase/firebase-js-sdk/tree/master/packages/database) (when using the admin SDK).

When setting values that are (or contain) empty object/arrays and null values using the admin SDK, when fetching these values back, they're always `null` when calling `.val()`.
However when creating a `DataSnapshot` with these values and calling `.val()` the results don't match.
I'm only raising this as an issue because when the functions are being called, I'd expect the `DataSnapshot` provided to the function handler to match the `DataSnapshot` when fetched from the same location - currently this isn't the case.

### Code sample

Using the following data: `{ child: 'foo', nullChild: null, emptyChild: {}, emptyArray: [], child: { nullVal: null, key: 'bar' } }`

Setting and then getting this value using the admin SDK [2] results in `{ child1: 'foo', child2: { key: 'bar } }` when calling `.val()`, which is what you'd expect - all the empty/`null` nodes are dropped.

However using the `firebase-functions` `database.DataSnapshot` constructor the result includes these empty/`null` values `{ child: 'foo', nullChild: null, emptyChild: {}, emptyArray: {}, child: { nullVal: null, key: 'bar' } }` when calling `.val()`.

I think this is mostly due to the lack of [this](https://github.com/firebase/firebase-js-sdk/blob/8599d91416ae8ac5202742f11cee00666d3360ec/packages/database/src/core/snap/ChildrenNode.ts#L198-L200) check the `@firebase/database` does when calling `.val()`.

The discrepancies are mainly when using `.val()` but I've updated the other methods and their tests (`.exists()`, `.numChildren()`, `.hasChildren()`, etc.) for completeness... 

[1] noticed when using `firebase-functions-test` [`makeDataSnapshot`](https://github.com/firebase/firebase-functions-test/blob/5361ad3757529e8da0b5f0356657b3236fcadbe9/src/providers/database.ts#L29-L52) method (which calls this repos `database.DataSnapshot` constructor) for mocking parameters for testing a Firebase Function.

[2] By using:

```js
const firebase = require('firebase-admin')
firebase.initializeApp({ databaseURL: 'https://<databaseName>.firebaseio.com' })

const setAndGet = async val => {
   const ref = firebase.database().ref('node')
   
   await ref.set(val)
   return await ref.get()
}
```

### DIFF
--- a/spec/v1/providers/database.spec.ts
+++ b/spec/v1/providers/database.spec.ts
@@ -541,16 +541,51 @@ describe('Database Functions', () => {
         expect(subject.val()).to.equal(0);
         populate({ myKey: 0 });
         expect(subject.val()).to.deep.equal({ myKey: 0 });
-
-        // Null values are still reported as null.
-        populate({ myKey: null });
-        expect(subject.val()).to.deep.equal({ myKey: null });
       });
 
       // Regression test: .val() was returning array of nulls when there's a property called length (BUG#37683995)
       it('should return correct values when data has "length" property', () => {
         populate({ length: 3, foo: 'bar' });
         expect(subject.val()).to.deep.equal({ length: 3, foo: 'bar' });
+      });
+
+      it('should deal with null-values appropriately', () => {
+        populate(null);
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: null });
+        expect(subject.val()).to.be.null;
+      });
+
+      it('should deal with empty object values appropriately', () => {
+        populate({});
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: {} });
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: { child: null } });
+        expect(subject.val()).to.be.null;
+      });
+
+      it('should deal with empty array values appropriately', () => {
+        populate([]);
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: [] });
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: [null] });
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: [{}] });
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: [{ myKey: null }] });
+        expect(subject.val()).to.be.null;
+
+        populate({ myKey: [{ myKey: {} }] });
+        expect(subject.val()).to.be.null;
       });
     });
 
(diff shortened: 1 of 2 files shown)


spec/v1/providers/database.spec.ts
```diff
@@ -541,17 +541,52 @@ describe('Database Functions', () => {
         expect(subject.val()).to.equal(0);
         populate({ myKey: 0 });
         expect(subject.val()).to.deep.equal({ myKey: 0 });
-
-        // Null values are still reported as null.
-        populate({ myKey: null });
-        expect(subject.val()).to.deep.equal({ myKey: null });
```
rhodgkins, line 547: I've removed this test case as its incorrect compared to the Firebase SDK, the correct version is now in `should deal with null-values appropriately`

src/v1/providers/database.ts
```diff
@@ -598,6 +612,11 @@ export class DataSnapshot {
       }
     }
 
+    if (numKeys === 0) {
+      // Empty node
+      return null;
+    }
+
```
rhodgkins, line 619: This simulates the `this.isEmpty()` [check the Firebase SDK does](https://github.com/firebase/firebase-js-sdk/blob/8599d91416ae8ac5202742f11cee00666d3360ec/packages/database/src/core/snap/ChildrenNode.ts#L198-L200).

spec/v1/providers/database.spec.ts
```diff
@@ -653,13 +722,51 @@ describe('Database Functions', () => {
...
+      it('should be false for non-objects', () => {
+        populate(23);
+        expect(subject.hasChildren()).to.be.false;
+
+        populate({
+          nullChild: null,
+          emptyObjectChild: {},
+          emptyArrayChild: [],
+        });
+        expect(subject.hasChildren()).to.be.false;
```
rhodgkins, line 769: `.hasChildren()` had no tests so add some by mirroring the `.numChildren()` tests above.

rhodgkins:
@inlined due to https://github.com/firebase/firebase-functions/pull/943 this PR needs a bit of an update - shall I rebase off master and do it against the old non-v1 directory structure?

inlined:
Let me get back to you. We got rid of the v1 directory temporarily to unbreak the emulator and people depending on undocumented behavior. We want to move everything back ASAP by releasing a breaking change. That breaking change might actually be the best time to accept your change since it's a good change but changes behavior notably.

rhodgkins:
> Let me get back to you. We got rid of the v1 directory temporarily to unbreak the emulator and people depending on undocumented behavior. We want to move everything back ASAP by releasing a breaking change. That breaking change might actually be the best time to accept your change since it's a good change but changes behavior notably.

Any update on this @inlined ?

inlined:
Closing this since the changes have been ported to the `launch.next` branch and will be included in the v4 SDK. Thank you very much for both your technical contribution and your patience.

admin-01-web:
[gh pr checkout 926]

admin-01-web:
- [ ]  @admin-01-web

admin-01-web:
#1298

PR #928, opened 2021-08-02
Adds "params" for environment variable helpers and type coercion

mbleigh:
As per review in other branch.


inlined (approved):
Looks like there's some crazy diffs going on here, but as long as it merges cleanly, LGTM

PR #930, opened 2021-08-03
Print a useful error explaining why Runtime Config doesn't work

inlined:
What it says on the box


src/v1/config.ts
```diff
@@ -26,6 +26,14 @@ import * as path from 'path';
 import * as firebase from 'firebase-admin';
 
 export function config(): config.Config {
+  // K_CONFIGURATION is only set in GCFv2
+  if (process.env.K_CONFIGURATION) {
+    throw new Error(
+      'RuntimeConfig is no longer available in Google Cloud Functions v2. ' +
```
mbleigh, line 32:
```suggestion
      'functions.config() is no longer available in Cloud Functions for Firebase v2. ' +
```

we never actually called it  "Runtime Config" other than the filename, and only Firebase has it at all

mbleigh (approved):

spec/v1/config.spec.ts
```diff
@@ -47,6 +47,23 @@ describe('config()', () => {
     (config as any).firebaseConfigCache = null;
     delete process.env.FIREBASE_CONFIG;
     delete process.env.CLOUD_RUNTIME_CONFIG;
+    delete process.env.K_CONFIGURATION;
+  });
+
+  it('Will never load in GCFv2', () => {
```
taeold, line 53:
```suggestion
  it('will never load in GCFv2', () => {
```

taeold (approved):

PR #933, opened 2021-08-05
Bug bash fixes

inlined:
1. Always return promises in HTTPS (should help emulator)
2. Export undocumented path that the emulator monkeypatches (sad panda)
3. Make @types/express a production dependency

Can test with https://storage.googleapis.com/firebase-preview-drop/node/firebase-functions/firebase-functions-3.14.2-rc1.tgz


src/common/providers/https.ts
```diff
@@ -598,9 +598,12 @@ export function onCallHandler<Req = any, Res = any>(
...
 ): (req: Request, res: express.Response) => Promise<void> {
   const wrapped = wrapOnCallHandler(handler);
-  return async (req: Request, res: express.Response) => {
-    cors(options)(req, res, () => {
-      return wrapped(req, res);
+  return (req: Request, res: express.Response) => {
+    return new Promise((resolve) => {
+      res.on('finish', resolve);
+      cors(options)(req, res, () => {
+        resolve(wrapped(req, res));
```
mbleigh, line 605: As written, I think this promise is going to get resolved twice in the situation where CORS passes, and should it potentially be rejecting instead of resolving if CORS halts the request?

You might want something like:

```suggestion
      let resolved = false;
      res.on('finish', () => {
        if (!resolved) {
          resolved = true;
          resolve();
        }
      });
      cors(options)(req, res, () => {
        resolved = true;
        resolve(wrapped(req, res));
```

inlined: Promises are defined to only resolve once and be idempotent if called multiple times.

mbleigh (approved):

PR #935, opened 2021-08-10
Set Access Control for HTTP Functions (v2)

colerogers:
### Description

Creating a way for customers to explicitly set the access control on v2 HTTP functions. The access control is set at the function level, so different functions can have different access policies. The customer can choose to set access control to `public`, `private`, or a custom service account. Access control is set in the `onRequest({})` options object, by setting the invoker property. To handle multiple service accounts, invoker can accept an array of strings.

### Code sample

`export const myFunction = v2.https.onRequest({ invoker: 'private' }, (req, res) => { res.send("Done"); });`

### DIFF
--- a/spec/v2/providers/https.spec.ts
+++ b/spec/v2/providers/https.spec.ts
@@ -82,7 +82,6 @@ describe('onRequest', () => {
         allowInsecure: false,
       },
       labels: {},
-      invoker: 'public',
     });
   });
 
@@ -91,7 +90,7 @@ describe('onRequest', () => {
       {
         ...FULL_OPTIONS,
         region: ['us-west1', 'us-central1'],
-        invoker: ['service-account1', 'service-account2'],
+        invoker: ['service-account1@', 'service-account2@'],
       },
       (req, res) => {
         res.send(200);
@@ -103,7 +102,7 @@ describe('onRequest', () => {
         allowInsecure: false,
       },
       regions: ['us-west1', 'us-central1'],
-      invoker: ['service-account1', 'service-account2'],
+      invoker: ['service-account1@', 'service-account2@'],
     });
   });
 
@@ -112,6 +111,7 @@ describe('onRequest', () => {
       concurrency: 20,
       region: 'europe-west1',
       minInstances: 1,
+      invoker: 'public',
     });
 
     const result = https.onRequest(
@@ -135,7 +135,7 @@ describe('onRequest', () => {
       minInstances: 3,
       regions: ['us-west1', 'us-central1'],
       labels: {},
-      invoker: 'private',
+      invoker: ['private'],
     });
   });
 
--- a/spec/v2/providers/https.spec.ts
+++ b/spec/v2/providers/https.spec.ts
@@ -82,6 +82,7 @@ describe('onRequest', () => {
         allowInsecure: false,
       },
       labels: {},
+      invoker: 'public',
     });
   });
 
@@ -90,6 +91,7 @@ describe('onRequest', () => {
       {
         ...FULL_OPTIONS,
         region: ['us-west1', 'us-central1'],
+        invoker: ['service-account1', 'service-account2'],
       },
       (req, res) => {
         res.send(200);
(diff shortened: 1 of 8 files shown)


spec/v2/providers/https.spec.ts
```diff
@@ -90,6 +91,7 @@ describe('onRequest', () => {
       {
         ...FULL_OPTIONS,
         region: ['us-west1', 'us-central1'],
+        invoker: ['service-account1', 'service-account2'],
```
inlined, line 94: Don't these need to end with '@'?

src/v2/providers/https.ts
```diff
@@ -32,12 +32,18 @@ export type CallableRequest<T = any> = common.CallableRequest<T>;
...
+ */
+export type Invoker = 'public' | 'private' | string;
+
 export interface HttpsOptions extends Omit<options.GlobalOptions, 'region'> {
   region?:
     | options.SupportedRegion
     | string
     | Array<options.SupportedRegion | string>;
   cors?: string | boolean;
+  invoker?: Invoker | Invoker[];
```
inlined, line 46: @mbleigh Should we make this an HTTPS only option or should we make it also an event trigger option and set the service account of EventArc accordingly? I guess maybe the former since we don't have a clear answer what to do if the user passes two SAs.

src/v2/providers/https.ts
```diff
@@ -113,6 +120,7 @@ export function onRequest(
         httpsTrigger: {
           allowInsecure: false,
         },
+        invoker,
```
inlined, line 123: Let's omit if it's not set and always make it an array if set.

src/v2/providers/https.ts
mbleigh: Yeah, I think let's leave this HTTPS-only for now.

spec/v2/providers/https.spec.ts
colerogers: In the CLI, I've wrote a helper function to take either `service-account`, `service-account@`, or `service-account@project-id.iam.gserviceaccount.com`

src/v2/providers/https.ts
```diff
@@ -81,6 +87,7 @@ export function onRequest(
     opts = optsOrHandler as HttpsOptions;
   }
 
+  const invoker = 'invoker' in opts ? opts.invoker : 'public';
```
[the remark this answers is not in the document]
inlined: I think that's a fine TS signature though the CLI should handle the valid-but-wtf case

spec/v2/providers/https.spec.ts
colerogers: Updated the unit test to use service accounts that end with a '@'

src/v2/providers/https.ts
colerogers: Removed here, added it to global options, and added a `convertIfPresent()` call to the change invoker to an array.

src/v2/providers/https.ts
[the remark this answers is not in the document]
colerogers: I think that all 3 of these cases should be treated as bad and should throw an error. The way I read the type definition is that the user can set `'public'`, `'private'`, `'$service-account'`, or `[ '$service-account-1', '$service-account-2', ... ]`. They should not mix `public` or `private` with the array of service accounts. I've added a helper method to `src/common/encoding` that will error out of the user sets `public` or `private` in the array.

src/v2/options.ts
```diff
@@ -182,6 +188,11 @@ export interface GlobalOptions {
    * User labels to set on the function.
    */
   labels?: Record<string, string>;
+
+  /**
+   * Invoker to set access control on https functions.
+   */
+  invoker?: Invoker | Invoker[];
```
inlined, line 195: Given your new defensive programming stance I agree with Daniel now that we can/should change our typing.

```
invoker?: 'public' | 'private' | string | string[];
```

This will make sure code complete doesn't suggest "public" or "private" when the customer enters an array.

inlined (approved):

src/v2/options.ts
colerogers: Changed the type definition

PR #936, opened 2021-08-10
[apidocs] Improve file detection and updates toc.yaml

mbleigh:


inlined (approved):

docgen/generate-docs.js
```diff
@@ -198,7 +198,10 @@ async function checkForMissingFilesAndFixFilenameCase(tocText) {
   const filenames = tocText
     .split('\n')
     .filter((line) => line.includes('path:'))
-    .map((line) => line.split(devsitePath)[1].replace(/#.*$/, ''));
+    .map((line) => {
```
egilmorez, line 201: Wow I would have had no idea how to refactor that.

Do we need to update the value for 'const devsitePath' as well? Or does this basically override that and extrapolate it from the toc.yaml file (which makes a lot of sense, since that is in a versioned directory that can be updated as we go)?

egilmorez (approved):
Thanks!

One question on const devsitePath.

docgen/generate-docs.js
mbleigh: The only other thing `devsitePath` feeds is `_toc_autogenerated.yaml` which I'm not sure if you use. That being said, I now fork devsitePath based on version.

PR #955, opened 2021-08-16
Fix bug where v1 HTTP used v2 API if callback had len 1

### LINKED ISSUE #947: http.onCall now sends raw header and body in the first parameter
### [REQUIRED] Version info

**node:**
v14.17.3

**firebase-functions:**
3.15.2

**firebase-tools:**
9.10.0

**firebase-admin:**
9.11.0

### [REQUIRED] Test case
```ts
// functions/index.ts

export const myCallable = functions.region("europe-west1").https.onCall(async data => {
  console.log(data);
});```

// myApp/api.ts
import firebase from "firebase/app";
import "firebase/functions";

const firebaseApp = firebase.initializeApp(eval(process.env.VUE_APP_FIREBASE_CONFIG!!));
const functions = firebaseApp.functions("europe-west1");
const callMyFunction = functions.httpsCallable('myCallable');

const result = "";
callMyFunction({text: "Some text"}).then(r => result = r);
```

Call HTTP function from WebApp.

Expected to see `{text: "Some text"}` in firebase console -> functions -> logs

Screenshot 2021-08-13 at 18 00 04
It outputs a very long log of what seems to be header + body of a whole request.

Functions were deployed successfully without any error messages.

inlined:
The Callable API is complex enough that we share code between the v1 and v2 SDKs. The shared code sniffed the callback length to determine which API to surface, but this is a mistake because the callback is customer controlled and they may drop parameters they don't care about. By wrapping the customer-provided handler with a known-sized handler we can fix the sniffing behavior.

Fixes #947

### DIFF
--- a/spec/v1/providers/https.spec.ts
+++ b/spec/v1/providers/https.spec.ts
@@ -179,20 +179,23 @@ describe('#onCall', () => {
   // Regression test for firebase-functions#947
   it('should lock to the v1 API even with function.length == 1', async () => {
     let gotData: Record<string, any>;
-    const func = https.onCall(data => {
+    const func = https.onCall((data) => {
       gotData = data;
     });
 
-    const req = new MockRequest({
-      data: {"foo": "bar"},
-    }, {
-      "content-type": "application/json",
-    });
-    req.method = "POST";
+    const req = new MockRequest(
+      {
+        data: { foo: 'bar' },
+      },
+      {
+        'content-type': 'application/json',
+      }
+    );
+    req.method = 'POST';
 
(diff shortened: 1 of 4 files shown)


spec/v1/providers/https.spec.ts
```diff
@@ -175,6 +175,25 @@ describe('#onCall', () => {
     };
     expect(cf.run(data, context)).to.deep.equal({ data, context });
   });
+
+  // Regression test for firebase-functions#947
```
taeold, line 179: 👍

taeold (approved):

PR #959, opened 2021-08-21
fix: rawBody is optional type in funcs framework

jthegedus:
### Description

Working on the [Firebase Adapter](https://github.com/jthegedus/svelte-adapter-firebase) for [SvelteKit](kit.svelte.dev) I discovered the type here is different from that in the Functions Framework which I believe is an upstream of this?

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/e76b9f6c49bcf8c12d8206212cae84cfb931a077/src/invoker.ts#L44

### DIFF
--- a/src/common/providers/https.ts
+++ b/src/common/providers/https.ts
@@ -32,7 +32,7 @@ import { apps } from '../../apps';
 
 /** @hidden */
 export interface Request extends express.Request {
-  rawBody: Buffer;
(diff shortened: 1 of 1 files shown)


jthegedus:
Do these unit tests require an approver to run, or is the CI in flux and the "required" tag for each not applicable?

PR #963, opened 2021-08-29
Added missing export: 'lib/function-configuration'

Benglin:
### Description

When I upgraded from `firebase-functions@3.14.1` to `3.15.4`, I could no longer deploy function with this command:

```sh
$ firebase deploy --only functions:myFunction
```

The errors are:

```
i  deploying functions
i  functions: ensuring required API cloudfunctions.googleapis.com is enabled...
i  functions: ensuring required API cloudbuild.googleapis.com is enabled...
v  functions: required API cloudbuild.googleapis.com is enabled
v  functions: required API cloudfunctions.googleapis.com is enabled

Error: Error occurred while parsing your function triggers.

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/function-configuration' is not defined by "exports" in /home/ben/repos/my-project/src/functions/node_modules/firebase-functions/package.json

    at throwExportsNotFound (internal/modules/esm/resolve.js:299:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:522:3)
    ...
```

### Code sample

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "./lib/logger": "./lib/logger/index.js",
     "./lib/providers/auth": "./lib/providers/auth.js",
     "./lib/providers/analytics": "./lib/providers/analytics.js",
+    "./lib/function-configuration": "./lib/function-configuration.js",
     "./lib/providers/database": "./lib/providers/database.js",
     "./lib/providers/firestore": "./lib/providers/firestore.js",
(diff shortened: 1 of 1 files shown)


taeold:
@Benglin Can you describe your usecase? I believe that `./lib/function-configuration` intentionally not exposed (it's an internal file).

PR #972, opened 2021-09-07
Fixes v1 doc structure

mbleigh:
Since we reverted doc structure for v1, this should generate docs successfully though I haven't vetted thoroughly if it catches everything we need/expect.

### DIFF
--- a/docgen/content-sources/v1/toc.yaml
+++ b/docgen/content-sources/v1/toc.yaml
@@ -1,148 +1,148 @@
 toc:
   - title: 'functions'
-    path: /docs/reference/functions/v1_cloud_functions_.html
+    path: /docs/reference/functions/cloud_functions.html
     section:
       - title: 'CloudFunction'
-        path: /docs/reference/functions/v1_cloud_functions_.html#cloudfunction
+        path: /docs/reference/functions/cloud_functions.html#cloudfunction
       - title: 'HttpsFunction'
-        path: /docs/reference/functions/v1_cloud_functions_.html#httpsfunction
+        path: /docs/reference/functions/cloud_functions.html#httpsfunction
       - title: 'EventContext'
-        path: /docs/reference/functions/v1_cloud_functions_.eventcontext.html
+        path: /docs/reference/functions/cloud_functions.EventContext.html
       - title: 'FunctionBuilder'
-        path: /docs/reference/functions/v1_function_builder_.functionbuilder.html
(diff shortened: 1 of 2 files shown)


docgen/content-sources/v1/toc.yaml
```diff
@@ -1,148 +1,148 @@
 toc:
   - title: 'functions'
-    path: /docs/reference/functions/v1_cloud_functions_.html
```
egilmorez, line 3: This is a surprise to me.  The currently live reference docs don't use this versioned path (they're just https://firebase.google.com/docs/reference/functions/cloud_functions_#cloudfunction, etc).

Also, in the past I believe we manually created this file and then generated the usable, publishable _toc.yaml file from it.  Did you devise a way to autogenerate content-sources/toc.yaml?

docgen/content-sources/v1/toc.yaml
```diff
@@ -1,148 +1,148 @@
...
+    path: /docs/reference/functions/cloud_functions.html
```
egilmorez, line 3: Filenames currently live on firesite still use the underscores, such as cloud_functions_#cloudfunction and config_.config.

If we're 100% sure the script will generate the files without underscore naming, this should work fine.

egilmorez (commented):
Couple of questions for your mbleigh, thanks!

LMK here or offline whether there's a way I can "sync" to your PR and give this a test drive.

docgen/content-sources/v1/toc.yaml
mbleigh: No, I manually corrected this based on the output of the script in this branch.

docgen/content-sources/v1/toc.yaml
mbleigh: This was modified to match the output of the gen script.

mbleigh (commented):
Assuming `origin` is the public functions SDK repo:

```
git fetch origin
git checkout -b mb-docdir origin/mb-docdir
```

egilmorez:
Let me know when/if it's time to take another look and try another run Michael. Thanks!

egilmorez:
Hi Michael!

Since these issues were all resolved separately, can you close/drop this PR?

Until you do, I think I'm going to keep getting a bot bug in my queue.

taeold:
@egilmorez Think https://github.com/firebase/firebase-functions/pull/990 supersedes this PR, so I'll close it!

PR #980, opened 2021-09-20
Adjust acceptable options values

inlined:
Add support for new options in v2

### DIFF
--- a/src/v2/providers/https.ts
+++ b/src/v2/providers/https.ts
@@ -40,7 +40,7 @@ export interface HttpsOptions extends Omit<options.GlobalOptions, 'region'> {
     | options.SupportedRegion
     | string
     | Array<options.SupportedRegion | string>;
-  cors?: string | boolean | RegExp | (string | RegExp)[];
+  cors?: string | boolean | RegExp | Array<string | RegExp>;
 }
 
 export type HttpsFunction = ((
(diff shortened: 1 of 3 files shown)


src/v2/providers/https.ts
```diff
@@ -40,7 +40,7 @@ export interface HttpsOptions extends Omit<options.GlobalOptions, 'region'> {
     | options.SupportedRegion
     | string
     | Array<options.SupportedRegion | string>;
-  cors?: string | boolean | RegExp | (string | RegExp)[];
+  cors?: string | boolean | RegExp | Array<string | RegExp>;
 }
```
inlined, line 44: Looks like this line was added by `npm fix:lint`

src/v2/providers/https.ts
```diff
@@ -40,7 +40,7 @@ export interface HttpsOptions extends Omit<options.GlobalOptions, 'region'> {
     | options.SupportedRegion
     | string
     | Array<options.SupportedRegion | string>;
-  cors?: string | boolean | RegExp | (string | RegExp)[];
+  cors?: string | boolean | RegExp | Array<string | RegExp>;
```
colerogers, line 43: Just curious, why do we prefer this syntax here?

When I touched this area, the underlying origin property of the cors library uses the bracket notation

colerogers (approved):

src/v2/providers/https.ts
taeold: Commented below, `npm fix:lint` did it for us.

taeold (approved):

PR #981, opened 2021-09-23
GCS Enhancement

colerogers:
### Description

Enhancement to how customers use GCS Triggered Functions.

### Code sample

```
onObjectArchived()
onObjectFinalized()
onObjectDeleted()
onObjectMetadataUpdated()
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
     "./v2/options": "./lib/v2/options.js",
     "./v2/https": "./lib/v2/providers/https.js",
     "./v2/params": "./lib/v2/params/index.js",
-    "./v2/pubsub": "./lib/v2/providers/pubsub.js"
+    "./v2/pubsub": "./lib/v2/providers/pubsub.js",
+    "./v2/storage": "./lib/v2/providers/storage.js"
   },
   "typesVersions": {
     "*": {
@@ -110,6 +111,9 @@
       ],
       "v2/pubsub": [
         "lib/v2/providers/pubsub"
+      ],
+      "v2/storage": [
+        "lib/v2/providers/storage"
       ]
     }
   },
--- a/spec/v2/providers/storage.spec.ts
+++ b/spec/v2/providers/storage.spec.ts
@@ -5,21 +5,35 @@ import * as options from '../../../src/v2/options';
 import * as storage from '../../../src/v2/providers/storage';
 import { FULL_OPTIONS, FULL_TRIGGER } from './helpers';
 
+const EVENT_TRIGGER = {
+  eventType: 'event-type',
+  resource: 'some-bucket',
+  service: storage.service,
+};
+
 describe('v2/storage', () => {
   describe('_getOptsAndBucket', () => {
-    it('should return throw error without a bucket', () => {
-      expect(() => storage._getOptsAndBucket()).to.throw();
-    });
-
-    it('should return the default bucket on no params', () => {
+    it('should return the default bucket with empty opts', () => {
       const configStub = sinon
         .stub(config, 'firebaseConfig')
         .returns({ storageBucket: 'default-bucket' });
 
-      const [opts, bucket] = storage._getOptsAndBucket();
+      const [opts, bucket] = storage._getOptsAndBucket({});
 
       configStub.restore();
       expect(opts).to.deep.equal({});
+      expect(bucket).to.eq('default-bucket');
+    });
+
+    it('should return the default bucket with opts param', () => {
+      const configStub = sinon
+        .stub(config, 'firebaseConfig')
+        .returns({ storageBucket: 'default-bucket' });
+
+      const [opts, bucket] = storage._getOptsAndBucket({ region: 'us-west1' });
+
+      configStub.restore();
+      expect(opts).to.deep.equal({ region: 'us-west1' });
       expect(bucket).to.eq('default-bucket');
     });
 
@@ -42,12 +56,6 @@ describe('v2/storage', () => {
   });
 
   describe('_onOperation', () => {
-    const EVENT_TRIGGER = {
-      eventType: 'google.cloud.storage.event-type',
-      resource: 'my-bucket',
-      service: 'storage.googleapis.com',
-    };
-
     beforeEach(() => {
       options.setGlobalOptions({});
       process.env.GCLOUD_PROJECT = 'aProject';
@@ -58,8 +66,70 @@ describe('v2/storage', () => {
       delete process.env.GCLOUD_PROJECT;
     });
 
-    it('should create a minimal trigger', () => {
-      const result = storage._onOperation(() => 42, 'event-type', 'my-bucket');
+    it('should create a minimal trigger default bucket', () => {
+      const configStub = sinon
+        .stub(config, 'firebaseConfig')
+        .returns({ storageBucket: 'default-bucket' });
+
+      const result = storage._onOperation('event-type', () => 42);
+
+      configStub.restore();
+      expect(result.__trigger).to.deep.equal({
+        apiVersion: 2,
+        platform: 'gcfv2',
+        labels: {},
+        eventTrigger: {
+          ...EVENT_TRIGGER,
+          resource: 'default-bucket',
+        },
+      });
+    });
+
+    it('should create a minimal trigger with bucket', () => {
+      const result = storage._onOperation(
+        'event-type',
+        'some-bucket',
+        () => 42
+      );
+
+      expect(result.__trigger).to.deep.equal({
+        apiVersion: 2,
+        platform: 'gcfv2',
+        labels: {},
+        eventTrigger: EVENT_TRIGGER,
+      });
+    });
+
+    it('should create a minimal trigger with opts', () => {
+      const configStub = sinon
+        .stub(config, 'firebaseConfig')
+        .returns({ storageBucket: 'default-bucket' });
+
+      const result = storage._onOperation(
+        'event-type',
+        { region: 'us-west1' },
+        () => 42
+      );
+
+      configStub.restore();
+      expect(result.__trigger).to.deep.equal({
+        apiVersion: 2,
+        platform: 'gcfv2',
+        labels: {},
+        eventTrigger: {
+          ...EVENT_TRIGGER,
+          resource: 'default-bucket',
+        },
+        regions: ['us-west1'],
+      });
+    });
+
+    it('should create a minimal trigger with bucket with opts and bucket', () => {
+      const result = storage._onOperation(
+        'event-type',
+        { bucket: 'some-bucket' },
+        () => 42
+      );
 
       expect(result.__trigger).to.deep.equal({
         apiVersion: 2,
@@ -70,10 +140,14 @@ describe('v2/storage', () => {
     });
 
     it('should create a complex trigger with appropriate values', () => {
-      const result = storage._onOperation(() => 42, 'event-type', {
-        ...FULL_OPTIONS,
-        bucket: 'my-bucket',
-      });
+      const result = storage._onOperation(
+        'event-type',
+        {
+          ...FULL_OPTIONS,
+          bucket: 'some-bucket',
+        },
+        () => 42
+      );
 
       expect(result.__trigger).to.deep.equal({
         ...FULL_TRIGGER,
@@ -88,11 +162,15 @@ describe('v2/storage', () => {
         minInstances: 1,
       });
 
-      const result = storage._onOperation(() => 42, 'event-type', {
-        bucket: 'my-bucket',
-        region: 'us-west1',
-        minInstances: 3,
-      });
+      const result = storage._onOperation(
+        'event-type',
+        {
+          bucket: 'some-bucket',
+          region: 'us-west1',
+          minInstances: 3,
+        },
+        () => 42
+      );
 
       expect(result.__trigger).to.deep.equal({
         apiVersion: 2,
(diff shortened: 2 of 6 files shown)


src/v2/providers/storage.ts
```diff
@@ -0,0 +1,419 @@
...
+      event: CloudEvent<StorageObjectData>
+    ) => any | Promise<any>;
+    _onOperation(handler, archivedEvent);
+  }
+  return _onOperation(
+    handler,
+    archivedEvent,
+    optsOrHandler as string | StorageOptions
+  );
+}
```
taeold, line 80: Hmmm, it looks like we repeat this block for each of the four event type. It would be cool if we can contain all "make sure we handle `optOrHandler` appropriately in `_onOperation`; can we start a discussion in our chat room to see if there's a clever TS thing we can do?

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  return [opts, bucket];
+}
+
+/**
+ * An object within Google Cloud Storage.
+ * Ref: https://github.com/googleapis/google-cloudevents-nodejs/blob/main/cloud/storage/v1/StorageObjectData.ts
+ */
+export interface StorageObjectData {
```
taeold, line 277: nit* Can this be declared at the top of the file instead?

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,419 @@
...
+        platform: 'gcfv2',
+        ...baseOpts,
+        ...specificOpts,
+        labels: {
+          ...baseOpts?.labels,
+          ...specificOpts?.labels,
+        },
+        eventTrigger: {
+          eventType: provider + '.' + eventType,
+          resource: bucket, // TODO(colerogers): replace with bucket: 'my-bucket' when container contract is finished
```
taeold, line 232: Can you expand on what you mean by this TODO? I'm a confused by the suggestion that we should set this to a constant `my-bucket`. Also probably better to leave out mentioning container contract here.

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,419 @@
...
+  } else {
+    bucket = bucketOrOpts.bucket || firebaseConfig().storageBucket;
+    opts = { ...bucketOrOpts };
+    delete (opts as any).bucket;
+  }
+
+  if (!bucket) {
+    throw new Error(
+      'Missing bucket name. If you are unit testing, please provide a bucket name' +
+        ' through set process.env.FIREBASE_CONFIG.'
```
taeold, line 263:
```suggestion
        'Missing bucket name. If you are unit testing, please provide a bucket name' +
          ' by providing bucket name directly in the event handler or by setting process.env.FIREBASE_CONFIG.'
      );
```

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,419 @@
...
+): CloudFunction<StorageObjectData>;
+
+/** Handle a storage object metadata updated */
+export function onObjectMetadataUpdated(
+  opts: StorageOptions,
+  handler: (event: CloudEvent<StorageObjectData>) => any | Promise<any>
+): CloudFunction<StorageObjectData>;
+
+export function onObjectMetadataUpdated(
+  optsOrHandler:
```
taeold, line 174: nits* It's a mouthful, but I think I would prefer `butcketOrOptsOrHandler` for accuracy.

spec/v2/providers/storage.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  describe('_onOperation', () => {
+    const EVENT_TRIGGER = {
+      eventType: 'google.cloud.storage.event-type',
+      resource: 'my-bucket',
+      service: 'storage.googleapis.com',
+    };
+
+    beforeEach(() => {
+      options.setGlobalOptions({});
```
taeold, line 52: This seems redundant no?

spec/v2/providers/storage.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+      process.env.GCLOUD_PROJECT = 'aProject';
```
taeold, line 53: Hmm can you remind me why setting this environment variable is required for the test to work?

spec/v2/providers/storage.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+      expect(bucket).to.eq('my-bucket');
+    });
+  });
+
+  describe('_onOperation', () => {
+    const EVENT_TRIGGER = {
+      eventType: 'google.cloud.storage.event-type',
+      resource: 'my-bucket',
+      service: 'storage.googleapis.com',
+    };
```
taeold, line 49: nit* I'd make this module-level constant instead of scoping it to this test.

spec/v2/providers/storage.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+        platform: 'gcfv2',
+        labels: {},
+        eventTrigger: {
+          eventType: 'google.cloud.storage.object.v1.archived',
+          resource: 'my-bucket',
+          service: 'storage.googleapis.com',
+        },
+        regions: ['us-west1'],
+      });
+    });
```
taeold, line 162: Can we also test where opt is provided without a `bucket` property?

taeold (commented):
Minor nits. Hoping we can find a way to reduce some repetition in code.

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,419 @@
...
+/** @hidden */
+export const service = 'storage.googleapis.com';
+/** @hidden */
+export const archivedEvent = 'object.v1.archived';
+/** @hidden */
+export const finalizedEvent = 'object.v1.finalized';
+/** @hidden */
+export const deletedEvent = 'object.v1.deleted';
+/** @hidden */
+export const metadataUpdatedEvent = 'object.v1.metadataUpdated';
```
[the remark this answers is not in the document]
colerogers: I'm using them in the unit tests

src/v2/providers/storage.ts
colerogers: I've cleaned it up and moved the logic into `_onOperation`. The 4 functions now only serve to inject the correct event type.

src/v2/providers/storage.ts
colerogers: Moved to the top

src/v2/providers/storage.ts
colerogers: Good catch, I removed the CC part. Eventually, we'll need to remove the resource field and just map to a bucket field.

src/v2/providers/storage.ts
colerogers: Fixed

src/v2/providers/storage.ts
colerogers: Makes sense, fixed

spec/v2/providers/storage.spec.ts
colerogers: If we don't set it, then when we try and perform `optionsToTriggerAnnotations` it will call `serviceAccountFromShorthand` which requires this env var to be set and throw an error otherwise

spec/v2/providers/storage.spec.ts
colerogers: Moved to the top and added in specific event constants for each event type.

spec/v2/providers/storage.spec.ts
colerogers: Added tests for each event type tests and the onOperation tests.

spec/v2/providers/storage.spec.ts
colerogers: removed

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,389 @@
...
+export interface StorageOptions extends options.EventHandlerOptions {
+  bucket?: string;
+}
+
+/** Handle a storage object archived */
+export function onObjectArchived(
+  handler: (event: CloudEvent<StorageObjectData>) => any | Promise<any>
+): CloudFunction<StorageObjectData>;
+
+/** Handle a storage object archived */
```
taeold, line 197: ugh does ESLint for us to repeat JSDoc comment on each of the overloaded function 🤦‍♂️ ?

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,389 @@
...
+  handler?: (event: CloudEvent<StorageObjectData>) => any | Promise<any>
+): CloudFunction<StorageObjectData> {
+  return _onOperation(metadataUpdatedEvent, buketOrOptsOrHandler, handler);
+}
+
+/** @hidden */
+export function _onOperation(
+  // handler: (event: CloudEvent<StorageObjectData>) => any | Promise<any>,
+  // eventType: string,
+  // bucketOrOpts?: string | StorageOptions
```
taeold, line 304:
```suggestion
```

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,389 @@
...
+  eventType: string,
+  bucketOrOptsOrHandler:
+    | string
+    | StorageOptions
+    | ((event: CloudEvent<StorageObjectData>) => any | Promise<any>),
+  handler?: (event: CloudEvent<StorageObjectData>) => any | Promise<any>
+): CloudFunction<StorageObjectData> {
+  if (arguments.length === 2) {
```
taeold, line 312: Isn't there a subtle bug here since `_onOperation` is always called with 3 arguments now?

e.g.
```js
> function foo(a, b) { return arguments.length }
undefined
> foo(1)
1
> foo(1, 2)
2
> foo(1, undefined)
2
```

So I'd expect `arguments.length` to always be `3`

taeold: Maybe this will work:

```
if (typeof bucketOrOptsOrHandler === "function") {
  ...
}
```

src/v2/providers/storage.ts
colerogers: I was following what we did with pubsub, but since GCS has a lot more overloads, I removed the repeated ones and left the top comment.

src/v2/providers/storage.ts
colerogers: Thanks, removed

src/v2/providers/storage.ts
colerogers: Oh dang, yeah that's a bug alright. For some reason in my head passing an undefined handler was equivalent to leaving off the last parameter. Thanks for catching this!

taeold (approved):
lgtm; I'd recommend we wait until @inlined review as this is our first major enhancement on v2 provider namespace outside of his code.

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,379 @@
...
+        },
+      };
+    },
+  });
+
+  return func;
+}
+
+/** @hidden */
+export function _getOptsAndBucket(
```
inlined, line 354: While you'll still find `_` prefixes in old codebases, that was pre access-control. Also, I think you can use @internal instead of @hidden (would need to double check that it doesn't show up in docs).

inlined (approved):
Sorry... I reviewed this a while ago and never hit "submit"

package.json
```diff
@@ -110,6 +111,9 @@
       ],
       "v2/pubsub": [
         "lib/v2/providers/pubsub"
+      ],
+      "v2/storage": [
+        "lib/v2/providers/storage"
```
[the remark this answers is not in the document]
colerogers: Thanks, created /v2/storage.js

spec/v2/providers/storage.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+      expect(opts).to.deep.equal({ region: 'us-west1' });
+      expect(bucket).to.eq('my-bucket');
+    });
+  });
+
+  describe('_onOperation', () => {
+    const EVENT_TRIGGER = {
+      eventType: 'google.cloud.storage.event-type',
+      resource: 'my-bucket',
+      service: 'storage.googleapis.com',
```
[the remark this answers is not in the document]
colerogers: Removed

spec/v2/providers/storage.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+    it('should accept only handler', () => {
+      const configStub = sinon
+        .stub(config, 'firebaseConfig')
+        .returns({ storageBucket: 'default-bucket' });
+
+      const result = storage.onObjectArchived(() => 42);
+
+      configStub.restore();
+      expect(result.__trigger).to.deep.equal({
+        apiVersion: 2,
```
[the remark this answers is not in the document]
colerogers: Removed

src/v2/providers/storage.ts
```diff
@@ -0,0 +1,379 @@
...
+   */
+  encryptionAlgorithm?: string;
+  /**
+   * SHA256 hash value of the encryption key.
+   */
+  keySha256?: string;
+}
+
+/** @hidden */
+export const service = 'storage.googleapis.com';
```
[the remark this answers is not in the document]
colerogers: Removed

src/v2/providers/storage.ts
colerogers: Removed the `_` prefixes and changed to using `@internal`. The `typedoc` docs indicate that both hidden and internal will remove from the code from documentation as long as we don't pass in `--excludeInternal`.

PR #983, opened 2021-09-29
Allow requests from the emulator to bypass token verifications

taeold:
To replace monkey-patching of the Firebase Functions SDK in the Functions Emulator ([code](https://github.com/firebase/firebase-tools/blob/c2feb0836f6f64236e117f2906ef6083840e212b/src/emulator/functionsEmulatorRuntime.ts#L401-L496)), we provide native support for bypassing token verification for `onCall` handlers for both v1 and v2 namespaces.

Auth/App Check token included in the request will be decoded but no verified under following conditions:

1. FUNCTIONS_EMULATOR environment variable exists.
2. Request's host is set to `localhost`.

These conditions will make it difficult to accidentally trigger "skip verification" mode outside of the emulator environment.

Note that we add dependency to `jws` package ([~10KB](https://bundlephobia.com/package/jws@3.0.0)). This is unappealing, but I don't think this will make a big dent on user code.


package.json
```diff
@@ -133,12 +133,14 @@
...
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "jws": "^3.0.0",
     "lodash": "^4.17.14"
   },
   "devDependencies": {
     "@types/chai": "^4.1.7",
     "@types/chai-as-promised": "^7.1.0",
     "@types/jsonwebtoken": "^8.3.2",
+    "@types/jws": "^3.2.4",
```
colerogers, line 143: I'm just curious, what's the process for adding a new package into the SDK?

spec/common/providers/https.spec.ts
```diff
@@ -446,10 +476,11 @@ describe('onCallHandler', () => {
   });
 
   it('should reject bad AppCheck token', async () => {
+    const projectId = appsNamespace().admin.options.projectId;
+    const appId = '1:65211879909:web:3ae38ef1cdcb2e01fe5f0c';
```
colerogers, line 480: Is this fake/should we be exposing this here?

colerogers (approved):
lgtm, but I'd hold off merging until you get another set of eyes on it

spec/common/providers/https.spec.ts
inlined: Though this is vaguely PII it doesn't come with any security privileges. Think of this as hard-coding the name of your test project in code. We probably already did so with our mock credentials as well.

src/common/providers/https.ts
```diff
@@ -500,6 +502,53 @@ interface CallableTokenStatus {
...
+  }
+  return payload;
+}
+
+/**
+ * Decodes Auth ID token.
+ *
+ * This is exposed only for testing.
+ */
+/** @hidden */
```
inlined, line 528: Let's use @internal to avoid letting other packages depend on this (in TypeScript at least..)

spec/fixtures/mockrequest.ts
```diff
@@ -85,6 +87,24 @@ export function generateIdToken(projectId: string): string {
   return jwt.sign(claims, mockKey.private_key, options);
 }
 
+/**
+ * Generates a mocked, unsigned Firebase ID token.
+ */
+export function generateUnsignedIdToken(projectId: string): string {
```
inlined, line 93: Let's talk tomorrow. We could hypothetically build and parse invalid JWTs without the `jws` library. Right now, you're creating totally valid JWTs where the signature is signed with secret ''. If you wanted to create a JWT that just didn't have a valid signature you could do:

```
[
  {alg: 'HS256', typ: 'JWT'},
  { aud: projectId, sub: mokKey.user_id},
  "Invalid signature"
]
  .map(str => JSON.stringify(str))
  .map(str => Buffer.from(str).toString('base64'))
  .join('.')
```

And to parse without validating, we'd say:

```
JSON.parse(Buffer.from(token.split('.')[1], "base64").toString());
```

inlined (commented):
Excited!

package.json
taeold: Great question - I haven't thought of it and I haven't been flagged by my lack of insight. I'll follow internal documentations to audit the dependencies I'm adding.

spec/common/providers/https.spec.ts
taeold: For safe measures, I made the `appId` obviously fake.

src/common/providers/https.ts
taeold: Fixed!

spec/fixtures/mockrequest.ts
taeold: This works!! Thanks for the suggestion - able to remove `jws` from our dep.

spec/common/providers/https.spec.ts
```diff
@@ -133,6 +143,56 @@ async function runTest(test: CallTest): Promise<any> {
...
+  expect(context.app).to.not.be.null;
+  expect(context.app.appId).to.equal(appId);
+  expect(context.app.token.app_id).to.be.equal(appId);
+  expect(context.app.token.sub).to.be.equal(appId);
+  expect(context.app.token.aud).to.be.deep.equal([`projects/${projectId}`]);
+  expect(context.auth).to.be.undefined;
+  expect(context.instanceIdToken).to.be.undefined;
+}
+
+function checkAuthRequest(request: CallableRequest, projectId: string) {
```
inlined, line 171: nit: personally I feel a bit cleaner if we pass the mock user as a parameter rather than checkAuthRequest hard coding that it checks against a particular user. Not something I'd make a stink over though since we use the test fixture pretty regularly though.

spec/fixtures/mockrequest.ts
```diff
@@ -85,6 +85,20 @@ export function generateIdToken(projectId: string): string {
...
+  return [
+    { alg: 'RS256', typ: 'JWT' },
+    { aud: projectId, sub: mockKey.user_id },
+    'Invalid signature',
+  ]
+    .map((str) => JSON.stringify(str))
```
inlined, line 97: tip: Up to you if you think this is too dense or neat, but you can also say `.map(JSON.stringify)`

src/common/providers/https.ts
```diff
@@ -500,6 +504,55 @@ interface CallableTokenStatus {
   auth: TokenStatus;
 }
 
+function decodeToken(token: string): unknown {
```
inlined, line 507: I really like the "unsafe" prefix in other places. Maybe we should mark this the same as well so future authors don't blindly export & use it.

src/common/providers/https.ts
```diff
@@ -500,6 +504,55 @@ interface CallableTokenStatus {
...
+  }
+  return payload;
+}
+
+/**
+ * Decode, but not verify, a Auth ID token.
+ *
+ * Do not use in production. Token should always be verified using the Admin SDK.
+ *
+ * This is exposed only for testing.
```
inlined, line 531: And for debug mode/emulator, right?

src/common/providers/https.ts
```diff
@@ -527,12 +584,18 @@ async function checkTokens(
...
+        appCheckData = { appId: decodedToken.app_id, token: decodedToken };
+      } else {
+        appCheckData = await apps()
+          .admin.appCheck()
+          .verifyToken(appCheck);
+      }
       ctx.app = {
-        appId: appCheckToken.appId,
-        token: appCheckToken.token,
+        appId: appCheckData.appId,
```
inlined, line 597: I feel like I'm misreading something here. Why are we using `appCheckData` as a temporary instead of just saving to `ctx.app`?

spec/fixtures/mockrequest.ts
taeold: woo functional

src/common/providers/https.ts
taeold: I think the comment is correct - this function will only be called internally. It will be executed in debug mode but will never be invoked directly.

src/common/providers/https.ts
taeold: good point!

spec/common/providers/https.spec.ts
taeold: That's a fair point. Updated.

PR #985, opened 2021-10-05
Tries to fix v1 docs.

mbleigh:


egilmorez:
As we noted in offline discussions:

1) To make the v1 TOC work as usual, we'd need to change back the v2-ish values in https://github.com/firebase/firebase-functions/blob/master/docgen/theme/layouts/default.hbs

2) Filenames linked from index.html generated with this PR don't match the actual HTML filenames.  I think/hope that could be fixed in the script, but am not sure exactly how.

Values for 1) would be:

 <meta name="book_path" value="/docs/reference/_book.yaml" />
 <meta name="gtm_var" data-key="docType" data-value="reference">

Thanks!

PR #986, opened 2021-10-05
Allow users to specify options for callable functions.

taeold:
We extend `functions.https.onCall` function builder to accept an (optional) options object that expose users more control over callable function's behaviors.

2 options are exposed:

1. `cors`: Allows users to control `Access-Control-Allow-Origin` to override default behavior (`cors: { origin: true }`).
2. `allowInvalidAppCheckToken`: Allows users to control behavior of callable function when given invalid App Check tokens. By default, requests with invalid App Check tokens are rejected prior to executing user code.

e.g.

```js

export const helloWorld = functions
    .https
    .onCall({allowInvalidAppCheckToken: true}, (data, context) => {
      // context.app will be undefined if the request doesn't include a valid
      // App Check token.
      if (context.app == undefined) {
        throw new functions.https.HttpsError(
            "failed-precondition",
            "The function must be called from an App Check verified app.");
      }
      return `Hello ${data.name}`;
    });
```

### DIFF
--- a/src/function-builder.ts
+++ b/src/function-builder.ts
@@ -23,7 +23,12 @@
 import * as express from 'express';
 import * as _ from 'lodash';
 
-import { CloudFunction, EventContext, HttpsFunction, Runnable } from './cloud-functions';
+import {
+  CloudFunction,
+  EventContext,
+  HttpsFunction,
+  Runnable,
+} from './cloud-functions';
 import {
   DeploymentOptions,
   INGRESS_SETTINGS_OPTIONS,
@@ -44,7 +49,6 @@ import * as pubsub from './providers/pubsub';
 import * as remoteConfig from './providers/remoteConfig';
 import * as storage from './providers/storage';
 import * as testLab from './providers/testLab';
-
 
 /**
  * Assert that the runtime options passed in are valid.
--- a/src/function-builder.ts
+++ b/src/function-builder.ts
@@ -23,7 +23,7 @@
 import * as express from 'express';
 import * as _ from 'lodash';
 
-import { CloudFunction, EventContext } from './cloud-functions';
+import { CloudFunction, EventContext, HttpsFunction, Runnable } from './cloud-functions';
 import {
   DeploymentOptions,
   INGRESS_SETTINGS_OPTIONS,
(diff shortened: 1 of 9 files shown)


src/function-builder.ts
```diff
@@ -284,6 +290,29 @@ export function runWith(runtimeOptions: RuntimeOptions): FunctionBuilder {
   }
 }
 
+function _onCall<T = any, Return = any | Promise<any>>(
```
inlined, line 293: Aside, I'm somewhat disgruntled that we still have `_` prefixes in our codebase. It was the style when programming with Closure, but is not part of the style guide anymore. I'm not sure wether it's worse to propagate the old style or have divergent styles.

src/providers/https.ts
```diff
@@ -46,14 +46,37 @@ export function onRequest(
   return _onRequestWithOptions(handler, {});
 }
 
+export interface CallableV1Options {
```
inlined, line 49: We're in the V1 namespace so we don't need to put V1 in the type here. It'll especially draw attention to a V2 that's private when people read the docs.

src/function-builder.ts
```diff
@@ -347,14 +376,10 @@ export class FunctionBuilder {
...
+       * @param opts Configuration options for a callable function.
        * @param handler A method that takes a data and context and returns a value.
        */
-      onCall: (
-        handler: (
-          data: any,
-          context: https.CallableContext
-        ) => any | Promise<any>
-      ) => https._onCallWithOptions(handler, this.options),
+      onCall: _onCall.bind(this),
```
inlined, line 382: I'm not sure I feel comfortable with a free-floating function that has a "this" method that must be used with "bind". Why can't this just be an instance method?

inlined (commented):
I know you've gone through the work already to build this in the v2-style API, but I'm wondering if that's the right approach once I see it all together.

1. Per HTTPS function CORS options is a v2 thing. We _could_ add it to v1, but it's not a bad thing to offer carrots in v2. It can also be omitted if we think it complicates the API surface too much
2. It occurs to me that allowing invalid app check tokens is likely a global option that applies to all (HTTPS) functions. Maybe it should be a `runWith` option?

What do you think of a simpler approach where we just add `allowInvalidAppCheckToken` in `DeployOptions` for v1?

PR #990, opened 2021-10-13
Fix docgen script

taeold:
Notable changes:

* We revert back prefixing v1 docs with `v1_`.
* Dangling `_` suffix is removed, e.g. `cloud_functions_` => `cloud_functions`
* Bunch more documents not in the ToC is created.
* Some odd file naming, e.g. `config.config-1.html` instead of `config.config1.html`
* Some type definiitions moved under `common/` directory and hence the filename changed, e.g. `providers_https` is now `common_providers_https`.
* Added more pages to provders.https namesapce (CallbleRequest, AppCheckData, AuthData).


egilmorez (approved):
Generated content looks good!  Thanks for taking this on Daniel.

PR #992, opened 2021-10-14
Add debug mode

taeold:
Debug mode is intended to be used by the Functions Emulator to enable/disable certain aspect of the Functions SDK. Debug mode allows the Functions SDK to operate in development friendly way without the need to monkey-patch the Functions SDK. For example, auth or App Check token verification can be disabled during development to make Functions SDK compatible with the Auth Emulator.

This feature is intended for internal use-cases and won't expose any public interfaces. No one outside the Firebase team should use it, and if they do, we can't promise any stability or feature support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - GCS Enhancement
+- Add option to allow callable functions to ignore invalid App Check tokens.
--- a/spec/common/providers/https.spec.ts
+++ b/spec/common/providers/https.spec.ts
@@ -40,6 +40,8 @@ interface CallTest {
 
   callableFunction2: (request: https.CallableRequest<any>) => any;
 
+  callableOption?: https.CallableOptions;
+
   // The expected shape of the http response returned to the callable SDK.
   expectedHttpResponse: RunHandlerResult;
 }
@@ -104,7 +106,10 @@ function runHandler(
 
 // Runs a CallTest test.
 async function runTest(test: CallTest): Promise<any> {
-  const opts = { origin: true, methods: 'POST' };
+  const opts = {
+    cors: { origin: true, methods: 'POST' },
+    ...test.callableOption,
+  };
   const callableFunctionV1 = https.onCallHandler(opts, (data, context) => {
     expect(data).to.deep.equal(test.expectedData);
     return test.callableFunction(data, context);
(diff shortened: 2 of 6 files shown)


colerogers (approved):
Looks cool 😎

src/debug.ts
```diff
@@ -0,0 +1,52 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// Do NOT turn on a debug feature in production.
+const debugMode = process.env.FIREBASE_DEBUG_MODE === 'true';
```
colerogers, line 24: Could we also check that FIREBASE_DEBUG_FEATURES exists? Is having two env vars just used for some security?

colerogers (approved):
Looks cool 😎

src/debug.ts
taeold: Yeah I thought extra, repeated step makes it slightly harder to accidentally invoke a debug mode.

Another thing I was trying to do here is to make sure that we minimize the number of environment variable read. Apparently, reading env var is not the cheapest thing in Nodejs, so this guard ensures that we issue env var per request only when the debug mode is turned on.

src/debug.ts
```diff
@@ -0,0 +1,52 @@
...
+
+interface DebugFeatures {
+  skipCallableTokenVerification?: boolean;
```
inlined, line 27: I believe we also will have a disableTimeouts, right?

src/debug.ts
```diff
@@ -0,0 +1,52 @@
+// The MIT License (MIT)
```
inlined, line 1: I wonder if this should go into src/common since most of the stuff under src/ will go into v1 in the next major change.

PR #999, opened 2021-10-27
Annotate function triggers with __endpoint property

taeold:
In addition to annotating function triggers with `__trigger` property, we add `__endpoint` annotation. This property will be used by the to-be-developed functions runtime to generate/declare deployment manifest that the CLI will use to deploy the function.

There are lots of code duplication between the utility functions for annotating the `__trigger` and `__endpoint` properties. I didn't try to refactor the common code since I expect that we will favor `__endpoint` property in the future.


src/cloud-functions.ts
```diff
@@ -37,6 +37,7 @@ import {
   durationFromSeconds,
   serviceAccountFromShorthand,
 } from './common/encoding';
+import { ManifestEndpoint } from './common/manifest/v1alpha1';
```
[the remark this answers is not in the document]
taeold: That make sense. Unrolled the dir.

src/v2/providers/https.ts
```diff
@@ -130,6 +132,38 @@ export function onRequest(
       return trigger;
     },
   });
+
+  Object.defineProperty(handler, '__endpoint', {
+    get: () => {
```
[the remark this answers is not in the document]
taeold: Yes for storage, that will be a problem. Will prefer to plainly assign unless necessary like the storage trigger's case.

spec/v2/providers/https.spec.ts
```diff
@@ -221,10 +249,19 @@ describe('onCall', () => {
         'deployment-callable': 'true',
       },
     });
+
+    expect(result.__endpoint).to.deep.equal({
+      platform: 'gcfv2',
+      httpsTrigger: {},
+      labels: {
+        'deployment-callable': 'true',
```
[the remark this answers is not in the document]
taeold: Offline discussion happened - we are going with `callableTrigger`.

spec/v2/providers/pubsub.spec.ts
```diff
@@ -3,13 +3,21 @@ import { expect } from 'chai';
...
 
 const EVENT_TRIGGER = {
   eventType: 'google.cloud.pubsub.topic.v1.messagePublished',
   resource: 'projects/aProject/topics/topic',
 };
 
+const ENDPOINT_EVENT_TRIGGER = {
+  eventType: 'google.cloud.pubsub.topic.v1.messagePublished',
+  eventFilters: {
+    resource: 'projects/aProject/topics/topic',
```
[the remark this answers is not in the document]
taeold: Yep! I learned that cross-project pubsub is not a thing :/.

src/v2/providers/https.ts
```diff
@@ -46,7 +47,7 @@ export interface HttpsOptions extends Omit<options.GlobalOptions, 'region'> {
 export type HttpsFunction = ((
   req: Request,
   res: express.Response
-) => void | Promise<void>) & { __trigger: unknown };
+) => void | Promise<void>) & { __trigger: unknown } & { __endpoint: unknown };
```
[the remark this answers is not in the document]
taeold: sounds good. also made similar changes in core.ts.

src/v2/providers/pubsub.ts
```diff
@@ -164,5 +166,25 @@ export function onMessagePublished<T = any>(
...
+      ...specificOpts?.labels,
+    },
+    eventTrigger: {
+      eventType: 'google.cloud.pubsub.topic.v1.messagePublished',
+      eventFilters: { topic },
+      retry: false,
+    },
+  };
+  copyIfPresent(endpoint.eventTrigger, opts, 'retry', 'retry');
+  func.__endpoint = endpoint;
```
colerogers, line 187: We seem to be alternating between using `Object.defineProperty()` and directly setting the property in the v2 providers. Would it be beneficial to stick with one format for all the providers?

src/v2/providers/storage.ts
```diff
@@ -341,6 +344,34 @@ export function onOperation(
...
+        ...baseOpts,
+        ...specificOpts,
+        labels: {
+          ...baseOpts?.labels,
+          ...specificOpts?.labels,
+        },
+        eventTrigger: {
+          eventType: eventType,
+          eventFilters: {
+            bucket,
```
colerogers, line 365: Yay! So much easier to add new filters 🎆

src/cloud-functions.ts
```diff
@@ -322,9 +333,9 @@ export interface MakeCloudFunctionArgs<EventData> {
   dataConstructor?: (raw: Event) => EventData;
   eventType: string;
   handler?: (data: EventData, context: EventContext) => PromiseLike<any> | any;
-  labels?: { [key: string]: any };
+  labels?: Record<string, string>;
```
colerogers, line 336: Is there a risk that some projects might have non-string values here? If so, are we releasing this as a breaking change?

src/cloud-functions.ts
```diff
@@ -322,9 +333,9 @@ export interface MakeCloudFunctionArgs<EventData> {
...
   legacyEventType?: string;
-  options?: { [key: string]: any };
+  options?: DeploymentOptions;
```
colerogers, line 338: Same question as above, what if projects don't have `options` that fit into `DeploymentOptions`?

colerogers (commented):
Left a few questions, also do you want to add a CHANGELOG entry for these changes?

src/cloud-functions.ts
inlined: The backend never accepted anything but a string.

src/cloud-functions.ts
```diff
@@ -545,3 +587,51 @@ export function optionsToTrigger(options: DeploymentOptions) {
...
+  copyIfPresent(
+    endpoint,
+    options,
+    'minInstances',
+    'maxInstances',
+    'ingressSettings',
+    'labels',
+    'timeoutSeconds'
+  );
+  convertIfPresent(endpoint, options, 'region', 'regions', (r) => r);
```
inlined, line 604: nit: IIRC the transform function defaults to identity and can be removed.

src/cloud-functions.ts
```diff
@@ -545,3 +587,51 @@ export function optionsToTrigger(options: DeploymentOptions) {
...
+  );
+  convertIfPresent(endpoint, options, 'vpc', 'vpcConnector', (connector) => {
+    return { connector };
+  });
+  convertIfPresent(
+    endpoint,
+    options,
+    'vpc',
+    'vpcConnectorEgressSettings',
+    (egressSettings, vpc) => {
```
inlined, line 620: This is confusing to me. Is this an edit in place as an optional second parameter? I'm wondering if a more explicit approach will be clearer since now convertIfPresent only works when the calls are in the right order.

Maybe something vaguely like:

```
const vpc: <whateverthetypeishere> = {};
convertIfPresent(vpc, options, 'connector', 'vpcConnector');
convertIfPresent(vpc, options, 'egressSettings', 'vpcConnectorEgressSettings');
if (Object.keys(vpc).length !== 0) {
  endpoint.vpc = vpc;
}
```

src/common/manifest.ts
```diff
@@ -0,0 +1,83 @@
...
+  environmentVariables?: Record<string, string>;
+
+  httpsTrigger?: {
+    invoker?: string[];
+  };
+
+  callableTrigger?: {};
+
+  eventTrigger?: {
+    eventFilters: {
```
inlined, line 51: I suspect this should just be `Record<string, string>` instead of an explicit list of all known filters

src/providers/https.ts
```diff
@@ -77,6 +82,19 @@ export function _onRequestWithOptions(
...
+    platform: 'gcfv1',
+    ...optionsToEndpoint(options),
+    httpsTrigger: {},
+  };
+  convertIfPresent(
+    cloudFunction.__endpoint.httpsTrigger,
+    options,
+    'invoker',
+    'invoker',
+    convertInvoker
```
inlined, line 96: doesn't `convertInvoker` turn `foo@` into `foo@project.gserviceaccount.com` and we didn't want that?

src/v2/providers/pubsub.ts
```diff
@@ -133,7 +135,7 @@ export function onMessagePublished<T = any>(
 
   func.run = handler;
 
-  // TypeScript doesn't recongize defineProperty as adding a property and complains
+  // TypeScript doesn't recognize defineProperty as adding a property and complains
   // that __trigger doesn't exist. We can either cast to any and lose all type safety
   // or we can just assign a meaningless value before calling defineProperty.
   func.__trigger = 'silence the transpiler';
```
inlined, line 141: We can just cut these lines now that __trigger is optional

src/v2/providers/pubsub.ts
inlined: I have mixed feels. We could always use define property because it's safest, but it is a bit arcane and is only needed for endpoint types that rely on firebaseConfig (RTDB and Storage)

inlined (approved):

src/cloud-functions.ts
taeold: So my perception is - yes this is breaking but uses who used anything other than string would not have succeeded in deploying a function with non-string. This seems okay to do.

src/cloud-functions.ts
taeold: Another good point. Fortunately, no one should be calling `makeCloudFunction` directly, and in our public APIs, we make sure that passed options are typed with `DeploymentOption`. e.g.

https://github.com/firebase/firebase-functions/blob/38f1946df4804002be9c585f386c5ae61fd45b10/src/providers/pubsub.ts#L75

Think we should be okay making this change.

src/v2/providers/pubsub.ts
taeold: I've been leaning towards "don't use `defineProperty()` unless we have to. And we have to use `defineProperty()` in some cases b/c user code will throw an error if environment variables like `FIREBASE_CONFIG` does not exist (which we use to define a trigger, like default bucket info).

src/cloud-functions.ts
taeold: That's fair. I agree that the suggested code is simpler!

src/providers/https.ts
taeold: We have that kind of logic for service accounts:

https://github.com/firebase/firebase-functions/blob/38f1946df4804002be9c585f386c5ae61fd45b10/src/common/encoding.ts#L53-L72

But not for invoker - we only do light validation checks, e.g. to make sure that invoker option with `public` or `private` doesn't conflict, etc.

inlined: Let's keep the shorthand syntax in the manifest so we can stay project-independent.

PR #1003, opened 2021-11-09
Firebase Functions ships with a binary

taeold:
We introduce a bin associated with the firebase-functions SDK! The binary exposes a simple server on localhost that describes the Firebase Functions defined in the current directory:

```
$ STACK_CONTROL_API_PORT=8181 npx firebase-functions  # Optionally can pass in a FUNCTIONS_DIR as first arg 
Serving at port 8181

$ curl localhost:8181/__/stack.yaml | jq
{
  "endpoints": {
    "onreqv2": {
      "platform": "gcfv2",
      "labels": {},
      "httpsTrigger": {},
      "entryPoint": "onreqv2"
    },
    "onRequest": {
      "platform": "gcfv1",
      "httpsTrigger": {},
      "entryPoint": "onRequest"
    }
  },
  "specVersion": "v1alpha1",
  "requiredAPIs": []
}
```


spec/fixtures/sources/commonjs-grouped/g1.js
```diff
@@ -0,0 +1,5 @@
+const functions = require("../../../../src/index");
+
+exports.v1callable = functions.https.onCall((_data, _ctx) => {
```
inlined, line 3: nit: if you're not going to use either parameter, why take any?

spec/fixtures/sources/commonjs-grouped/index.js
```diff
@@ -0,0 +1,7 @@
+const functions = require("../../../../src/index");
+
+exports.v1http = functions.https.onRequest((_req, resp) => {
```
inlined, line 3: While I get that you're prefixing params with _ to indicate they're unused, that seems to violate style guides for naming.

spec/runtime/loader.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+      {
+        name: 'scheduled',
+        modulePath: './spec/fixtures/sources/commonjs-scheduled',
+        expected: {
+          endpoints: {
+            scheduled: {
+              platform: 'gcfv1',
+              entryPoint: 'scheduled',
+              labels: {
+                'deployment-scheduled': 'true',
```
inlined, line 166: I don't think we want to require this in the manifest. We'll just add it in the deployment process as necessary (that way we keep authoring a stack.yaml easy)

src/bin/firebase-functions.ts
```diff
@@ -0,0 +1,50 @@
...
+const args = process.argv.slice(2);
+if (args.length != 1) {
+  printUsage();
+}
+const functionsDir = args[0];
+
+let server;
+const app = express();
+
+app.get('/backend.yaml', async (req, res) => {
```
inlined, line 27: `/__/stack.yaml`

src/bin/firebase-functions.ts
```diff
@@ -0,0 +1,50 @@
...
+    res.setHeader('content-type', 'text/yaml');
+    res.send(JSON.stringify(backend));
+  } catch (e) {
+    res
+      .status(400)
+      .send(`Failed to generate manifest from function source: ${e}`);
+  }
+});
+
+app.get('/quitquitquit', async (req, res) => {
```
inlined, line 39: `/__/quitquitquit`

src/bin/firebase-functions.ts
```diff
@@ -0,0 +1,50 @@
...
+  res.send('ok');
+  server.close(() => console.log('shutdown requested via /quitquitquit'));
+});
+
+let port = 8080;
```
inlined, line 44: You're conforming to what's in the CLI now. We never worked around the fact that Python _really_ doesn't want to allow two ports. So instead of having a control API on one port and the main API on another, we'd expose the control API when STACK_CONTROL_API is set. That's what's in the WTF-is-? deck at least. I don't love the name "STACK_CONTROL_API" so feel free to suggest a different one.

src/runtime/loader.ts
```diff
@@ -0,0 +1,98 @@
...
+  const mod = await loadModule(functionsDir);
+
+  extractEndpoints(mod, endpoints);
+
+  const backend: ManifestBackend = {
+    endpoints,
+    specVersion: 'v1alpha1',
+    requiredAPIs: {},
+  };
+  if (Object.values(endpoints).find((ep) => ep.scheduleTrigger)) {
```
inlined, line 93: Lol. This isn't the most cohesive place to put this. Requiring Pub/Sub is also only accurate for v1 (we'll use direct HTTP triggers for v2 so we can actually use the backoff behavior we advertise). What if we had a __requiredAPIs annotation along with __endpoint?

src/runtime/loader.ts
```diff
@@ -0,0 +1,98 @@
...
+    backend.requiredAPIs['pubsub'] = 'pubsub.googleapis.com';
```
inlined, line 94: I had map <friendly name, API> in the CLI because I thought we were supposed to print "pubsub" on deploy. I've since learned that we should print "functions" instead, so the key here is useless.

OTOH, I thought we _also_ wanted to add "reason" here. So the manifest would be

```yaml
requiredAPIs:
  - api: pubsub.googleapis.com
     reason: needed for v1 scheduled functions
  - api: cloudscheduler.googleapis.com
    reason: needed for scheduled functions
```

inlined (commented):
I like it, but let's get the port/Env var + the requiredAPIs stuff sorted out.

All in all though, this is much better than I could have done myself. We're really lucky that you've got experience with ESM modules vs pacakges.

spec/fixtures/sources/commonjs-grouped/g1.js
taeold: I'm still a go programmer at heart :o

src/runtime/loader.ts
```diff
@@ -76,34 +76,42 @@ export function extractStack(
...
     }
   }
 }
 
+function mergeRequiredAPIs(
+  requiredAPIs: ManifestRequiredAPI[]
+): ManifestRequiredAPI[] {
+  const apiToReasons: Record<string, Set<string>> = {};
+  for (const { api, reason } of requiredAPIs) {
+    const reasons = apiToReasons[api] || new Set();
```
inlined, line 92: I'm confused. Doesn't this never actually insert a new key into `apiToReasons`?

taeold: 🤦🏼 Definitely need tests here.

src/runtime/loader.ts
taeold: Done!

src/runtime/loader.ts
taeold: Fixed!

src/bin/firebase-functions.ts
taeold: Fixed.

spec/runtime/loader.spec.ts
taeold: Done!

spec/fixtures/sources/commonjs-grouped/index.js
taeold: Fixed!

src/runtime/manifest.ts
```diff
@@ -0,0 +1,85 @@
...
+  environmentVariables?: Record<string, string>;
+
+  httpsTrigger?: {
+    invoker?: string[];
+  };
+
+  callableTrigger?: {};
+
+  eventTrigger?: {
+    eventFilters: Record<string, string>;
```
colerogers, line 51: There was some discussion about how event filters can also take an operator field. Have we decided how we're going to handle that?

colerogers (commented):
looks good, just one thing that I had a question on

spec/runtime/loader.spec.ts
```diff
@@ -60,12 +61,12 @@ describe('extractStack', () => {
       },
     });
 
-    expect(requiredAPIs).to.be.deep.equal({
-      'cloudtasks.googleapis.com': {
+    expect(requiredAPIs).to.be.deep.equal([
+      {
         api: 'cloudtasks.googleapis.com',
         reason: 'Needed for v1 task queue functions',
```
inlined, line 67: nit: let's leave out "v1" so we can consolidate messages in mixed version codebases.

spec/v1/providers/analytics.spec.ts
```diff
@@ -326,7 +327,7 @@ describe('Analytics Functions', () => {
           () => null
         );
         expect(cloudFunction.__trigger).to.deep.equal({});
-        expect(cloudFunction.__endpoint).to.deep.equal({});
+        expect(cloudFunction.__endpoint).to.undefined;
```
inlined, line 330: nit: `to.be.undefined`

src/bin/firebase-functions.ts
```diff
@@ -15,20 +15,24 @@ Arguments:
...
+    printUsageAndExit();
+  }
+  functionsDir = args[0];
 }
-const functionsDir = args[0];
 
 let server;
 const app = express();
 
 app.get('/stack.yaml', async (req, res) => {
```
inlined, line 31: __/stack.yaml. We're reserving the namespace "__"

inlined, line 31: we can also add `app.post('__/quitquitquit')`

The prototypes used this instead of pkill because some of the launchers don't kill their children when killed (e.g. `go run` will leave the server running, but `POST http://localhost:$PORT/__/quitquitquit` will tell the service to shut down and the runner will gracefully follow.

src/bin/firebase-functions.ts
```diff
@@ -15,20 +15,24 @@ Arguments:
...
   try {
-    const backend = await loadStack(functionsDir);
+    const stack = await loadStack(functionsDir);
     res.setHeader('content-type', 'text/yaml');
-    res.send(JSON.stringify(backend));
+    res.send(JSON.stringify(stack));
```
inlined, line 35: I mean.... technically JSON is YAML?

src/cloud-functions.ts
```diff
@@ -466,23 +467,20 @@ export function makeCloudFunction<EventData>({
...
 
   if (options.schedule) {
     cloudFunction.__requiredAPIs = [
-      {
-        api: 'pubsub.googleapis.com',
-        reason: 'Needed for v1 scheduled functions.',
-      },
       {
         api: 'cloudscheduler.googleapis.com',
-        reason: 'Needed for v1 scheduled functions.',
```
inlined: here's a case where I could argue for using "v1" because v2 will not use Pub/Sub.

Alternatively you can just remove this required API because it's part of the APIs that get enabled with Cloud Functions.

inlined (approved):
Still looking good.

src/runtime/manifest.ts
taeold: Nope. I think we need to have a separate design/discussion, but since this manifest is still in alpha and not documented for our customers, we can improve it as we go.

src/cloud-functions.ts
taeold: I don't think Cloud Scheduler gets enabled as part of GCF, but I'm not sure what enables it since I do see it enabled when creating a new GCP project 🤔 . I'm going to add 'v1' to err on side of being safe, but can remove later when I dig in and figure out if we can rely on the cloud scheduler API being enabled on users project by some means.

PR #1004, opened 2021-11-12
Rename enqueue to dispatch to match API proposal

inlined:
Shoot! Not only did I rush and make the v2 API present tense instead of past tense, but I used enqueue vs dispatch in the naming conventions! Should now match go/task-functions

### DIFF
--- a/src/common/providers/https.ts
+++ b/src/common/providers/https.ts
@@ -171,16 +171,28 @@ export interface CallableRequest<T = any> {
 
 /** How a task should be retried in the event of a non-2xx return. */
 export interface TaskRetryConfig {
-  // If left unspecified, will default to 3
+  /**
+   * Maximum number of times a request should be attempted
+   * If left unspecified, will default to 3.
+   */
   maxAttempts?: number;
 
-  // If left unspecified will default to 1hr
+  /**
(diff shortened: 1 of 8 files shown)


src/common/providers/https.ts
```diff
@@ -171,16 +171,28 @@ export interface CallableRequest<T = any> {
 
 /** How a task should be retried in the event of a non-2xx return. */
 export interface TaskRetryConfig {
-  // If left unspecified, will default to 3
+  /**
+   * Maximum number of times a request should be attempted
```
taeold, line 175:
```suggestion
   * Maximum number of times a request should be attempted.
```

src/function-builder.ts
```diff
@@ -358,7 +358,7 @@ export class FunctionBuilder {
 
       /**
        * Declares a task queue function for clients to call using a Firebase Admin SDK.
-       * @param handler A method that takes a data and context and returns void or Promise<void>.
+       * @param options Configuratins for the task queue function.
```
taeold, line 361:
```suggestion
       * @param options Configurations for the task queue function.
```

taeold (approved):

PR #1009, opened 2021-11-29
More of annotate endpoint property for functions

taeold:
Follows up https://github.com/firebase/firebase-functions/pull/999 to annotate each funuctions with `__endpoint` property.

Highlight of changes:

* Extend unit test coverage for all v1 providers
* Add `__endpoint` annotation to v1 task queue functions
* Add `__requiredAPIs` annotation to task queue and scheduler functions
* Explicitly add an empty `__endpoint` annotation to functions defined in the handler namespace, similar in style to the `__trigger` annotation.
* Other minor nits

One more follow up to expect: I'll move the responsibility of adding labels for callable/scheduled function (e.g. `deployment-callable`, `deployment-scheduled`) from the SDK to the CLI soon.

### DIFF
--- a/spec/v1/providers/https.spec.ts
+++ b/spec/v1/providers/https.spec.ts
@@ -94,7 +94,7 @@ function runHandler(
 
 describe('CloudHttpsBuilder', () => {
   describe('#onRequest', () => {
-    it('should return a trigger/endpoint with appropriate values', () => {
+    it('should return a trigger with appropriate values', () => {
       const result = https.onRequest((req, resp) => {
         resp.send(200);
       });
@@ -130,7 +130,7 @@ describe('CloudHttpsBuilder', () => {
 
 describe('handler namespace', () => {
   describe('#onRequest', () => {
-    it('should return an empty trigger and endpoint', () => {
+    it('should return an empty trigger', () => {
       const result = functions.handler.https.onRequest((req, res) => {
         res.send(200);
       });
@@ -140,7 +140,7 @@ describe('handler namespace', () => {
   });
 
   describe('#onCall', () => {
-    it('should return an empty trigger and endpoint', () => {
+    it('should return an empty trigger', () => {
       const result = functions.handler.https.onCall(() => null);
       expect(result.__trigger).to.deep.equal({});
       expect(result.__endpoint).to.deep.equal({});
@@ -151,6 +151,7 @@ describe('handler namespace', () => {
     it('should return an empty trigger', () => {
       const result = functions.handler.https.taskQueue.onEnqueue(() => null);
       expect(result.__trigger).to.deep.equal({});
+      expect(result.__endpoint).to.deep.equal({});
     });
   });
 });
@@ -231,7 +232,7 @@ describe('#onCall', () => {
 });
 
 describe('#onEnqueue', () => {
-  it('should return a Trigger with appropriate values', () => {
+  it('should return a trigger/endpoint with appropriate values', () => {
     const result = https
       .taskQueue({
         rateLimits: {
@@ -248,7 +249,26 @@ describe('#onEnqueue', () => {
         invoker: 'private',
       })
       .onDispatch(() => {});
+
     expect(result.__trigger).to.deep.equal({
+      taskQueueTrigger: {
+        rateLimits: {
+          maxBurstSize: 20,
+          maxConcurrentDispatches: 30,
+          maxDispatchesPerSecond: 40,
+        },
+        retryConfig: {
+          maxAttempts: 5,
+          maxBackoffSeconds: 20,
+          maxDoublings: 3,
+          minBackoffSeconds: 5,
+        },
+        invoker: ['private'],
+      },
+    });
+
+    expect(result.__endpoint).to.deep.equal({
+      platform: 'gcfv1',
       taskQueueTrigger: {
         rateLimits: {
           maxBurstSize: 20,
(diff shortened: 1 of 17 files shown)


src/v2/providers/pubsub.ts
```diff
@@ -135,11 +135,6 @@ export function onMessagePublished<T = any>(
 
   func.run = handler;
 
-  // TypeScript doesn't recognize defineProperty as adding a property and complains
-  // that __trigger doesn't exist. We can either cast to any and lose all type safety
-  // or we can just assign a meaningless value before calling defineProperty.
-  func.__trigger = 'silence the transpiler';
```
colerogers, line 141: We don't need this anymore?

colerogers (approved):
Nice work on adding all these tests!

src/cloud-functions.ts
```diff
@@ -474,6 +474,19 @@ export function makeCloudFunction<EventData>({
     },
   });
 
+  if (options.schedule) {
+    cloudFunction.__requiredAPIs = [
```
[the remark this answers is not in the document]
taeold: We can have this annotation in the CLI, but the end result I'm reaching for is to have the SDK be able to fully describe a stack, including `__requiredAPIs`. Then, the "engine" (i.e. CLI) can simply execute on the manifest to make the stack happen.

In an offline discussion, we decided to reserve the use of required API to only include API that is required for deploying a function trigger and not ones required to run it. We are also excluding APIs that is required for all function triggers, e.g. the GCF API.

src/v2/providers/pubsub.ts
taeold: Nope - trigger annotation is now optional.

spec/v1/providers/https.spec.ts
```diff
@@ -151,6 +151,7 @@ describe('handler namespace', () => {
     it('should return an empty trigger', () => {
       const result = functions.handler.https.taskQueue.onEnqueue(() => null);
       expect(result.__trigger).to.deep.equal({});
+      expect(result.__endpoint).to.deep.equal({});
```
inlined, line 154: Maybe I'm missing something here. Doesn't _adding_ an `__endpoint` annotation now give you the additional work of checking whether an endpoint is empty before you report it in the manifest? It seems like code that is never run, but would cause some weird crashes if we ever tried.

src/cloud-functions.ts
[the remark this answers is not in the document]
inlined: We'll also eventually expose this as a user-facing API:

E.g.

```
requireApi('translate.googleapis.com', "To look up translations of user messages");
```

should be added to our sample code for a chat app that translates user messages. Then customers who copy + paste the code will always succeed rather htan getting runtime errors about an API not being enabled.

src/cloud-functions.ts
```diff
@@ -474,6 +474,19 @@ export function makeCloudFunction<EventData>({
...
+      {
+        api: 'pubsub.googleapis.com',
```
inlined, line 480: I wonder if we need to declare pubsub here. It's a core API that is always enabled with GCF and would presumedly be enabled in any future services.

src/cloud-functions.ts
```diff
@@ -474,6 +474,19 @@ export function makeCloudFunction<EventData>({
...
+        reason: 'Needed for v1 scheduled functions.',
+      },
+      {
+        api: 'cloudscheduler.googleapis.com',
+        reason: 'Needed for v1 scheduled functions.',
```
inlined, line 485: nit: can probably drop the "v1"

src/cloud-functions.ts
```diff
@@ -609,15 +622,14 @@ export function optionsToEndpoint(
...
-      vpc,
+      endpoint.vpc,
       options,
       'egressSettings',
       'vpcConnectorEgressSettings'
     );
-    endpoint.vpc = vpc;
   }
   convertIfPresent(endpoint, options, 'availableMemoryMb', 'memory', (mem) => {
     const memoryLookup = {
```
inlined, line 635: FML that Entity has already been defined to have memoryMB yet Run now uses these friendly names (almost) directly.

src/providers/https.ts
```diff
@@ -132,6 +138,28 @@ export class TaskQueueBuilder {
...
+      taskQueueTrigger: {},
+    };
+    copyIfPresent(func.__endpoint.taskQueueTrigger, this.tqOpts, 'retryConfig');
+    copyIfPresent(func.__endpoint.taskQueueTrigger, this.tqOpts, 'rateLimits');
+    convertIfPresent(
+      func.__endpoint.taskQueueTrigger,
+      this.tqOpts,
+      'invoker',
+      'invoker',
+      convertInvoker
```
inlined, line 153: Still think we should be using the invoker shorthand in the manifest

inlined (approved):
Good god that must have required patience to author. Good job!

spec/v1/providers/https.spec.ts
taeold: Wow I learned now that trying to deploy a function source with triggers defined using the handler namespace fails for the exact reason you described 😮 . With `__endpoint`, I'll make it so that this isn't the case.

src/providers/https.ts
taeold: convertInvoker keeps it as shorthand if passed in a shorthand invoker.

PR #1016, opened 2021-12-22
Check for missing options / databaseURL on emulator start

Hoppi164:
### Description

Bug Fix
When using firebase emulators Realtime Database throws an error due to app.options not existing.
I've found adding an extra check to the database constructor stops this error, and results in the emulator starting up correctly.

### Code sample

Instead of 
`if (app && app.options.databaseURL.startsWith('http:'))`
I've added
`if ( app && app.options && app.options.databaseURL && app.options.databaseURL.startsWith('http:') )`

Note: I've deliberately not used the modern ES6 optional chaining.
This syntax would be much cleaner, but is not supported by your lowest node version.
`app?.options?.databaseURL?.startsWith('http:')`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,3 @@
 - Parallelizes network calls that occur when validating authorization for onCall handlers.
 - Adds new regions to V2 API
-- Bug Fix - Check for missing options / databaseURL on emulator start
+- Fixes bug where the emulator crashed when given app without an `options` property.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Parallelizes network calls that occur when validating authorization for onCall handlers.
 - Adds new regions to V2 API
+- Bug Fix - Check for missing options / databaseURL on emulator start
--- a/src/providers/database.ts
+++ b/src/providers/database.ts
@@ -369,7 +369,7 @@ export class DataSnapshot {
     private app?: firebase.app.App,
     instance?: string
   ) {
-    if ( app?.options?.databaseURL?.startsWith('http:') ) {
+    if (app?.options?.databaseURL?.startsWith('http:')) {
       // In this case we're dealing with an emulator
       this.instance = app.options.databaseURL;
     } else if (instance) {
--- a/src/providers/database.ts
+++ b/src/providers/database.ts
@@ -369,7 +369,7 @@ export class DataSnapshot {
     private app?: firebase.app.App,
     instance?: string
   ) {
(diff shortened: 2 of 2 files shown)


src/providers/database.ts
```diff
@@ -369,7 +369,7 @@ export class DataSnapshot {
     private app?: firebase.app.App,
     instance?: string
   ) {
-    if (app && app.options.databaseURL.startsWith('http:')) {
+    if ( app && app.options && app.options.databaseURL && app.options.databaseURL.startsWith('http:') ) {
```
taeold, line 372: I actually think we can use optional chaining. Since we are using Typescript and targeting ES5, tsc will take care of translating the optional chaining into something that's supported in the runtime that does not support it.

taeold (changes requested):
Thank you for your great contribution!

Can you add an entry in the CHANGELOG to describe the bug fix?

src/providers/database.ts
Hoppi164: Hi Daniel

Thanks for pointing that out!
I've updated database.ts:372 to use optional chaining.

I've also added an entry to the CHANGELOG.md 
Please let me know if this is sufficient, or requires more detail in the changelog.

CHANGELOG.md
```diff
@@ -1,2 +1,3 @@
 - Parallelizes network calls that occur when validating authorization for onCall handlers.
 - Adds new regions to V2 API
+- Bug Fix - Check for missing options / databaseURL on emulator start
```
taeold, line 3: A small suggestion to make changelog entry in consistent:

```suggestion
- Fixes bug where the emulator crashed when given app without an `options` property.
```

taeold (changes requested):
One more change and we should 🚢

Hoppi164:
No problem.
Let me know if there's anything else you need me to change :)

taeold:
@Hoppi164 It looks like the linter is failing - can you run `npm run lint` or `npm run format` to fix the tests?

Hoppi164:
Hi @taeold I've run format:fix and removed the extra spacing i had on the `if` statement which the linter didn't like.
I've pushed this as another commit.

(note: there was a few other lint warnings in unrelated files, i've not corrected these as it seems out-of-scope for this PR)

taeold (approved):

taeold:
lgtm - thank you for your contribution (and your patience!)

Hoppi164:
No problem at all!
and thanks for all your help 😄 
Let me know if there's anything else you need me to do..

PR #1018, opened 2022-01-05
Add runtime option to associate secrets to functions

taeold:
Google Cloud Functions (GCF) now have support for loading secrets stored in Secret Manager to securely access application secrets (e.g. Third party API keys) on function executions. See the [documentations(https://cloud.google.com/functions/docs/configuring/secrets) for more details.

We propose extending the function runtime option to associate secrets. Then, the secrets can be access as an environment once deployed to GCF:

```js

const helloSecret = functions
  .runWith({ 
    // Loads secret
    //   projects/{PROJECT}/secrets/MY_SECRET/versions/{current latest secret version}
    secrets: ['MY_SECRET']
  })
  .https
  .onRequest((res, req) => {
    req.send(`Shhhhh. Our secret: ${process.env.MY_SECRET}`);
  });

const helloSecretVersion = functions
  .runWith({ 
    // Loads secret
    //   projects/{PROJECT}/secrets/MY_SECRET/versions/3
    secrets: ['MY_SECRET@3']
  })
  .https
  .onRequest((res, req) => {
    req.send(`Shhhhh. Our secret: ${process.env.MY_SECRET}`);
  });


// Optionally provide the full resource name.
const helloSecretFullName = functions
  .runWith({ 
    secrets: [
      'projects/my-project/secrets/MY_SECRET', // as secret resource name
      'projects/my-project/secrets/ANOTHER_SECRET/versions/3', // or secret version resource name
    ] 
  })
  .onRequest((res, req) => {
    // Environment variable name is still the secret id.
    req.send(`Shhhhh. Our secret: ${process.env.MY_SECRET}`);
  });
```

### DIFF
--- a/spec/v1/function-builder.spec.ts
+++ b/spec/v1/function-builder.spec.ts
@@ -472,10 +472,58 @@ describe('FunctionBuilder', () => {
     ).to.throw();
   });
 
-  it('', () => {
+  it('should throw an error if private identifier is in the invoker array', () => {
     expect(() =>
       functions.runWith({
         invoker: ['service-account1', 'private', 'service-account2'],
+      })
+    ).to.throw();
+  });
+
+  it('should allow valid secret config expressed using short form', () => {
+    const secrets = ['API_KEY', 'API_KEY@3', 'API_KEY@latest'];
+    const fn = functions
+      .runWith({ secrets })
+      .auth.user()
+      .onCreate((user) => user);
+
+    expect(fn.__trigger.secrets).to.deep.equal(secrets);
+  });
+
+  it('should allow valid secret config expressed with full resource name', () => {
+    const secrets = [
+      'projects/my-project/secrets/API_KEY',
+      'projects/my-project/secrets/API_KEY/versions/3',
+      'projects/my-project/secrets/API_KEY/versions/latest',
+    ];
+    const fn = functions
+      .runWith({ secrets })
+      .auth.user()
+      .onCreate((user) => user);
+
+    expect(fn.__trigger.secrets).to.deep.equal(secrets);
+  });
+
+  it('should throw error given invalid secret config', () => {
+    expect(() =>
+      functions.runWith({
+        secrets: ['ABC!@#$'],
+      })
+    ).to.throw();
+  });
+
+  it('should throw error given invalid secret version on short form', () => {
+    expect(() =>
+      functions.runWith({
+        secrets: ['ABC@live'],
+      })
+    ).to.throw();
+  });
+
+  it('should throw error given invalid secret version on long form', () => {
+    expect(() =>
+      functions.runWith({
+        secrets: ['projects/my-project/secrets/API_KEY/versions/live'],
       })
     ).to.throw();
   });
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -499,7 +499,8 @@ export function optionsToTrigger(options: DeploymentOptions) {
     'ingressSettings',
     'vpcConnectorEgressSettings',
     'vpcConnector',
-    'labels'
+    'labels',
+    'secrets'
   );
   convertIfPresent(
     trigger,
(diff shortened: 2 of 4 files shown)


spec/v1/function-builder.spec.ts
```diff
@@ -472,11 +472,59 @@ describe('FunctionBuilder', () => {
...
+  it('should throw an error if private identifier is in the invoker array', () => {
     expect(() =>
       functions.runWith({
         invoker: ['service-account1', 'private', 'service-account2'],
       })
     ).to.throw();
   });
+
+  it('should allow valid secret config expressed using short form', () => {
+    const secrets = ['API_KEY', 'API_KEY@3', 'API_KEY@latest'];
```
inlined, line 484: Can we cut @version from scope? I'd like to not create more of a legacy burden than we need.

spec/v1/function-builder.spec.ts
```diff
@@ -472,11 +472,59 @@ describe('FunctionBuilder', () => {
...
+      .runWith({ secrets })
+      .auth.user()
+      .onCreate((user) => user);
+
+    expect(fn.__trigger.secrets).to.deep.equal(secrets);
+  });
+
+  it('should allow valid secret config expressed with full resource name', () => {
+    const secrets = [
+      'projects/my-project/secrets/API_KEY',
```
inlined, line 495: Can we cut full resource names from scope? I'd like to not create more of a legacy burden than we need? I'm also worried about increasing scope to include cross-project secrets.

inlined (changes requested):

spec/v1/function-builder.spec.ts
taeold: Discussed offline. Not supporting full resource name (i.e. cross project secrets) in the first iteration.

spec/v1/function-builder.spec.ts
taeold: Discussed offline. Not supporting versions in the first iteration.

src/function-builder.ts
```diff
@@ -229,6 +229,19 @@ function assertRuntimeOptionsValid(runtimeOptions: RuntimeOptions): boolean {
...
 
+  if (runtimeOptions.secrets !== undefined) {
+    const invalidSecrets = runtimeOptions.secrets.filter(
+      (s) => !/^[A-Za-z\d\-_]+$/.test(s)
+    );
+    if (invalidSecrets.length > 0) {
+      throw new Error(
+        `Invalid secrets: ${invalidSecrets.join(',')}. ` +
+          'Secret must be configured using the short form (e.g. API_KEY@1) ' +
+          'or with full resource name (e.g. projects/my-project/secrets/API_KEY).'
```
colerogers, line 240: Should we update this error message? The tests show that we can only configure secrets in the short form name without versions, but this error message states otherwise.

colerogers (changes requested):

colerogers (approved):

src/cloud-functions.ts
```diff
@@ -620,6 +622,13 @@ export function optionsToEndpoint(
     'serviceAccount',
     (sa) => sa
   );
+  convertIfPresent(
+    endpoint,
+    options,
+    'secretEnvironmentVariables',
+    'secrets',
+    (secrets) => secrets.map((secret) => ({ secret }))
```
inlined, line 630: This is where the fully qualified names/versions gets weird because we agreed elsewhere that we'd be setting `key` and not `secret` so that we could bind later. I guess we can, in this case, bind both and not bind `secret` in the params case?

taeold: Ah that's good point.

I've been thinking the opposite - that v1 Functions SDK will receive secret and the CLI will bind the env key using the secret name. With params context, having the option to later bind the secret makes sense. 

Being restrictive here - to set both `key` and `secret` to the secret input is good and letting params be the one that only binds the `secret` makes more sense to me.

inlined (approved):

PR #1020, opened 2022-01-19
Generic interface for provider enhancement

colerogers:
Change for the generic interface. Exposes `onAlertPublished` from the `v2/alerts` namespace

Ex ~
```
import { onAlertPublished } from "firebase-functions/v2/alerts";

export const func1 = onAlertPublished("crashlytics.newFatalIssue",(event) => { … });

export const func2 = onAlertPublished(
  { alertType: "crashlytics.newFatalIssue" },
  (event) => { … }
);

export const func3 = onAlertPublished(
  { 
    alertType: "crashlytics.newFatalIssue",
    appId: "123456789"
  },
  (event) => { … }
);
```

### DIFF
--- a/spec/v2/providers/alerts/alerts.spec.ts
+++ b/spec/v2/providers/alerts/alerts.spec.ts
@@ -1,18 +1,10 @@
 import { expect } from 'chai';
-import { CloudEvent, CloudFunction } from '../../../../src/v2/core';
 import * as options from '../../../../src/v2/options';
 import * as alerts from '../../../../src/v2/providers/alerts';
 import { FULL_ENDPOINT, FULL_OPTIONS } from '../helpers';
 
 const ALERT_TYPE = 'new-alert-type';
 const APPID = '123456789';
-
-function getMockFunction(): CloudFunction<alerts.FirebaseAlertData<String>> {
-  const func = (raw: CloudEvent<unknown>) => 42;
-  func.run = (event: CloudEvent<alerts.FirebaseAlertData<String>>) => 42;
-  func.__endpoint = {};
-  return func;
-}
 
 describe('alerts', () => {
   describe('onAlertPublished', () => {
@@ -75,11 +67,7 @@ describe('alerts', () => {
     });
 
     it('should define the endpoint without appId and opts', () => {
-      const func = getMockFunction();
-
-      func.__endpoint = alerts.getEndpointAnnotation({}, ALERT_TYPE);
-
-      expect(func.__endpoint).to.deep.equal({
+      expect(alerts.getEndpointAnnotation({}, ALERT_TYPE)).to.deep.equal({
         platform: 'gcfv2',
         labels: {},
         eventTrigger: {
@@ -93,14 +81,9 @@ describe('alerts', () => {
     });
 
     it('should define a complex endpoint without appId', () => {
-      const func = getMockFunction();
-
-      func.__endpoint = alerts.getEndpointAnnotation(
-        { ...FULL_OPTIONS },
-        ALERT_TYPE
-      );
-
-      expect(func.__endpoint).to.deep.equal({
+      expect(
+        alerts.getEndpointAnnotation({ ...FULL_OPTIONS }, ALERT_TYPE)
+      ).to.deep.equal({
         ...FULL_ENDPOINT,
         eventTrigger: {
           eventType: alerts.eventType,
@@ -113,15 +96,9 @@ describe('alerts', () => {
     });
 
     it('should define a complex endpoint', () => {
-      const func = getMockFunction();
-
-      func.__endpoint = alerts.getEndpointAnnotation(
-        { ...FULL_OPTIONS },
-        ALERT_TYPE,
-        APPID
-      );
-
-      expect(func.__endpoint).to.deep.equal({
+      expect(
+        alerts.getEndpointAnnotation({ ...FULL_OPTIONS }, ALERT_TYPE, APPID)
+      ).to.deep.equal({
         ...FULL_ENDPOINT,
         eventTrigger: {
           eventType: alerts.eventType,
@@ -144,15 +121,10 @@ describe('alerts', () => {
         region: 'us-west1',
         minInstances: 3,
       };
-      const func = getMockFunction();
 
-      func.__endpoint = alerts.getEndpointAnnotation(
-        specificOpts,
-        ALERT_TYPE,
-        APPID
-      );
-
-      expect(func.__endpoint).to.deep.equal({
+      expect(
+        alerts.getEndpointAnnotation(specificOpts, ALERT_TYPE, APPID)
+      ).to.deep.equal({
         platform: 'gcfv2',
         labels: {},
         concurrency: 20,
--- a/spec/v2/providers/alerts/alerts.spec.ts
+++ b/spec/v2/providers/alerts/alerts.spec.ts
@@ -1,15 +1,8 @@
 import { expect } from 'chai';
-import * as sinon from 'sinon';
-import * as config from '../../../../src/config';
 import { CloudEvent, CloudFunction } from '../../../../src/v2/core';
 import * as options from '../../../../src/v2/options';
 import * as alerts from '../../../../src/v2/providers/alerts';
-import {
-  BASIC_ENDPOINT,
-  BASIC_OPTIONS,
-  FULL_ENDPOINT,
-  FULL_OPTIONS,
-} from '../helpers';
+import { FULL_ENDPOINT, FULL_OPTIONS } from '../helpers';
 
 const ALERT_TYPE = 'new-alert-type';
 const APPID = '123456789';
@@ -17,7 +10,6 @@ const APPID = '123456789';
 function getMockFunction(): CloudFunction<alerts.FirebaseAlertData<String>> {
   const func = (raw: CloudEvent<unknown>) => 42;
   func.run = (event: CloudEvent<alerts.FirebaseAlertData<String>>) => 42;
-  func.__trigger = 'silence the transpiler';
   func.__endpoint = {};
   return func;
 }
@@ -43,7 +35,7 @@ describe('alerts', () => {
     it('should create the function with opts', () => {
       const result = alerts.onAlertPublished(
         {
-          ...BASIC_OPTIONS,
+          ...FULL_OPTIONS,
           alertType: ALERT_TYPE,
           appId: APPID,
         },
@@ -51,7 +43,7 @@ describe('alerts', () => {
       );
 
       expect(result.__endpoint).to.deep.equal({
-        ...BASIC_ENDPOINT,
+        ...FULL_ENDPOINT,
         eventTrigger: {
           eventType: alerts.eventType,
           eventFilters: {
@@ -72,24 +64,20 @@ describe('alerts', () => {
     });
   });
 
-  describe('defineEndpoint', () => {
-    let configStub: sinon.SinonStub;
-
+  describe('getEndpointAnnotation', () => {
     beforeEach(() => {
       process.env.GCLOUD_PROJECT = 'aProject';
-      configStub = sinon.stub(config, 'firebaseConfig');
     });
 
     afterEach(() => {
       options.setGlobalOptions({});
       delete process.env.GCLOUD_PROJECT;
-      configStub.restore();
     });
 
     it('should define the endpoint without appId and opts', () => {
       const func = getMockFunction();
 
-      alerts.defineEndpoint(func, {}, ALERT_TYPE);
+      func.__endpoint = alerts.getEndpointAnnotation({}, ALERT_TYPE);
 
       expect(func.__endpoint).to.deep.equal({
         platform: 'gcfv2',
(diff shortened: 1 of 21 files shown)


spec/v2/providers/helpers.ts
```diff
@@ -1,5 +1,20 @@
 import * as options from '../../../src/v2/options';
 
+export const BASIC_OPTIONS: options.EventHandlerOptions = {
```
taeold, line 3: q: Was `FULL_OPTIONS` not appropriate here? IIUC, using `FULL_OPTIONS` wouldn't necessarily add more lines of test code and cover more cases no?

src/v2/core.ts
```diff
@@ -49,10 +49,10 @@ export interface TriggerAnnotation {
 }
 
 /**
- * A CloudEvent is a cross-platform format for encoding a serverless event.
+ * A CloudEventBase is the base of a cross-platform format for encoding a serverless event.
  * More information can be found in https://github.com/cloudevents/spec
  */
-export interface CloudEvent<T> {
+export interface CloudEventBase<T> {
```
taeold, line 55: hm maye this no longer needs to be exported?

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+  | 'appDistribution.newTesterIosDevice'
+  | string;
+
+/** Options */
+export interface FirebaseAlertOptions extends options.EventHandlerOptions {
+  alertType: AlertType;
+  appId?: string;
+}
+
+/** Cloud Event Type */
```
taeold, line 34:
```suggestion
/** CloudEvent Type */
```

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+  | 'crashlytics.regression'
+  | 'crashlytics.stabilityDigest'
+  | 'crashlytics.velocity'
+  | 'crashlytics.newAnrIssue'
+  | 'billing.planUpdate'
+  | 'billing.automatedPlanUpdate'
+  | 'appDistribution.newTesterIosDevice'
+  | string;
+
+/** Options */
```
taeold, line 28: nit* I found some of these comments kinda unhelpful - e.g. I can tell from the interface name that this is an option.  Can we expand on the documentations or get rid of them?

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+export interface FirebaseAlertOptions extends options.EventHandlerOptions {
+  alertType: AlertType;
+  appId?: string;
+}
+
+/** Cloud Event Type */
+interface WithAlertTypeAndApp {
```
taeold, line 35: I think type names should be a noun and not adjectives, so maybe something like `AlertEventExt` is more appropriate.

(I think the `&` type intersection encompasses the `with` you are trying to capture here).

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+      raw as CloudEvent<FirebaseAlertData<T>, WithAlertTypeAndApp>
+    );
+  };
+
+  func.run = handler;
+
+  // TypeScript doesn't recognize defineProperty as adding a property and complains
+  // that __endpoint doesn't exist. We can either cast to any and lose all type safety
+  // or we can just assign a meaningless value before calling defineProperty.
+  func.__trigger = 'silence the transpiler';
```
taeold, line 63: Think `__trigger` prop is optional now and we don't require this hack.

spec/v2/providers/alerts/alerts.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+    it('should have a .run method', () => {
+      const func = alerts.onAlertPublished(ALERT_TYPE, (event) => event);
+
+      const res = func.run('input' as any);
+
+      expect(res).to.equal('input');
+    });
+  });
+
+  describe('defineEndpoint', () => {
```
taeold, line 75: I kinda liked our way of testing high-level, external API (e.g. `onAlertPublished`) vs internal helper function like `defineEndpoint`. So we could just add a test case to check `__endpoint` prop is approriately set on each external API instead. WDYT?

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+ * @internal
+ * Helper function for defining an endpoint for alert handling functions
+ */
+export function defineEndpoint(
+  func: CloudFunction<FirebaseAlertData<any>>,
+  opts: options.EventHandlerOptions,
+  alertType: string,
+  appId?: string
+): void {
+  Object.defineProperty(func, '__endpoint', {
```
taeold, line 80: I don't think this has to be a dynamic property. We did this in storage trigger only because we needed to get `bucket` from `FIREBASE_CONFIG`, but that's not the case here. See `https.ts` for an example of static __endpoint definition!

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+    );
+  };
+
+  func.run = handler;
+
+  // TypeScript doesn't recognize defineProperty as adding a property and complains
+  // that __endpoint doesn't exist. We can either cast to any and lose all type safety
+  // or we can just assign a meaningless value before calling defineProperty.
+  func.__trigger = 'silence the transpiler';
+  func.__endpoint = {} as ManifestEndpoint;
```
taeold, line 64: Would prefer

```
func.__endpoint = defineEndpoint(...)
```

since I think we should shy away from type assertion if it's unnecessary.

Just inlining the function implementation here could also work fine?

taeold (changes requested):
Couple of suggestions and changes.

spec/v2/providers/alerts/alerts.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+    let configStub: sinon.SinonStub;
+
+    beforeEach(() => {
+      process.env.GCLOUD_PROJECT = 'aProject';
+      configStub = sinon.stub(config, 'firebaseConfig');
```
taeold, line 80: hmm do you know why these are necessary? I thought only storage trigger made use of firebaseconfig when defining the trigger/endpoint.

src/v2/providers/alerts/alerts.ts
taeold: I notice that this name was approved in the API proposal. I should've objected then - feel free to ignore.

spec/v2/providers/alerts/alerts.spec.ts
taeold: I think i now understand the purpose of defineEndpoint - it's meant to be used across all alerts API. This feels fine then. A minor suggestion for using `addEndpointAnnotation` to make it a little more obvious that calliing this fn has side-effect.

spec/v2/providers/alerts/alerts.spec.ts
colerogers: It's not necessary to have this in, removed in the latest commit

src/v2/providers/alerts/alerts.ts
colerogers: Re-named and changed to

```
func.__endpoint = getEndpointAnnotation(...);
```

src/v2/providers/alerts/alerts.ts
colerogers: Makes sense, changed to static definition

spec/v2/providers/alerts/alerts.spec.ts
colerogers: I ended up changing the function to return the endpoint object (from your comment below) and re-named it to `getEndpointAnnotation`. How do you feel about that?

src/v2/providers/alerts/alerts.ts
colerogers: thanks, removed

src/v2/providers/alerts/alerts.ts
colerogers: I had those comments to break up the file into sections, but since we don't really do that anywhere else, I've made them a bit more descriptive

src/v2/providers/alerts/alerts.ts
colerogers: fixed

src/v2/core.ts
colerogers: removed the export

spec/v2/providers/helpers.ts
colerogers: You're right, I originally had them in here when I stubbed out `__endpoint` and `FULL_ENDPOINT` didn't exist. Removed them and cleaned up the tests

spec/v2/providers/alerts/alerts.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+  FULL_OPTIONS,
+} from '../helpers';
+
+const ALERT_TYPE = 'new-alert-type';
+const APPID = '123456789';
+
+function getMockFunction(): CloudFunction<alerts.FirebaseAlertData<String>> {
+  const func = (raw: CloudEvent<unknown>) => 42;
+  func.run = (event: CloudEvent<alerts.FirebaseAlertData<String>>) => 42;
+  func.__trigger = 'silence the transpiler';
```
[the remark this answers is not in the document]
colerogers: thanks, removed

spec/v2/providers/alerts/alerts.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      process.env.GCLOUD_PROJECT = 'aProject';
+    });
+
+    afterEach(() => {
+      options.setGlobalOptions({});
+      delete process.env.GCLOUD_PROJECT;
+    });
+
+    it('should define the endpoint without appId and opts', () => {
+      const func = getMockFunction();
```
taeold, line 78: Here and onward: Now that getEndpointAnnotation is a purefunction, it looks like we don't need any of the `getMockFunction` artifacts, e.g.:

```js
expect(alerts.getEndpointAnnotation(...)).to.deep.equal.to(...)
```

should suffice?

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+  endTime: string;
+  payload: T;
+}
+
+interface WithAlertTypeAndApp {
+  alertType: string;
+  appId?: string;
+}
+/**
+ * A custom CloudEvent for Firebase Alerts with custom extension attributes defined
```
taeold, line 19:
```suggestion
 * A custom CloudEvent for Firebase Alerts (with custom extension attributes).
```

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
+import { ManifestEndpoint } from '../../../common/manifest';
+import { CloudEvent, CloudFunction } from '../../core';
+import * as options from '../../options';
+
+/**
+ * The data object that is emitted from Firebase Alerts inside the CloudEvent
```
taeold, line 6:
```suggestion
 * The CloudEvent data emitted by Firebase Alerts.
```

taeold (approved):
lgtm, one minor comment on test

src/v2/providers/alerts/alerts.ts
colerogers: Fixed

src/v2/providers/alerts/alerts.ts
colerogers: Fixed

spec/v2/providers/alerts/alerts.spec.ts
colerogers: Fixed

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+    labels: {
+      ...baseOpts?.labels,
+      ...specificOpts?.labels,
+    },
+    eventTrigger: {
+      eventType,
+      eventFilters: {
+        alertType,
+      },
+      retry: false,
```
inlined, line 97: Do we always hard code retry: false? We should be supporting booleans here now that GCF gen 2 supports retry (though I actually have a thread with them about changing this from a boolean to a more featured struct)

src/v2/providers/alerts/alerts.ts
```diff
@@ -0,0 +1,127 @@
...
+  return endpoint;
+}
+
+/**
+ * @internal
+ * Helper function to parse the function opts, alert type, and appId.
+ */
+export function getOptsAndAlertTypeAndApp(
+  alertTypeOrOpts: AlertType | FirebaseAlertOptions
+): [options.EventHandlerOptions, string, string | undefined] {
```
inlined, line 112: Returning a tuple is pythonic. Returning an object would have probably been more idiomatic JS. Regardless, this is probably easier to read so LGTM.

inlined (approved):

src/v2/providers/alerts/alerts.ts
colerogers: Replaced this hard coded value with `opts.retry` in the latest commit

PR #1021, opened 2022-01-19
Specific interface for provider enhancement (1/3)

colerogers:
Change for the the specific app distro interface. Exposes `onNewTesterIosDevicePublished` from the `v2/alerts/appDistribution` namespace

Ex ~
```
import { onNewTesterIosDevicePublished } from "firebase-functions/v2/alerts/appDistribution";

// newTesterIosDevicePublished interface
export const func = onNewTesterIosDevicePublished((event) => { … });
export const func1 = onNewTesterIosDevicePublished("123456789", (event) => { … });
export const func2 = onNewTesterIosDevicePublished({ appId: "123456789" }, (event) => { … });
```

### DIFF
--- a/src/v2/providers/alerts/appDistribution.ts
+++ b/src/v2/providers/alerts/appDistribution.ts
@@ -1,10 +1,10 @@
-import { defineEndpoint, FirebaseAlertData } from '.';
-import { ManifestEndpoint } from '../../../common/manifest';
+import { getEndpointAnnotation, FirebaseAlertData } from './alerts';
 import { CloudEvent, CloudFunction } from '../../core';
 import * as options from '../../options';
 
-/** Data */
-/** Payload for a new tester device */
+/**
+ * The internal payload object that is wrapped inside a FirebaseAlertData object
+ */
 export interface NewTesterDevicePayload {
   ['@type']: 'com.google.firebase.firebasealerts.NewTesterDevicePayload';
   testerName: string;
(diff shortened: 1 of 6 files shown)


src/v2/providers/alerts/appDistribution.ts
```diff
@@ -0,0 +1,104 @@
...
+    return handler(raw as AppDistributionEvent<NewTesterDevicePayload>);
+  };
+
+  func.run = handler;
+
+  // TypeScript doesn't recognize defineProperty as adding a property and complains
+  // that __endpoint doesn't exist. We can either cast to any and lose all type safety
+  // or we can just assign a meaningless value before calling defineProperty.
+  func.__trigger = 'silence the transpiler';
+  func.__endpoint = {} as ManifestEndpoint;
```
taeold, line 83: nit* Both of these are unnecessary AFAIK.

src/v2/providers/alerts/appDistribution.ts
```diff
@@ -0,0 +1,104 @@
+import { defineEndpoint, FirebaseAlertData } from '.';
```
taeold, line 1: nit* I would've prefer having this point to a file instead, e.g. './alerts`. Saves me one jump to get to the actual source.

src/v2/providers/alerts/index.ts
```diff
@@ -1 +1,5 @@
+import * as appDistribution from './appDistribution';
```
taeold, line 1: nit* think this might be same as

```
export * as appDistribution from './appDistribution';
```

Also, why is this needed if we are explicitly declaring export path in package.json?

taeold (commented):
mostly lgtm. waiting on some changes to flow from https://github.com/firebase/firebase-functions/pull/1020.

src/v2/providers/alerts/appDistribution.ts
colerogers: Removed

src/v2/providers/alerts/appDistribution.ts
colerogers: fixed

src/v2/providers/alerts/index.ts
colerogers: removed since we declare export path in the package file

src/v2/providers/alerts/appDistribution.ts
```diff
@@ -1,10 +1,10 @@
-import { defineEndpoint, FirebaseAlertData } from '.';
-import { ManifestEndpoint } from '../../../common/manifest';
+import { getEndpointAnnotation, FirebaseAlertData } from './alerts';
 import { CloudEvent, CloudFunction } from '../../core';
 import * as options from '../../options';
 
-/** Data */
-/** Payload for a new tester device */
+/**
+ * The internal payload object that is wrapped inside a FirebaseAlertData object
```
taeold, line 6: For consistency, I think we should end each comment with a period `.`

taeold (approved):
lgtm w/ a comment on adding punctuation on public doc strings.

src/v2/providers/alerts/appDistribution.ts
colerogers: Added periods to public comments

inlined (approved):
I love it

PR #1022, opened 2022-01-19
Specific interface for provider enhancement (2/3)

colerogers:
Change for the the specific billing interface. Exposes `onPlanUpdatePublished` & `onAutomatedPlanUpdatePublished` from the `v2/alerts/billing` namespace

Ex ~ 
```
import {
  onPlanUpdatePublished,
  onAutomatedPlanUpdatePublished
} from "firebase-functions/v2/alerts/billing";

// planUpdatePublished interface
export const func1 = onPlanUpdatePublished((event) => { … });
export const func2 = onPlanUpdatePublished({ region: "us-west1" }, (event) => { … });
// robotPlanUpdatePublished interface
export const func3 = onAutomatedPlanUpdatePublished((event) => { … });
export const func5 = onAutomatedPlanUpdatePublished({ region: "us-west1" }, (event) => { … });
```

### DIFF
--- a/src/v2/providers/alerts/billing.ts
+++ b/src/v2/providers/alerts/billing.ts
@@ -4,19 +4,16 @@ import { CloudEvent, CloudFunction } from '../../core';
 import * as options from '../../options';
 
 /** Data */
-// billing.planUpdate
 export interface PlanUpdatePayload {
   ['@type']: 'com.google.firebase.firebasealerts.PlanUpdatePayload';
   billingPlan: string;
   principalEmail: string;
 }
-// billing.automatedPlanUpdate
 export interface PlanAutomatedUpdatePayload {
   ['@type']: 'com.google.firebase.firebasealerts.PlanAutomatedUpdatePayload';
   billingPlan: string;
 }
 
-/** Events */
 /** @internal */
 export const planUpdateAlert = 'billing.planUpdate';
 /** @internal */
--- a/src/v2/providers/alerts/billing.ts
+++ b/src/v2/providers/alerts/billing.ts
@@ -37,8 +37,9 @@ export const automatedPlanUpdateAlert = 'billing.automatedPlanUpdate';
 /** @internal */
 type BillingEventHandler<T> = (event: BillingEvent<T>) => any | Promise<any>;
 
-/** Handlers */
-/** Handle a plan update published */
+/**
+ * Declares a function that can handle a billing plan update event.
+ */
 export function onPlanUpdatePublished(
   handler: BillingEventHandler<PlanUpdatePayload>
 ): CloudFunction<FirebaseAlertData<PlanUpdatePayload>>;
@@ -59,7 +60,9 @@ export function onPlanUpdatePublished(
   );
 }
 
-/** Handle an automated plan update published */
+/**
+ * Declares a function that can handle an automated billing plan update event.
+ */
 export function onAutomatedPlanUpdatePublished(
   handler: BillingEventHandler<PlanAutomatedUpdatePayload>
 ): CloudFunction<FirebaseAlertData<PlanAutomatedUpdatePayload>>;
--- a/src/v2/providers/alerts/billing.ts
+++ b/src/v2/providers/alerts/billing.ts
@@ -1,0 +1,105 @@
+import { defineEndpoint, FirebaseAlertData } from '.';
+import { ManifestEndpoint } from '../../../common/manifest';
+import { CloudEvent, CloudFunction } from '../../core';
+import * as options from '../../options';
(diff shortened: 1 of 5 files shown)


src/v2/providers/alerts/billing.ts
```diff
@@ -0,0 +1,106 @@
...
+/** @internal */
+export function onOperation<T>(
+  alertType: string,
+  optsOrHandler:
+    | options.EventHandlerOptions
+    | ((event: BillingEvent<T>) => any | Promise<any>),
+  handler: (event: BillingEvent<T>) => any | Promise<any>
+): CloudFunction<FirebaseAlertData<T>> {
+  if (typeof optsOrHandler === 'function') {
+    handler = optsOrHandler as (event: BillingEvent<T>) => any | Promise<any>;
```
taeold, line 88: Woo `(event: BillingEvent<T>) => any | Promise<any>;` is used thrice! This meets the threshold for DRY for me:

```ts
type BillingEventHandler<T> = (event: BillingEvent<T>) => any | Promise<any>;
```

src/v2/providers/alerts/billing.ts
```diff
@@ -0,0 +1,106 @@
...
+    return handler(raw as BillingEvent<T>);
+  };
+
+  func.run = handler;
+
+  // TypeScript doesn't recognize defineProperty as adding a property and complains
+  // that __endpoint doesn't exist. We can either cast to any and lose all type safety
+  // or we can just assign a meaningless value before calling defineProperty.
+  func.__trigger = 'silence the transpiler';
+  func.__endpoint = {} as ManifestEndpoint;
```
taeold, line 102: Can be removed.

src/v2/providers/alerts/billing.ts
```diff
@@ -0,0 +1,106 @@
...
+  defineEndpoint(func, optsOrHandler, alertType, undefined);
```
taeold, line 103:
```suggestion
  defineEndpoint(func, optsOrHandler, alertType);
```

src/v2/providers/alerts/billing.ts
colerogers: Oh that looks way better, thanks for the suggestion!

src/v2/providers/alerts/billing.ts
colerogers: removed

src/v2/providers/alerts/billing.ts
colerogers: removed

taeold (approved):
🔥 Great stuff.

src/v2/providers/alerts/billing.ts
```diff
@@ -0,0 +1,105 @@
...
+ */
+export type BillingEvent<T> = CloudEvent<FirebaseAlertData<T>, WithAlertType>;
+
+/** @internal */
+export const planUpdateAlert = 'billing.planUpdate';
+/** @internal */
+export const automatedPlanUpdateAlert = 'billing.automatedPlanUpdate';
+
+/** @internal */
+type BillingEventHandler<T> = (event: BillingEvent<T>) => any | Promise<any>;
```
inlined, line 38: This typedef makes code complete worse (it asks for a BillingEventHandler instead of a function signature)

inlined (approved):

src/v2/providers/alerts/billing.ts
colerogers: Oh shoot, we should probably change it back for better usability then, right?

taeold: I wanted to test out the automcomplete and see what shows, but I couldn't get to any of the billing handlers at all.

@colerogers Can you double check if this is working for you?

1) `npm run build:pack`
2) Install the generated tgz locally on your function
3) Try to get import to work.

taeold: In fact, I couldn't even get `lib/v2/providers/billing.js` to show up. what gives?

colerogers: @taeold I was able to get it to work, just pushed an update. Seems like we can't rely on the package.json export path alone, we need to also export from the root index.ts. I think there was a comment about exporting on one of the other PRs and I ended up changing it on all of them

colerogers: @inlined Removed BillingEventHandler type from the latest commit

src/v2/providers/alerts/billing.ts
colerogers: @taeold Removed based on typing feedback below

PR #1026, opened 2022-02-01
Move common auth items into common path

colerogers:
Simple refactor that moves common auth code pieces into the common directory under the identity namespace.

### DIFF
--- a/spec/common/providers/identity.spec.ts
+++ b/spec/common/providers/identity.spec.ts
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2017 Firebase
+// Copyright (c) 2022 Firebase
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/spec/common/providers/identity.spec.ts
+++ b/spec/common/providers/identity.spec.ts
@@ -1,0 +1,88 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017 Firebase
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


spec/common/providers/identity.spec.ts
```diff
@@ -0,0 +1,88 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017 Firebase
```
taeold, line 3:
```suggestion
// Copyright (c) 2022 Firebase
```

src/common/providers/identity.ts
```diff
@@ -0,0 +1,111 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017 Firebase
```
taeold, line 3:
```suggestion
// Copyright (c) 2022 Firebase
```

src/providers/auth.ts
```diff
@@ -52,21 +54,9 @@ export function _userWithOptions(options: DeploymentOptions) {
   }, options);
 }
 
-export class UserRecordMetadata implements firebase.auth.UserMetadata {
```
taeold, line 55: I think it's possible that we break some user's code if we remove exported class like this one (similarly for any other exported objects)

Let's create an alias to prevent breaking change (and maybe add a note in go/cf3-v4-sdk to clean it up in the next major version.

taeold (changes requested):

spec/common/providers/identity.spec.ts
colerogers: Fixed

src/common/providers/identity.ts
colerogers: Fixed

src/providers/auth.ts
colerogers: Oh yeah that's a great point. I re-exported them in the auth code path, added a todo comment about yanking them, and updated that doc.

taeold (approved):

src/common/providers/identity.ts
```diff
@@ -0,0 +1,111 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import * as firebase from 'firebase-admin';
+import * as _ from 'lodash';
```
inlined, line 24: ew. We generally don't allow lodash anymore. I guess I can let it slide because you're just moving code...

inlined (approved):

PR #1028, opened 2022-02-03
Add integration test for firebase-functions binary

taeold:
Add integration test for firebase-functions binary.

The integration setup is more robust than what we had before - it builds and installs the Functions SDK package at the branch commit and try to run the produced binary that ships with the Functions SDK.

We also modify the github CI/release script setting to run the integration test as part of its workflow.


.mocharc.yaml
```diff
@@ -1,10 +1,8 @@
 exit: true
 extension:
   - ts
-file:
-  - mocha/setup.ts
 package: ./package.json
 reporter: spec
 require:
   - 'ts-node/register'
-spec: spec/**/*.spec.ts
```
colerogers, line 10: removing this just means that we now have test files outside the `spec/` dir and they don't need spec in the title, correct?

colerogers (approved):
Nice stuff! yay tests!

.mocharc.yaml
taeold: Yes - having the spec defined here made it impossible for me to not run them for integration tests. So we now pass in the path/glob of test we want to run in the npm script definition (in package.json) itself.

colerogers: Ahhh the changed line - `"test": "mocha --file ./mocha/setup.ts spec/**/*.spec.ts ",`. Thanks!

inlined: Hmm... there's no other way to achieve our goal? Changing mocha to need a --file means that running an individual file's tests with `npx mocha` got harder.

scripts/bin-test/test.ts
```diff
@@ -0,0 +1,233 @@
...
+};
+
+async function retryUntil(
+  fn: () => Promise<boolean>,
+  timeoutMs: number,
+  sleepMs: number = TIMEOUT_S
+) {
+  const sleep = () => {
+    return new Promise<void>((resolve) => {
+      setTimeout(() => resolve(), sleepMs);
```
inlined, line 59: nit: you can just pass a function rather than a lambda that calls the function:

```
setTimeout(resolve, sleepMs)
```

scripts/bin-test/test.ts
```diff
@@ -0,0 +1,233 @@
...
+    let cleanup: () => Promise<void>;
+
+    before(async () => {
+      const r = await startBin(tc);
+      port = r.port;
+      cleanup = r.cleanup;
+    });
+
+    after(async () => {
+      await cleanup?.();
```
inlined, line 147: TIL `.()` exists. That's neat.

scripts/bin-test/test.ts
```diff
@@ -0,0 +1,233 @@
...
+    ];
+
+    for (const tc of testcases) {
+      describe(tc.name, async () => {
+        await runTests(tc);
+      });
+    }
+  }).timeout(TIMEOUT_L);
+
+  if (semver.gt(process.versions.node, '13.2.0')) {
```
inlined, line 205: nit: since 13 isnt' a supported major it would have been fine to say `'14'`

.mocharc.yaml
taeold: Hmm don't you still need to pass individual file via `npx mocha`, e.g. `npx mocha spec/path/to/my/test.ts`?

I'm not sure what devx I just broke here. I don't know if other ways exists, but happy to explore.

PR #1045, opened 2022-02-21
Fix handler typo in billing alerts

colerogers:
Replaces `onAutomatedPlanUpdatePublished()` with the correct `onPlanAutomatedUpdatePublished()` in billing alerts.

### DIFF
--- a/src/v2/providers/alerts/billing.ts
+++ b/src/v2/providers/alerts/billing.ts
@@ -32,7 +32,7 @@ export type BillingEvent<T> = CloudEvent<FirebaseAlertData<T>, WithAlertType>;
 /** @internal */
 export const planUpdateAlert = 'billing.planUpdate';
 /** @internal */
-export const planAutomatedUpdateAlert = 'billing.automatedPlanUpdate';
+export const planAutomatedUpdateAlert = 'billing.planAutomatedUpdate';
 
 /**
  * Declares a function that can handle a billing plan update event.
--- a/src/v2/providers/alerts/billing.ts
+++ b/src/v2/providers/alerts/billing.ts
@@ -60,18 +60,18 @@ export function onPlanUpdatePublished(
 /**
  * Declares a function that can handle an automated billing plan update event.
  */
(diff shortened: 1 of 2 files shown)


taeold (changes requested):
Let's change https://github.com/firebase/firebase-functions/pull/1045/files#diff-4b02d9f4f18deca9c29bad26c9ca35e17a5914eabbd4a0ab7b1109b2a77b505dR35 too.

src/v2/providers/alerts/billing.ts
```diff
@@ -32,7 +32,7 @@ export type BillingEvent<T> = CloudEvent<FirebaseAlertData<T>, WithAlertType>;
 /** @internal */
 export const planUpdateAlert = 'billing.planUpdate';
 /** @internal */
-export const automatedPlanUpdateAlert = 'billing.automatedPlanUpdate';
+export const planAutomatedUpdateAlert = 'billing.automatedPlanUpdate';
```
taeold, line 35: huh this is bit odd. The event type is still `automatedPlanUpdate` not `planAutomatedUpdate`?

taeold (approved):

src/v2/providers/alerts/billing.ts
colerogers: Just double checked the alert type string, it's set to `billing.automatedPlanUpdate`. Since all their other methods and structures have this as planAutomatedUpdate, I'm going to request that they change this value on the provider end. I'll leave this PR up until it gets sorted out.

inlined: What's in the spec? IDK if we should differ between the alerttype and the handler function name. It just seemed odd that this isn't object -> verb naming.

colerogers: @inlined @taeold I just submitted a change to the alerttype parameter to set the value to `billing.planAutomatedUpdate`

taeold (approved):

PR #1049, opened 2022-02-28
Add ts-comment on how to remove maxInstances limit for functions

donalffons:


donalffons:
Hi there! I just tested and verified that it is possible to remove the `maxInstances` limit on firebase functions by passing `0` for that value. I think this is an important function (not being able to remove the limit seems like a missing feature), but I couldn't find any documentation about this anywhere.

PR #1056, opened 2022-03-10
Fix for #1055

### LINKED ISSUE #1055: RTDB function - unable to use DataSnapshots .ref when a path component starts with refs
If you've got a Firebase RTDB path component that starts with `refs` you won't be able to use `.ref` of the provided `DataSnapshot` that's passed into the function.

Bit of a fun one this is...

### [REQUIRED] Version info

**node:** 14.17.4
**firebase-functions:** 3.18.1
**firebase-tools:** 10.1.0
**firebase-admin:** 10.0.2

### [REQUIRED] Test case

Deploy the below function:

```js
const functions = require('firebase-functions')

exports.pathUnableToUseRef = functions.database.ref('childStartsWithRef/{id}').onWrite(async (change, context) => {
   functions.logger.debug(`Change at on child of childStartsWithRef: change.after.key='${change.after.key
      }', context.params.id='${context.params.id}'`, { context })
   functions.logger.debug(`change.after.ref.toString(): ${change.after.ref.toString()}`)
})
```

Deploy the above function and then run:

`firebase database:set "/childStartsWithRef/ref" --data '"foobar"'` 

After checking the function logs, you can see this executes the function successfully:

```
Function execution started
Change at on child of childStartsWithRef: change.after.key='ref', context.params.id='ref'
change.after.ref.toString(): https://my-project.firebaseio.com/childStartsWithRef/ref
Function execution took 6 ms, finished with status: 'ok'
```

However, when running:
`firebase database:set "/childStartsWithRef/refs" --data '"foobar"'` 

This causes the function execution to fail.

For the `.ref` access to work and the following to be outputted on the second execution:

```
Function execution started
Change at on child of childStartsWithRef: change.after.key='refs', context.params.id='refs'
change.after.ref.toString(): https://my-project.firebaseio.com/childStartsWithRef/refs
Function execution took 6 ms, finished with status: 'ok'
```

Second execution of function fails:

```
Function execution started
Change at on child of childStartsWithRef: change.after.key='null', context.params.id='refs'
[2022-03-09T15:03:33.858Z]  @firebase/database: FIREBASE FATAL ERROR: Cannot parse Firebase url. Please use https://<YOUR FIREBASE>.firebaseio.com 
Function execution took 93 ms, finished with status: 'error'
```

Note also that the `.key` of the `DataSnapshot` is `null`.

Yes.

### Extra info

Pretty sure its because of the regex to extract the instance and path from the resource:
https://github.com/firebase/firebase-functions/blob/f45e45f28294f578c22f2e26f65c16cf1563720f/src/providers/database.ts#L323

```js
const resource = 'projects/_/instances/my-database/refs/childStartsWithRef/refs'
const resourceRegex = `projects/([^/]+)/instances/([a-zA-Z0-9\-^/]+)/refs(/.+)?`;
const match = resource.match(new RegExp(resourceRegex));
console.log(match)
```

Outputs:

```
[
  'projects/_/instances/my-database/refs/childStartsWithRef/refs',
  '_',
  'my-database/refs/childStartsWithRef',
  undefined,
  index: 0,
  input: 'projects/_/instances/my-database/refs/childStartsWithRef/refs',
  groups: undefined
]
```

But changing the instance path component regex to being non-greedy (`([a-zA-Z0-9\-^/]+?)`) does fix things - I'm not saying this is the best fix, but certainly fixes the example case:

```js
const resource = 'projects/_/instances/my-database/refs/childStartsWithRef/refs'
const resourceRegex = `projects/([^/]+)/instances/([a-zA-Z0-9\-^/]+?)/refs(/.+)?`;
const match = resource.match(new RegExp(resourceRegex));
console.log(match)
```

Outputs:

```
[
  'projects/_/instances/my-database/refs/childStartsWithRef/refs',
  '_',
  'my-database',
  '/childStartsWithRef/refs',
  index: 0,
  input: 'projects/_/instances/my-database/refs/childStartsWithRef/refs',
  groups: undefined
]
```

rhodgkins:
Fixes #1055

### DIFF
--- a/spec/v1/providers/database.spec.ts
+++ b/spec/v1/providers/database.spec.ts
@@ -529,6 +529,22 @@ describe('Database Functions', () => {
       expect(path).to.equal('/bar');
     });
 
+    it('should return the correct instance and path strings if root path is /refs', () => {
+      const [instance, path] = database.extractInstanceAndPath(
+        'projects/_/instances/foo/refs/refs'
+      );
+      expect(instance).to.equal('https://foo.firebaseio.com');
+      expect(path).to.equal('/refs');
+    });
+
+    it('should return the correct instance and path strings if a child path contain /refs', () => {
+      const [instance, path] = database.extractInstanceAndPath(
+        'projects/_/instances/foo/refs/root/refs'
+      );
+      expect(instance).to.equal('https://foo.firebaseio.com');
+      expect(path).to.equal('/root/refs');
+    });
+
     it('should return the correct multi-region instance and path strings if domain is present', () => {
       const [instance, path] = database.extractInstanceAndPath(
         'projects/_/instances/foo/refs/bar',
--- a/src/providers/database.ts
+++ b/src/providers/database.ts
@@ -307,6 +307,8 @@ export class RefBuilder {
   };
 }
 
+const resourceRegex = /^projects\/([^/]+)\/instances\/([a-zA-Z0-9-]+)\/refs(\/.+)?/;
+
 /**
  * Utility function to extract database reference from resource string
  *
@@ -320,7 +322,6 @@ export function extractInstanceAndPath(
   resource: string,
   domain = 'firebaseio.com'
 ) {
-  const resourceRegex = `projects/([^/]+)/instances/([a-zA-Z0-9\-^/]+)/refs(/.+)?`;
   const match = resource.match(new RegExp(resourceRegex));
   if (!match) {
     throw new Error(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Add support for more regions and memory for v2 functions (#1037).
+- Fixes bug where some RTDB instance names were incorrectly parsed (#1056).


rhodgkins:
@taeold all sorted. Added extra couple of test cases too. 

I've pulled out the regex from the function too - I couldn't see any good reason for it being a local variable but can change if you'd prefer it left as it was!

spec/v1/providers/database.spec.ts
```diff
@@ -529,6 +529,22 @@ describe('Database Functions', () => {
...
+    });
+
+    it('should return the correct instance and path strings if a child path contain /refs', () => {
+      const [instance, path] = database.extractInstanceAndPath(
+        'projects/_/instances/foo/refs/root/refs'
+      );
+      expect(instance).to.equal('https://foo.firebaseio.com');
+      expect(path).to.equal('/root/refs');
+    });
+
```
taeold, line 547: ❤️  Thank you for adding tests!

src/providers/database.ts
```diff
@@ -307,6 +307,8 @@ export class RefBuilder {
   };
 }
 
+const resourceRegex = /^projects\/([^/]+)\/instances\/([a-zA-Z0-9-]+?)\/refs(\/.+)?/;
```
taeold, line 310: I'm not convinced (yet!) that we need non-greedy operator on the capturing group for instance based on the provided test cases:

e.g.
```js
const greedy = /^projects\/([^/]+)\/instances\/([a-zA-Z0-9-]+)\/refs(\/.+)?/
const nongreedy = /^projects\/([^/]+)\/instances\/([a-zA-Z0-9-]+?)\/refs(\/.+)?/

console.log("projects/_/instances/foo/refs/root/refs".match(greedy)[2]) //  => foo
console.log("projects/_/instances/foo/refs/root/refs".match(nongreedy)[2]) // => foo

console.log("projects/_/instances/foo/refs/refs".match(greedy)[2])  // => foo
console.log("projects/_/instances/foo/refs/refs".match(nongreedy)[2]) // => foo
```

Can you share edge cases that require non-greedy operator?

taeold (commented):
Thank you!

Can you add a changelog entry? Something like

`Fixes bug where some RTDB instance names were incorrectly parsed (#1056).`

src/providers/database.ts
rhodgkins: Yeh I think you're right, now we've taken out the `/` from the set in the instances capture group the non-greediness is no longer needed - shall update 👍

rhodgkins:
@taeold done - regex changed back to greedy and change log entry added 👍

taeold (approved):
🚀

rhodgkins:
Thanks @taeold for the quick turn around, any ideas on an ETA for a release?

rhodgkins:
@taeold any updates - or by "next Monday" you meant the 21st? (Sorry for being persistent but this is a big issue for us as we've had a user created with their `uid` starting with `refs` so all the functions triggered for that user fail!)

taeold:
@rhodgkins Sorry for the delay. This change was released today!

PR #1060, opened 2022-03-18
types: support function builder to have typed `context.params` based on `path`

pikax:
### Description

Support for better typescript when using FunctionBuilder, allowing to resolve the `context.params` based on the `path` and allow passing the type to `onCreate`, etc on `firebase`.

### Code sample

```ts
declare function expect<T>(a: T) :void

region("europe-west3")
.firestore
.document('test/{id}/{collection}/{childId}')
.onCreate<{ myDoc: number }>((snap, context)=> {
  expect<number>(snap.data().myDoc)
  // @ts-expect-error doesn't exist
  snap.data().error
  expect<{
    id: string,
    collection: string,
    childId:string
  }>(context.params)

  // @ts-expect-error doesn't exist
  context.params.error
})

```
image

## NOTES
This only applies to the `firestore` but we can support the other ones. 

- it would be nice to have types test in place (eg: https://github.com/vuejs/core/tree/main/test-dts)
- This should be backwards compatible and the only errors is when params are defined but the wrong param is accessed

### DIFF
--- a/src/function-builder.ts
+++ b/src/function-builder.ts
@@ -295,10 +295,10 @@ export function runWith(runtimeOptions: RuntimeOptions): FunctionBuilder {
     return new FunctionBuilder(runtimeOptions);
   }
 }
-
+  
 type ExtractParams<T> = T extends `${infer Pre}{${infer A}}${infer Post}`
   ? { [K in A]: string } & ExtractParams<Pre> & ExtractParams<Post>
-  : {};
+  : string extends T ? Record<string,string> :  {};
 
 export class FunctionBuilder {
   constructor(private options: DeploymentOptions) {}
--- a/src/function-builder.ts
+++ b/src/function-builder.ts
@@ -296,6 +296,10 @@ export function runWith(runtimeOptions: RuntimeOptions): FunctionBuilder {
   }
 }
 
+type ExtractParams<T> = T extends `${infer Pre}{${infer A}}${infer Post}`
+  ? { [K in A]: string } & ExtractParams<Pre> & ExtractParams<Post>
+  : {};
+
 export class FunctionBuilder {
   constructor(private options: DeploymentOptions) {}
 
@@ -427,8 +431,13 @@ export class FunctionBuilder {
        * collection is named "users" and the document is named "Ada", then the
        * path is "/users/Ada".
        */
-      document: (path: string) =>
-        firestore._documentWithOptions(path, this.options),
+
+      
+      document: <T extends string>(path: T) =>
+        firestore._documentWithOptions<ExtractParams<T>>(
+          path,
+          this.options
+        ),
 
       /** @hidden */
       namespace: (namespace: string) =>
--- a/src/cloud-functions.ts
+++ b/src/cloud-functions.ts
@@ -68,7 +68,9 @@ export interface Event {
  * - Authorization of the request that triggered the event, if applicable and
  *   available.
  */
-export interface EventContext {
+export interface EventContext<
+  Params extends Record<string, string> = Record<string, string>
+> {
   /**
    * Authentication information for the user that triggered the function.
    * This object contains `uid` and `token` properties for authenticated users.
@@ -130,7 +132,7 @@ export interface EventContext {
    * provided to the [`ref()`](providers_database_.html#ref) method for a Realtime
    * Database trigger. Cannot be accessed while inside the handler namespace.
    */
-  params: { [option: string]: any };
+  params: Params;
 
   /**
    * The resource that emitted the event. Valid values are:
--- a/src/providers/firestore.ts
+++ b/src/providers/firestore.ts
@@ -43,8 +43,12 @@ export const service = 'firestore.googleapis.com';
 /** @hidden */
 export const defaultDatabase = '(default)';
 let firestoreInstance: any;
-export type DocumentSnapshot = firebase.firestore.DocumentSnapshot;
-export type QueryDocumentSnapshot = firebase.firestore.QueryDocumentSnapshot;
+export type DocumentSnapshot<
+  T = { [field: string]: any }
+> = firebase.firestore.DocumentSnapshot<T>;
+export type QueryDocumentSnapshot<
+  T = { [field: string]: any }
+> = firebase.firestore.QueryDocumentSnapshot<T>;
 
 /**
  * Select the Firestore document to listen to for events.
@@ -84,8 +88,13 @@ export function _namespaceWithOptions(
 }
 
 /** @hidden */
-export function _documentWithOptions(path: string, options: DeploymentOptions) {
-  return _databaseWithOptions(defaultDatabase, options).document(path);
+export function _documentWithOptions<Params extends Record<string, string>>(
+  path: string,
+  options: DeploymentOptions
+) {
+  return _databaseWithOptions(defaultDatabase, options).document<Params>(
+    path
+  );
 }
 
 export class DatabaseBuilder {
@@ -96,8 +105,8 @@ export class DatabaseBuilder {
     return new NamespaceBuilder(this.database, this.options, namespace);
   }
 
-  document(path: string) {
-    return new NamespaceBuilder(this.database, this.options).document(path);
+  document<Params extends Record<string, string>>(path: string) {
+    return new NamespaceBuilder(this.database, this.options).document<Params>(path);
   }
 }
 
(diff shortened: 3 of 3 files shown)


src/function-builder.ts
```diff
@@ -295,6 +295,10 @@ export function runWith(runtimeOptions: RuntimeOptions): FunctionBuilder {
     return new FunctionBuilder(runtimeOptions);
   }
 }
+  
+type ExtractParams<T> = T extends `${infer Pre}{${infer A}}${infer Post}`
+  ? { [K in A]: string } & ExtractParams<Pre> & ExtractParams<Post>
+  : string extends T ? Record<string,string> :  {};
```
pikax, line 301:
```suggestion
  :  {};
```
Update to this to force the `context.params` to be `{}` if no params found - aka more strict

inlined:
This is an interesting idea and I'd like to play with it. I'm not convinced in the value of making `EventContext` take an input generic type which simply casts `context.params` is useful. It doesn't seem any more powerful than simply casting `context.params`. OTOH, this might be (more) possible to do automatically in the v2 API using string literal types. I'll play with it tonight.

inlined:
I've got a rough working draft for how we can make this work in the V2 API without the user providing any explicit template parameters. I'd have to look at whether it's possible to backport this to V1, which is not our top priority anymore.

pikax:
> I've got a rough working draft for how we can make this work in the V2 API without the user providing any explicit template parameters

The explicit template parameters are only required for to decorate the return `data()` otherwise you need always to cast it.

inlined:
Yeah, that's WAI. In the v2 SDK for Firestore I'd like to look at working with converters, since that's the normal SDK's solution. Putting an interface type as an input generic argument and casting as that interface type are functionally equivalent.

pikax:
Is not about functionality because you can use the same argument as using `any` as function parameters and casting is functionally the same... Which is sort of true but the code glow is really damaged, why look at the implementation of the trigger or function to know what type are you using?

Not to mention that is much easier to make mistakes with casting, because if you casting multiple times the overhead of updating is greater.

On my project I've wrappers to firebase function to be type safe on my project, I thought it would be a good idea to bring it to the firebase, since typescript support in firebase is lacking to say the least.

inlined:
Return-type only generics are generally frowned upon (and, in fact, banned in Google codebases) because they give the appearance of type safety but are really hiding an `any` conversion (it's literally the only way to implement those interfaces).

As mentioned before, I'm curious to investigate adding [`FirestoreDataConverter`]( https://firebase.google.com/docs/reference/node/firebase.firestore.FirestoreDataConverter) support to our V2 API, but we will not add an output-only template parameter because it gives the illusion of type safety. You're equally well protected stating:

```
const doc = snap.data() as { myDoc: number };
```

One would argue you're _safer_ because the conversion from a no-schema document to a well typed struct is called out as dangerous/an assumption because the `as` keyword is in your codebase instead of ours.

pikax:
> Return-type only generics are generally frowned upon (and, in fact, banned in Google codebases) because they give the appearance of type safety but are really hiding an `any` conversion

That explains why is so hard to implement a type safe code with `firebase*`, IMO is a bad decision to straight out ban something. 

I don't really agree with this statement of "Return-type only generics", since it exists for a reason, is a pretty common OOP to return based on what you receive:
```csharp
using System;
					
public class Program
{
	public static void Main()
	{
		Console.WriteLine("Hello World" + Program.Something(155));
	}
	
	public static T Something<T>(T item) {
          // do something to item
           return item;
	}
}
```



> they give the appearance of type safety but are really hiding an `any` conversion (it's literally the only way to implement those interfaces).

The implementation code would stay the same, you can use overloads to provide a nice function typing for usage and allow the users to have a greater UX. 
Firebase users (such as myself) don't really need to care much about how the code is implemented, we only care if the code is working and how we can use it, currently the code works, but the code is not very flexible to allow users to straightforwardly type their code.


 

> As mentioned before, I'm curious to investigate adding [`FirestoreDataConverter`](https://firebase.google.com/docs/reference/node/firebase.firestore.FirestoreDataConverter) support to our V2 API, but we will not add an output-only template parameter because it gives the illusion of type safety.

I wouldn't add javascript runtime code to decorate my types, that doesn't sound good at all, converters are to _convert_ data not to _decorate_ the type. Because this wouldn't work for [Cloud Firestore Triggers](https://firebase.google.com/docs/functions/firestore-events) unless you want to change the API just for adding types... and to be honest [firebase-functions are already slow enough for many years](https://issuetracker.google.com/issues/158014637)




> You're equally well protected stating:
> 
> ```
> const doc = snap.data() as { myDoc: number };
> ```

You can literally use that argument to anything in typescript:
```ts
function doSomethingComplex(item: DocumentData, item2: DocumentData) {
 const item_typed= item as { myDoc: number };
 const item2_typed = item2 as { myDoc: number, betterDoc: string }
}
```
We can say "You're equally well protected" using this, but the usage is plain rubbish, if you intend to write type safe code like this... there's a reason why typescript is so popular.

In terms of actual usage within `firebase-functions`:
```ts
if(snap.before.exists) {
  const doc = snap.before.data() as { myDoc: number };
  // many lines of code 
  if(myCondition && snap.after.exists) {
    const newDoc = snap.after.data() as { myDoc: number}
    // many lines of code
    const processedDoc = processDoc(newDoc.data()); // what type are we passing and returning?
    // many lines of code
    const diff = diffObject(doc, processedDoc)  // no type safe here.
  }
} else {
 // basically the same code
}
```
You can argue that we could and should create a `type` or `interface` but if the function is huge that won't solve much, because there's castings everywhere, which is a pain to keep track of the correct type, not even mention the visual pollution of using `as MyType`



> One would argue you're _safer_ because the conversion from a no-schema document to a well typed struct is called out as dangerous/an assumption because the `as` keyword is in your codebase instead of ours.

We can argue what is _safer_, but there's no argument to which is more user/DX friendly.
By declaring/passing the type on **one single** line, I would/will argue that's _safer_ than repeating the code in as many lines you need.


> because the `as` keyword is in your codebase instead of ours.

You don't need to use `as` in your codebase, if you need to use it constantly, that means your types are not correct and your codebase is not as healthy as you might think, there's also function overloads in which you can have your code not fully typed and having overloads which will expose the correct type based on the arguments.





--- 

I'll give an example of poor typescript in this [codebase](https://github.com/firebase/firebase-functions/blob/master/src/providers/firestore.ts#L46-L47):
```ts
export type DocumentSnapshot = firebase.firestore.DocumentSnapshot;
export type QueryDocumentSnapshot = firebase.firestore.QueryDocumentSnapshot;
```

This is so wrong because by doing that we cannot decorate the first template argument of [QueryDocumentSnapshot](https://github.com/googleapis/nodejs-firestore/blob/ce9abfd8cd65d0d2be4dd7f6a56fb679a100296a/dev/src/document.ts#L565), which limits the user _a lot_. 
You might say, `why don't you import QueryDocumentSnapshot from firebase admin and voila`
Which I might say: Good luck keeping the dependencies matching with each other, typescript doesn't like multiple versions of the same dependencies being used, you will most likely need to cast as any or add `//@ts-expect-error` 

This is just one of the many examples where it makes very hard to someone use firebase typesafely. 


---
 
Just to make it clear not trying to be aggressive or trash on code, I just don't agree on your statements, I appreciate you explaining the reasons why is not implemented, this conversation can easily be terminated by stating "this is a team/product choice" :D

--- 

The code that I use on my codebase:
```ts
export type EventContext<TPath extends string> = Omit<
  functions.EventContext,
  "params"
> & {
  params: ExtractParams<TPath>;
};

export function document<TPath extends string>(
  path: TPath,
  runtimeOptions: functions.RuntimeOptions = {}
): {
  onCreate: <T>(
    handler: (
      snapshot: QueryDocumentSnapshot<T>,
      context: EventContext<TPath>
    ) => PromiseLike<any> | any
  ) => CloudFunction<functions.Change<DocumentSnapshot>>;

  onDelete: <T>(
    handler: (
      snapshot: QueryDocumentSnapshot<T>,
      context: EventContext<TPath>
    ) => PromiseLike<any> | any
  ) => CloudFunction<functions.Change<QueryDocumentSnapshot>>;
  onUpdate: <T>(
    handler: (
      snapshot: functions.Change<QueryDocumentSnapshot<T>>,
      context: EventContext<TPath>
    ) => PromiseLike<any> | any
  ) => CloudFunction<functions.Change<QueryDocumentSnapshot>>;

  onWrite: <T>(
    handler: (
      snapshot: functions.Change<QueryDocumentSnapshot<T>>,
      context: EventContext<TPath>
    ) => PromiseLike<any> | any
  ) => CloudFunction<functions.Change<QueryDocumentSnapshot>>;
} {
  const d = functions
    .region("europe-west3")
    .runWith(runtimeOptions)
    .firestore.document(path);

  return d as any;
}
``` 

I know there's an `as any` but that can easily be solve by allowing [DocumentBuilder](https://github.com/firebase/firebase-functions/blob/master/src/providers/firestore.ts#L197) to be templated. 
Not saying this the ideal or the perfect implementation, but so far it works, I had already problems when updating dependencies which caused some problems with typescript.

PR #1066, opened 2022-03-26
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
### Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/src/common/providers/https.ts
+++ b/src/common/providers/https.ts
@@ -437,7 +437,7 @@ function isValidRequest(req: Request): req is HttpRequest {
   // If it has a charset, just ignore it for now.
   const semiColon = contentType.indexOf(';');
   if (semiColon >= 0) {
-    contentType = contentType.substr(0, semiColon).trim();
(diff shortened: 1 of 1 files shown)


taeold:
@CommanderRoot Hey! Thank you for your contribution.

I realize that Functions SDK is not setup to receive external contribution - we'd need to do some work in order to comply with internal policies. I'll keep you posted when we get that setup completed.

taeold:
@CommanderRoot I was wrong - we do have CLA setup in this repo. Thank you for your contribution!

taeold (approved):

PR #1069, opened 2022-03-30
Move some TQ triggers under new namespace

taeold:
Move all task queue related API under newly formed `tasks` package. Task queue function are now under `tasks` namespace:

```
export const factorize = functions.tasks.taskQueue({
  rateLimits: {
    maxBurstSize: 2,
    maxConcurrentDispatches: 1,
    maxDispatchesPerSecond: 1,
  },
  retryConfig: {
    maxAttempts: 100,
    maxBackoffSeconds: 20,
    maxDoublings: 10,
    minBackoffSeconds: 0.1,
  }
}).onDispatch((data, context) => {
  deducePrivateKey(data.publicKey);
});
```

Same goes for handler namespace.

We also add probably-forgotten `maxRetrySeconds` option for task queues.


spec/helper.ts
```diff
@@ -0,0 +1,127 @@
+// The MIT License (MIT)
```
inlined, line 1: nit: helper is a dirty word. Can we at least call it a `fooHelper`? Maybe `httpsBasedHelper`

inlined (commented):
Admittedly I scanned this one since much of it should be cut & paste.

PR #1074, opened 2022-04-12
Modernize integration test

taeold:
Small touchups to the existing integration test:

1) Use `node-fetch` to simplify HTTP calls
2) Group v1 triggers (in prep for running v2 trigger tests)
3) Replace use of `console.log` to `functions.logger.info`
4) Get rid of node10 as test target


integration_test/functions/src/index.ts
```diff
@@ -1,83 +1,116 @@
...
-export * from './database-tests';
-export * from './auth-tests';
-export * from './firestore-tests';
-export * from './https-tests';
-export * from './remoteConfig-tests';
-export * from './storage-tests';
-export * from './testLab-tests';
-const numTests = Object.keys(exports).length; // Assumption: every exported function is its own test.
+import * as v1 from './v1/index';
+const numTests = Object.keys(v1).length; // Assumption: every exported function is its own test.
```
inlined, line 9: If you wanted to avoid this assumption you could technically filter for those functions which had `__trigger` or `__endpoint`

integration_test/functions/src/v1/database-tests.ts
```diff
@@ -40,7 +40,7 @@ export const databaseTests: any = functions
             return expectMatches(
               url,
               new RegExp(
-                `^https://${process.env.GCLOUD_PROJECT}.firebaseio.com/dbTests`
+                `^https://${process.env.GCLOUD_PROJECT}-default-rtdb.firebaseio.com/dbTests`
```
inlined, line 43: Let's make `-default-rtdb` optional since it's only there for new projects.

integration_test/functions/src/v1/database-tests.ts
```diff
@@ -52,7 +52,7 @@ export const databaseTests: any = functions
       .it('should have refs resources', (change, context) =>
         expectEq(
           context.resource.name,
-          `projects/_/instances/${process.env.GCLOUD_PROJECT}/refs/dbTests/${context.params.testId}/start`
+          `projects/_/instances/${process.env.GCLOUD_PROJECT}-default-rtdb/refs/dbTests/${context.params.testId}/start`
```
inlined, line 55: again, make `-default-rtdb` optional

inlined (approved):

PR #1077, opened 2022-04-15
Add CPU option. Rip out deprecated apiVersion

inlined:
Also fixes a couple of build errors I was getting in identity.ts

### DIFF
--- a/integration_test/functions/src/v1/testLab-utils.ts
+++ b/integration_test/functions/src/v1/testLab-utils.ts
@@ -1,5 +1,5 @@
-import fetch from 'node-fetch';
 import * as admin from 'firebase-admin';
+import fetch from 'node-fetch';
 
 interface AndroidDevice {
   androidModelId: string;
--- a/src/runtime/loader.ts
+++ b/src/runtime/loader.ts
@@ -69,7 +69,7 @@ export function extractStack(
     const val: any = valAsUnknown;
     if (
       typeof val === 'function' &&
-      (val as any).__endpoint &&
+      val.__endpoint &&
       typeof val.__endpoint === 'object'
     ) {
       const funcName = prefix + name;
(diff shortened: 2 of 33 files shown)


integration_test/functions/src/v1/testLab-utils.ts
```diff
@@ -1,5 +1,5 @@
-import fetch from 'node-fetch';
 import * as admin from 'firebase-admin';
+import fetch from 'node-fetch';
```
taeold, line 2: q: do you have some eslint locally that automatically formats this? (or is it something you did manually)

src/runtime/loader.ts
```diff
@@ -64,19 +64,21 @@ export function extractStack(
...
 ) {
-  for (const [name, val] of Object.entries(module)) {
+  for (const [name, valAsUnknown] of Object.entries(module)) {
+    // We're introspecting untrusted code here. Any is appropraite
+    const val: any = valAsUnknown;
     if (
       typeof val === 'function' &&
-      val['__endpoint'] &&
-      typeof val['__endpoint'] === 'object'
+      (val as any).__endpoint &&
```
taeold, line 72: nit* is `(val as any)` required here? I'd guess it's already any based on L69.

src/v2/options.ts
```diff
@@ -167,9 +167,21 @@ export interface GlobalOptions {
...
    * Can only be applied to functions running on Cloud Functions v2.
-   * A value of null restores the default concurrency.
+   * A value of null restores the default concurrency (80 when CPU >= 1, 1 otherwise).
+   * Concurrency cannot be set to any value other than 1 if `cpu` is less than 1.
    */
   concurrency?: number | null;
+
+  /**
+   * Fractional number of CPUs to allocate to a function.
+   * defaults to 1 for functions with <= 2GB RAM and increases for larger memory sizes.
```
taeold, line 177:
```suggestion
   * Defaults to 1 for functions with <= 2GB RAM and increases for larger memory sizes.
```

taeold (approved):

integration_test/functions/src/v1/testLab-utils.ts
inlined: npm run format:fix && npm run lint:fix

src/runtime/loader.ts
inlined: You're right. I was doing this manually and then used the any cast above. Will remove.

PR #1079, opened 2022-04-16
Add support for secrets in v2

taeold:
In addition to adding secrets to v2 `__endpoint` annotation, I'm also making following (small) changes:

* Enforce that `platform` be part of trigger annotation - Firebase CLI already assumes that it's included.
* Endpoint secret environment variable annotation (for v1 & v2)


src/v2/options.ts
```diff
@@ -239,6 +244,15 @@ export function getGlobalOptions(): GlobalOptions {
   return globalOptions || {};
 }
 
+/**
+ * Options that can be set on an individual HTTPS Cloud Function.
+ */
+export interface HttpsOptions extends Omit<GlobalOptions, 'region'> {
```
colerogers, line 250: What do we think about also moving event handler options here and exporting them from their path?

colerogers (approved):

src/v2/options.ts
taeold: Yes - more breaking changes for v2 users before IO!

inlined: I think it should probably stay in https because it lines up with the fact that PubSubOptions are, for example, in pubsub.ts

src/v2/options.ts
```diff
@@ -350,6 +365,13 @@ export function optionsToEndpoint(
     }
     return region;
   });
+  convertIfPresent(
+    endpoint,
+    opts,
+    'secretEnvironmentVariables',
+    'secrets',
+    (secrets) => secrets.map((secret) => ({ secret, key: secret }))
```
inlined, line 373: I'm confused. Isn't this specifying both the env var and the secret name? DIdn't we not want to do that in the SDK but in the CLI since (in the future) the build process needs to not set the secret name so that extensions can?

inlined (approved):
I personally wouldn't have moved `HttpsOptions`

src/v2/providers/https.ts
```diff
@@ -36,14 +36,6 @@ import * as options from '../options';
 
 export { Request, CallableRequest, FunctionsErrorCode, HttpsError };
 
-export interface HttpsOptions extends Omit<options.GlobalOptions, 'region'> {
```
colerogers, line 39: If we go down this path, we need to import & re-export `HttpsOptions` from here so it's not a breaking change.

colerogers (changes requested):

src/v2/options.ts
```diff
@@ -215,6 +215,11 @@ export interface GlobalOptions {
    * Invoker to set access control on https functions.
    */
   invoker?: 'public' | 'private' | string | string[];
+
+  /*
+   * Secrets to bind to a functions.
+   */
+  secrets?: string[];
```
jhuleatt, line 222: Can secrets be set globally (since this type is called `GlobalOptions`)?

inlined: Yes, though it's a good way to cause a minor bill increase and lower security.

jhuleatt: Yeah, we should definitely encourage per-function. Interesting to know that it's possible globally though, thanks!

src/v2/options.ts
taeold: ah excellent catch.

colerogers (approved):

PR #1080, opened 2022-04-18
Add auth blocking handlers

colerogers:
**Still a WIP until Admin SDK changes are released, the build will FAIL until then.**

- Adds v1 & v2 handlers for auth blocking triggers
- Adds `blockingTrigger` to Trigger & Endpoint annotations
- Adds reference to Admin SDK blocking token verification
- Yanks custom rolled token verification & decoding
- Adds Unit tests


src/common/providers/identity.ts
```diff
@@ -342,6 +325,12 @@ export interface AuthEventContext extends EventContext {
   credential?: Credential;
 }
 
+/** Defines the auth event for v2 blocking events */
+
```
taeold, line 329: nit* extra space?

src/common/providers/identity.ts
```diff
@@ -1044,23 +818,43 @@ function wrapHandler(
...
+        : await apps()
+            .admin.auth()
+            ._verifyAuthBlockingToken(req.body.data.jwt);
+
       const authUserRecord = parseAuthUserRecord(decodedPayload.user_record);
       const authEventContext = parseAuthEventContext(decodedPayload, projectId);
-      const authResponse =
-        (await handler(authUserRecord, authEventContext)) || undefined;
+
+      const authResponse = v2
```
taeold, line 839: nit* I wonder if you can do this by checking the length of arguments expected by the handler rather than using additional argument e.g.

https://github.com/firebase/firebase-functions/blob/master/src/common/providers/https.ts#L736-L746

(This is arguably less clear what's going on)

src/common/providers/identity.ts
```diff
@@ -1072,9 +866,9 @@ function wrapHandler(
         err = new HttpsError('internal', 'An unexpected error occurred.');
       }
 
-      res.status(err.code);
-      res.setHeader('Content-Type', 'application/json');
-      res.send({ error: err.toJson() });
+      const { status } = err.httpErrorCode;
+      const body = { error: err.toJSON() };
```
taeold, line 870: Is there no need to set Content-Type header?

src/v2/providers/identity.ts
```diff
@@ -0,0 +1,233 @@
...
+  idToken?: boolean;
+  accessToken?: boolean;
+  refreshToken?: boolean;
+}
+
+/**
+ * Handle an event that is triggered before a user is created.
+ */
+export function beforeUserCreated(
+  handler: (
```
taeold, line 32: nit* The # of duplication here passes my threshold for defining a type!

src/v2/providers/identity.ts
```diff
@@ -0,0 +1,233 @@
...
+    optsOrHandler = {};
+  }
+
+  const { opts, accessToken, idToken, refreshToken } = getOpts(
+    optsOrHandler as BlockingOptions
+  );
+
+  const func: any = wrapHandler(eventType, handler, true);
+
+  const legacyEventType = `providers/cloud.auth/eventTypes/user.${eventType}`;
```
taeold, line 157: q: Is there non-legacy event type that's coming out soon?

src/v2/providers/identity.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  /** Trigger */
+  const baseOptsTrigger = options.optionsToTriggerAnnotations(
+    options.getGlobalOptions()
+  );
+  const specificOptsTrigger = options.optionsToTriggerAnnotations(opts);
+  func.__trigger = {
+    apiVersion: 2,
```
taeold, line 165: I don't think we use `apiVersion` anywhere. In fact, we can probably skip defining `__trigger` for any new v2 functions!

spec/v1/providers/auth.spec.ts
```diff
@@ -135,6 +143,190 @@ describe('Auth Functions', () => {
       });
     });
 
+    describe('beforeCreate', () => {
+      it('should create the function without options', () => {
+        wrapHandlerStub.returns(() => Promise.resolve());
```
taeold, line 148: q: Why is this stubbing needed for test🤔 ?

We are never calling the wrapped `fn` and I thought `wrapHandler` doesn't do anything "side-effect"-y unless the wrapped function is called.

spec/v2/providers/identity.spec.ts
```diff
@@ -0,0 +1,387 @@
...
+  refreshToken: false,
+  minInstances: 1,
+  region: 'us-west1',
+};
+
+describe('identity', () => {
+  let wrapHandlerStub: sinon.SinonStub;
+
+  beforeEach(() => {
+    wrapHandlerStub = sinon
```
taeold, line 56: similarly - is this stubbing really needed 🤔 ? We didn't have to do something like this when testing for callable function wrapping which seems very similar to auth blocking function wrapping.

tsconfig.release.json
```diff
@@ -7,6 +7,7 @@
     "noUnusedLocals": true,
     "outDir": "lib",
     "stripInternal": true,
+    "skipLibCheck": true,
```
taeold, line 10: q: what prompted this change?

src/v2/providers/identity.ts
colerogers: Unfortunately we'd lose out on typescript type completion by making this a separate type

src/v2/providers/identity.ts
colerogers: It's been thrown around that if v2 Auth Blocking functions get EventArc support, then we could update this event type

package.json
```diff
@@ -64,7 +64,8 @@
     "./v2/alerts/appDistribution": "./lib/v2/providers/alerts/appDistribution.js",
     "./v2/alerts/billing": "./lib/v2/providers/alerts/billing.js",
     "./v2/alerts/crashlytics": "./lib/v2/providers/alerts/crashlytics.js",
-    "./v2/eventarc": "./lib/v2/providers/eventarc.js"
+    "./v2/eventarc": "./lib/v2/providers/eventarc.js",
+    "./v2/identity": "./lib/v2/providers/identity"
```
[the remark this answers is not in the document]
colerogers: Fixed

src/common/providers/identity.ts
colerogers: fixed

src/common/providers/identity.ts
colerogers: Yeah that makes sense, i'll change it so we follow the same pattern

src/common/providers/identity.ts
colerogers: woops, fixed

src/v2/providers/identity.ts
colerogers: cool, yanked `__trigger` from the v2 namespace

spec/v1/providers/auth.spec.ts
colerogers: Ah you're right, when I had the custom verification I didn't want hack it so I put a stub. Removed it in the latest commit

spec/v2/providers/identity.spec.ts
colerogers: Ah you're right, when I had the custom verification I didn't want hack it so I put a stub. Removed it in the latest commit

tsconfig.release.json
colerogers: The Admin SDK updated the `@google-cloud/firestore` package which update the `google-gax` package and caused a nasty ts build error from the dependencies (https://github.com/firebase/firebase-functions/runs/6132285684?check_suite_focus=true), this change tries to silence that error.

Do you have a better suggestion on how to handle that?

src/v2/providers/identity.ts
taeold: wow that sucks

tsconfig.release.json
taeold: Yeah I'm a little worried that we are kicking the error down to the customer? Let's check to see if this error is legit worth fixing?

taeold: So I've learned that:

We could get around this issue by updating `tsconfig.release.json`:

```json
{
  "lib": ["es2018", "dom"]
}
```

This means that we are including all type definitions for the `dom` lib while we compile our project. [gax-nodejs](https://github.com/googleapis/gax-nodejs/blob/main/tsconfig.json) and [nodejs-firestore](https://github.com/googleapis/nodejs-firestore/blob/main/tsconfig.json) seems to do this. (Surprisingly, Firebase Admin SDK doesn't do this yet still seems to be fine)

[`AbortController`](https://nodejs.org/api/globals.html#class-abortcontroller) was introduced 14.17.0. This means that by adding `dom` we are sorta kinda implictly saying that you must use Node v14 and onward in order to use our SDK? Still thinking whether this is fine for us to do or we should just go ahead and try to use the `skipLibCheck` escape hatch. 🤔

colerogers: Yeah I found that also but didn't know if we should be adding all those definitions. Maybe @inlined has some thoughts about this?

taeold (approved):
lgtm - can you bring up skipLibCheck in maybe Firebase CLI chatroom to see if this is "dangerous"

inlined (commented):
Escalating the `dom` vs `skipLibCheck` concern. Hang tight.

tsconfig.release.json
inlined: Try removing `skipLibCheck` and re-add `node-fetch` to `package.json`. AFAICT it polyfills the missing types.

colerogers: @inlined build still fails with the same error

colerogers: thanks!

PR #1082, opened 2022-04-20
Bumping express patch version

colerogers:


inlined (commented):
What is the bug we're trying to work around with this min version bump?

FWIW, [the functions framework dependencies](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/package.json#L15-L24) list the _typings_ for `4.17.13` but the actual dependency is a whole minor less; `^4.16.4`. I wonder if we should have ever bumped to `4.17.x` since FF supports `4.16`.

inlined:
Ping

PR #1087, opened 2022-04-29
Update types for AlertPayloads

TheIronDev:
This commit updates the typing for the AlertPayloads.

There was a discrepency between the type definitions and what
prod was returning.

From:
com.google.firebase.firebasealerts.*

To
type.googleapis.com/google.events.firebase.firebasealerts.v1.*


taeold (approved):

inlined:
What a win to find these before our customers do!

PR #1091, opened 2022-05-02
Change type info to be inheritance friendly.

inlined:
Unwind the cleverness of CloudEvent<EventType, Extensions>. Types
now match the approved format of the Python API. CloudFunction now
templatizes the CloudEvent rather than the EventType so that we can
keep typing to include subtypes of events both in code complete and
in firebase-functions-test.

CC @TheIronDev since this affects firebase-functions-test

### DIFF
--- a/src/v2/core.ts
+++ b/src/v2/core.ts
@@ -57,7 +57,7 @@ export interface TriggerAnnotation {
  * A CloudEventBase is the base of a cross-platform format for encoding a serverless event.
  * More information can be found in https://github.com/cloudevents/spec
  */
-interface CloudEventBase<T> {
+export interface CloudEvent<T> {
   /** Version of the CloudEvents spec for this event. */
   readonly specversion: '1.0';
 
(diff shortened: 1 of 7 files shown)


src/v2/core.ts
```diff
@@ -78,32 +78,14 @@ interface CloudEventBase<T> {
...
-   * This map is only available on some event types that allow templates
-   * in the subject string, such as Firestore. When listening to a document
-   * template "/users/{uid}", an event with subject "/documents/users/1234"
-   * would have a params of {"uid": "1234"}.
-   *
-   * Params are generated inside the firebase-functions SDK and are not
-   * part of the CloudEvents spec nor the payload that a Cloud Function
-   * actually receives.
-   */
-  params?: Record<string, string>;
```
inlined, line 94: Note: params should probably only be present in Cloud Function types that actually support params.

TheIronDev:
THANK YOU!!!

TheIronDev (approved):

taeold (approved):
So clean.

PR #1095, opened 2022-05-04
Remove all errors when generating reference doc using api-extractor

taeold:
As is, generating reference doc for v2 sdk produces tons of errors. Most common errors are:

1) Exported type definition depending on unexported type definition 

2) Using `@hidden` tag which is unsupported in api-extractor

I solve both of these issues here.

I selected to use `@alpha` tag for params related types - that would prevent it from showing up in reference docs.


egilmorez (approved):
LGTM, thanks!

PR #1099, opened 2022-05-06
Add debug feature to always enable cors for emulated v2 https functions

taeold:
As suggested by @jhuleatt, we add a debug feature in the SDK to enable CORS setting (i.e. set CORS origin setting to allow requests from any origin) during function emulation to make emulator experience less painful. For example, given the following v2 https function:

```js
const f = onRequest({ cors: ['mysite.com']})...
```

Users would've had to add `localhost:XXX` to the cors setting to make function emulation work. With the new debug feature (which the Functions emulator will enable), users won't need to make such change.

Thanks @jhuleatt for pushing the idea for an excellent feature!


colerogers (approved):

src/v2/providers/https.ts
```diff
@@ -89,12 +90,13 @@ export function onRequest(
     opts = optsOrHandler as HttpsOptions;
   }
 
-  if ('cors' in opts) {
+  if (isDebugFeatureEnabled('enableCors') || 'cors' in opts) {
```
colerogers, line 93: should we add a simple test case to this?

jhuleatt: +1 to make sure we don't accidentally open someone's function up to the world

jhuleatt (approved):

inlined (approved):
I think we _also_ need to default cors to true to keep parity with V1.

inlined:
Whoops. I see where we did & still do default to CORS true for Callable functions in v2.

PR #1101, opened 2022-05-06
Update docs for root namespace

inlined:
Modify docs for the root namespace. Some basic changes:

* All fields are `@beta`
* Params is no longer exported in `index` because it is not ready for use (if you want, I can just mark everything as @alph instead, but IDK if we can avoid documenting the namespace at all.
* Constants are removed from options and instead are part of API comments
* Literal string types now have inlined their definitions so that values are part of the API docs.


colerogers (changes requested):

src/v2/core.ts
```diff
@@ -80,12 +82,28 @@ export interface CloudEvent<T> {
...
 export interface CloudFunction<EventType extends CloudEvent<unknown>> {
   (raw: CloudEvent<unknown>): any | Promise<any>;
 
+  /** @alpha */
   __trigger?: unknown;
+  /** @alpha */
   __endpoint: ManifestEndpoint;
 
+  /**
+   * The callback passed to the CloudFunction constructor.
```
egilmorez, line 100: Is CloudFunction a literal?

src/v2/core.ts
```diff
@@ -80,12 +82,28 @@ export interface CloudEvent<T> {
...
+   * Use run to test a CloudFunction
+   * @param event - The parsed event to handle.
+   * @returns Any return value. Google Cloud Functions will await any promise
+   *          before shutting down your Cloud Function. Resolved return values
```
egilmorez, line 104: Suggest "your function."

src/v2/options.ts
```diff
@@ -148,6 +105,11 @@ export interface GlobalOptions {
    * Timeout for the function in sections, possible values are 0 to 540.
    * HTTPS functions can specify a higher timeout.
    * A value of null restores the default of 60s
+   * The minimum timeout for a gen 2 function is 1s. The maximum timeout for a
+   * function depends on the type of function: Event handling functions have a
+   * maximum timeout of 540s (9 minutes). HTTPS and Callable functions have a
+   * maximum timeout of 36,00s (1 hour). TaskQueue functions have a maximum
```
egilmorez, line 111: Suggest "task queue functions" and "callable functions" unless these are literals.

src/v2/core.ts
```diff
@@ -80,12 +82,28 @@ export interface CloudEvent<T> {
   data: T;
 }
 
-/** A handler for CloudEvents. */
+/**
+ * A handler for CloudEvents.
```
egilmorez, line 86: Is "CloudEvents" a literal?

egilmorez (commented):
A few thoughts Thomas, thanks!

src/v2/core.ts
inlined: Kinda? It's the name of a class, but it's also the name of a specification (which is what I intended here). The capitalization is what is at https://cloudevents.io

colerogers (approved):

PR #1108, opened 2022-05-09
Update docs for auth

TheIronDev:
Update docs for auth

### DIFF
--- a/src/providers/auth.ts
+++ b/src/providers/auth.ts
@@ -54,7 +54,7 @@ export const provider = 'google.firebase.auth';
 /** @hidden */
 export const service = 'firebaseauth.googleapis.com';
 
-/** Resource level options */
+/** @public Resource level options */
 export interface UserOptions {
   blockingOptions?: {
     idToken?: boolean;
@@ -64,7 +64,7 @@ export interface UserOptions {
 }
 
 /**
- * Handle events related to Firebase authentication users.
+ * @public Handle events related to Firebase authentication users.
  */
 export function user(userOptions?: UserOptions) {
   return _userWithOptions({}, userOptions || {});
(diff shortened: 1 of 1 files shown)


colerogers (approved):

src/providers/auth.ts
```diff
@@ -54,7 +54,7 @@ export const provider = 'google.firebase.auth';
 /** @hidden */
 export const service = 'firebaseauth.googleapis.com';
 
-/** Resource level options */
+/** @public Resource level options */
```
inlined, line 57: visibility modifiers should be on their own line

src/providers/auth.ts
```diff
@@ -64,7 +64,7 @@ export interface UserOptions {
 }
 
 /**
- * Handle events related to Firebase authentication users.
+ * @public Handle events related to Firebase authentication users.
  */
 export function user(userOptions?: UserOptions) {
```
inlined, line 69: Add param and returns comments

src/providers/auth.ts
```diff
@@ -100,14 +100,14 @@ export class UserBuilder {
     private userOptions?: UserOptions
   ) {}
 
-  /** Respond to the creation of a Firebase Auth user. */
+  /** @public Respond to the creation of a Firebase Auth user. */
```
inlined, line 103: "Responds" since we pretend an implicit "this method..."

TheIronDev: Done

src/providers/auth.ts
TheIronDev: Done

src/providers/auth.ts
TheIronDev: Done

PR #1113, opened 2022-05-10
Move some types to interfaces; inline interfaces to make fields private

inlined:


TheIronDev (approved):
LGTM (format angry though)

PR #1116, opened 2022-05-10
Add package docs and copyright headers

inlined:
Unfortunately this doesn't actually fix the table of exported namespaces. I haven't figured out how to do that yet.

### DIFF
--- a/src/v2/params/index.ts
+++ b/src/v2/params/index.ts
@@ -1,4 +1,29 @@
-/** @hidden */
+// The MIT License (MIT)
+//
+// Copyright (c) 2017 Firebase
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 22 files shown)


TheIronDev (approved):

src/v2/params/index.ts
```diff
@@ -1,4 +1,29 @@
-/** @hidden */
+// The MIT License (MIT)
+//
+// Copyright (c) 2021 Firebase
```
colerogers, line 3: 2021?

src/v2/params/types.ts
```diff
@@ -1,3 +1,25 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2021 Firebase
```
colerogers, line 3: 2021?

colerogers (changes requested):

inlined:
Those files are written in 2021

colerogers (approved):

PR #1117, opened 2022-05-10
Updated PubSub docs, removing unused types + top-level comments

TheIronDev:
Updated PubSub docs, removing unused types + top-level comments

### DIFF
--- a/src/v2/providers/pubsub.ts
+++ b/src/v2/providers/pubsub.ts
@@ -14,7 +14,7 @@ import * as options from '../options';
  * PubSub Topic:
  * <ul>
  *   <li>A resource that you can publish messages to and then consume those messages via subscriptions.
- *   <li>An isolated data stream for Pub/Sub messages.
+ *   <li>An isolated data stream for pub/sub messages.
  *   <li>Messages are published to a topic.
  *   <li>Messages are listened to via a subscription.
  *   <li>Each subscription listens to the messages published to exactly one topic.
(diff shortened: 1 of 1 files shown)


src/v2/providers/pubsub.ts
```diff
@@ -4,20 +4,25 @@ import { CloudEvent, CloudFunction } from '../core';
 import * as options from '../options';
 
 /**
- * A PubSub Topic is:
+ * Google Cloud's Pub/Sub is a globally distributed message bus that automatically scales as you need it.
```
joehan, line 7: I think this should be just `Google Cloud Pub/Sub`

src/v2/providers/pubsub.ts
```diff
@@ -4,20 +4,25 @@ import { CloudEvent, CloudFunction } from '../core';
...
+ * You can create a function ({@link onMessagePublished}) that handles Pub/Sub events by using functions.pubsub.
+ *
+ * You can trigger a function whenever a new Pub/Sub message is sent to a specific topic.
```
joehan, line 10: Nit: 'This functions triggers whenever a new Pub/Sub...`

joehan (approved):
Some nits, but I see that this is copied from the linked docs, so feel free to ignore

src/v2/providers/pubsub.ts
inlined: +1

inlined (approved):
Great docs!

src/v2/providers/pubsub.ts
TheIronDev: Makes sense, updated!

PR #1119, opened 2022-05-10
Redefine fields in options so they get better docs

inlined:
Oof

### DIFF
--- a/src/common/providers/identity.ts
+++ b/src/common/providers/identity.ts
@@ -52,7 +52,8 @@ const CLAIMS_MAX_PAYLOAD_SIZE = 1000;
 
 /**
  * Shorthand auth blocking events from GCIP.
- * @internal
+ * @hidden
+ * @alpha
  */
 export type AuthBlockingEventType = 'beforeCreate' | 'beforeSignIn';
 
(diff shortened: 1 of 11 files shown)


src/common/providers/identity.ts
```diff
@@ -52,7 +52,8 @@ const CLAIMS_MAX_PAYLOAD_SIZE = 1000;
 
 /**
  * Shorthand auth blocking events from GCIP.
- * @internal
+ * @hidden
```
inlined, line 55: Note: Internal _should_ have been fine, but API Extractor was crashing so I changed it.

TheIronDev (approved):

src/v2/providers/tasks.ts
```diff
@@ -37,16 +37,113 @@ import { HttpsFunction } from './https';
...
+   */
+  ingressSettings?: options.IngressSetting | null;
+
+  /**
+   * User labels to set on the function.
+   */
+  labels?: Record<string, string>;
+
+  /*
+   * Secrets to bind to a functions.
```
TheIronDev, line 141: Nit: `function`

TheIronDev: + others

PR #1124, opened 2022-05-11
Temporarily remove docs for fireabse-functions-test

inlined:
Temporarily remove the reference to firebase-functions-test docs since the parser barfs on anything that doesn't have a `path`. @taeold will possibly have an idea what should actually happen here.


egilmorez (approved):
LG, thanks!

PR #1127, opened 2022-05-20
RTDB triggers for v2

colerogers:
WIP


src/common/providers/database.ts
```diff
@@ -0,0 +1,342 @@
...
+      // In this case we're dealing with an emulator
+      this.instance = app.options.databaseURL;
+    } else if (instance) {
+      // SDK always supplies instance, but user's unit tests may not
+      this.instance = instance;
+    } else if (app) {
+      this.instance = app.options.databaseURL;
+    } else if (process.env.GCLOUD_PROJECT) {
+      this.instance =
+        'https://' + process.env.GCLOUD_PROJECT + '.firebaseio.com';
```
fredzqm, line 60: When would we ever reach this? In tests only?

The default RTDB database should be `'https://' + process.env.GCLOUD_PROJECT + '-default-rtdb.firebaseio.com'` if not specified in firebase config.

colerogers: This is left over from v1 and we'd always supply the instance in the SDK. I'll change it to the correct default RTDB url.

fredzqm: Thanks!

src/v2/providers/database.ts
```diff
@@ -0,0 +1,452 @@
...
+  }
+}
+
+/** @internal */
+export function makeParams(
+  event: RawRTDBCloudEvent,
+  path: string,
+  instance: string
+): Record<string, string> {
+  const params: Record<string, string> = {};
```
inlined, line 323: WDYT about something like

```typescript
export function makeParams(
  event: RawRTDBCloudEvent,
  path: PathPattern,
  instance: PathPattern) {
 return {
   ...path.extractMatches(event.ref),
   ...instance.extractMatches(event.instance),
  }
}
 ```

File: src/v2/providers/database.ts [hunk 3]
```diff
@@ -0,0 +1,452 @@
...
+    let pathNdx = 0;
+    pathNdx < pathSegments.length && wildcardNdx < wildcards.length;
+    pathNdx++
+  ) {
+    const multiSegmentWildcard = pathSegments[pathNdx].includes('**');
+
+    const remainingPathSegments = pathSegments.length - 1 - pathNdx;
+    const nextRefElement = refSegments.length - remainingPathSegments;
+
+    if (pathSegments[pathNdx] === wildcards[wildcardNdx]) {
```
inlined, line 307: This logic breaks if there are two `**` components (you'll need recursion or maybe dynamic programming). Is that valid according to Eventarc? I strongly assume `**` was only allowed as the last component of a path in One Platform simply because allowing it in the middle allows expensive/ambiguous match calls (e.g. what does `/{a=**}/{b}/{c=**}` match against `/a/b/c/d/e/f`.

As it stands now, you might want to make a proper well-tested class for PathPattern with something like

```
interface PathSegment { ... }
class LiteralSegment extends PathSegment;
class SingleCaptureSegment extends PathSegment;
class GlobCaptureSegment extends PathSegment;
class PathPattern {
  private segments: PathSegment[];
  /** @throws on validation error */
  static compile(raw: string);
  matches(path: string): boolean;
  // If false, we don't need to use pathPattern as our eventarc match type.
  hasCaptures(): boolean;
  extractMatches(path: string): Record<string, string>;
}
```

src/v2/providers/database.ts
```diff
@@ -0,0 +1,452 @@
...
+}
+
+/** @hidden */
+function makeDatabaseEvent(
+  event: RawRTDBCloudEvent,
+  instance: string,
+  params: Record<string, string>
+): DatabaseEvent<DataSnapshot> {
+  const snapshot = new DataSnapshot(
+    event.data.delta,
```
inlined, line 345: It's delta in both create and delete events? I assumed it would be `event.data.delta` for beforeRefCreated and `event.data.data` for beforeRefDeleted.

inlined (changes requested):
Some overall comments:

1. We're missing `src/v2/database.js` which is used to quiet the import linter (since the maintainer of the repo says we should pay them if we want a fix.
2. We've got an outstanding PR to fix some behavior around null handling in the RTDB. Did we incorporate any of that here? Can we incorporate it without doing a major version bump? We should figure out how to not throw away the author's work
3. In general I think `_.isPlainObject` is more strongly verifying things than we need. The only thing we need to check is whether something has `typeof val === 'object'` and that the type is not `null`.
4. We can ship the current path pattern work for now if we need to, but I strongly suspect we'll want to create a PathPattern utilty class like we have in One Platform code. Maybe PathPattern.compile(string) which can throw if there's a malformed path. Then something like hasParams() and matchParams(atual: string): Record<string, string>.

spec/v2/providers/database.spec.ts
```diff
@@ -0,0 +1,606 @@
...
+      b: 'match_b',
+      c: 'match_c',
+    });
+  });
+
+  it('should parse multi segment with params after', () => {
+    const params = {};
+
+    database.matchParams(
+      'something/**/else/{a}/hello/{b}/world'.split('/'),
```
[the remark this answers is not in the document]
colerogers: Per the event arc docs (https://cloud.google.com/eventarc/docs/path-patterns), you can use a multisegment wildcard (**) with a single segment wildcard

src/v2/providers/database.ts
```diff
@@ -0,0 +1,452 @@
...
+export interface ReferenceOptions extends options.EventHandlerOptions {
+  /**
+   * Specify the handler to trigger on a database reference(s).
+   * This value can either be a single reference or a pattern.
+   * Examples~ '/foo/bar', '/foo/{bar} '
+   */
+  ref: string;
+  /**
+   * Specify the handler to trigger on a database instance(s).
+   * If present, this value can either be a single instance or a pattern.
```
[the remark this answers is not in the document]
colerogers: I'm proposing here that we will only support a single pattern for instance - "{a}"

src/v2/providers/database.ts
```diff
@@ -0,0 +1,452 @@
...
+    const remainingPathSegments = pathSegments.length - 1 - pathNdx;
+    const nextRefElement = refSegments.length - remainingPathSegments;
+
+    if (pathSegments[pathNdx] === wildcards[wildcardNdx]) {
+      params[trimParam(wildcards[wildcardNdx++])] = multiSegmentWildcard
+        ? refSegments.slice(refNdx, nextRefElement).join('/')
+        : refSegments[refNdx];
+    }
+
+    refNdx = multiSegmentWildcard ? nextRefElement : refNdx + 1;
```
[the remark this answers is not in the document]
colerogers: Per the eventarc docs (https://cloud.google.com/eventarc/docs/path-patterns), we can only have at most one mutli segment wildcard `**` and it does not have to be at the end

src/v2/providers/database.ts
colerogers: Per the eventarc docs (https://cloud.google.com/eventarc/docs/path-patterns), we can only have at most one mutli segment wildcard `**` and it does not have to be at the end.

Yeah having a separate class makes sense to me, I'll implement it. However, I don't think there should be any validation in the SDK since at runtime the function should always have a valid pattern string from eventarc. I think that validation logic should go into the CLI

src/v2/providers/database.ts
colerogers: Ahhh 🤦‍♂️ you're right

src/common/providers/database.ts
```diff
@@ -0,0 +1,344 @@
...
+    }
+    return this._ref;
+  }
+
+  /**
+   * The key (last part of the path) of the location of this `DataSnapshot`.
+   *
+   * The last token in a Database location is considered its key. For example,
+   * "ada" is the key for the `/users/ada/` node. Accessing the key on any
+   * `DataSnapshot` will return the key for the location that generated it.
```
egilmorez, line 93: "will return" could just be "returns" here and below.

Kudos from me if you check all "wills" and avoid them if possible :)

src/common/providers/database.ts
```diff
@@ -0,0 +1,344 @@
...
+
+  /**
+   * Extracts a JavaScript value from a `DataSnapshot`.
+   *
+   * Depending on the data in a `DataSnapshot`, the `val()` method may return a
+   * scalar type (string, number, or boolean), an array, or an object. It may also
+   * return `null`, indicating that the `DataSnapshot` is empty (contains no
+   * data).
+   *
+   * @return The DataSnapshot's contents as a JavaScript value (Object,
```
egilmorez, line 111: This should be a literal, or maybe just "The snapshot's contents?"   A snapshot can be a non-capitalized, generic noun IMO.

egilmorez (commented):
A couple nits for you Cole, thanks!

I'll take another look when only living code remains.

src/common/providers/database.ts
```diff
@@ -0,0 +1,344 @@
...
+        '-default-rtdb.firebaseio.com';
+    }
+
+    this._path = path;
+    this._data = data;
+  }
+
+  /**
+   * Returns a [`Reference`](/docs/reference/admin/node/admin.database.Reference)
+   * to the Database location where the triggering write occurred. Has
```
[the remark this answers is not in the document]
colerogers: Fixed

src/common/providers/database.ts
colerogers: Fixed

src/common/providers/database.ts
colerogers: Fixed

src/common/providers/database.ts
```diff
@@ -0,0 +1,344 @@
...
+  /**
+   * Exports the entire contents of the `DataSnapshot` as a JavaScript object.
+   *
+   * The `exportVal()` method is similar to `val()`, except priority information
+   * is included (if available), making it suitable for backing up your data.
+   *
+   * @return The contents of the `DataSnapshot` as a JavaScript value
+   *   (Object, Array, string, number, boolean, or `null`).
+   */
+  exportVal(): any {
```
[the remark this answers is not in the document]
colerogers: This looks like old documentation that we never stripped. I'm leaving in `exportVal` for legacy reasons, but I'll change the docs.

inlined (approved):

src/common/providers/database.ts
```diff
@@ -0,0 +1,344 @@
...
+
+  /**
+   * The key (last part of the path) of the location of this `DataSnapshot`.
+   *
+   * The last token in a Database location is considered its key. For example,
+   * "ada" is the key for the `/users/ada/` node. Accessing the key on any
+   * `DataSnapshot` will return the key for the location that generated it.
+   * However, accessing the key on the root URL of a Database will return `null`.
+   */
+  get key(): string {
```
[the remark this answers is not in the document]
colerogers: 👍

colerogers:
@rhodgkins Yep that's the one. This will apply both to v1 & v2 functions.

PR #1137, opened 2022-06-15
Update v2 rtdb to expose RawRTDBEvent and RawRTDBCloudEventData

TheIronDev:
This commit updates the v2 database provider to also expose the
RawRTDB Events.

This will be used in test-sdk when mocking returned DatabaseEvents.


colerogers (approved):
Should we throw `@hidden` on the interfaces to keep them out of doc?

TheIronDev:
Good catch! Updated!

PR #1140, opened 2022-06-15
Auth Blocking Triggers Bugbash

colerogers:
Now that the backend changes have been fully rolled out, change the SDK to handle v2 cloud run audience claims.

### DIFF
--- a/src/common/providers/identity.ts
+++ b/src/common/providers/identity.ts
@@ -835,9 +835,13 @@ export function wrapHandler(
         'skipTokenVerification'
       )
         ? unsafeDecodeAuthBlockingToken(req.body.data.jwt)
-        : handler.length === 2 
-          ? await apps().admin.auth()._verifyAuthBlockingToken(req.body.data.jwt)
-          : await apps().admin.auth()._verifyAuthBlockingToken(req.body.data.jwt, "run.app");
+        : handler.length === 2
+        ? await apps()
+            .admin.auth()
+            ._verifyAuthBlockingToken(req.body.data.jwt)
+        : await apps()
+            .admin.auth()
+            ._verifyAuthBlockingToken(req.body.data.jwt, 'run.app');
 
       const authUserRecord = parseAuthUserRecord(decodedPayload.user_record);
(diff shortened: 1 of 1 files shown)


src/common/providers/identity.ts
```diff
@@ -835,9 +835,13 @@ export function wrapHandler(
...
       )
         ? unsafeDecodeAuthBlockingToken(req.body.data.jwt)
+        : handler.length === 2
+        ? await apps()
+            .admin.auth()
+            ._verifyAuthBlockingToken(req.body.data.jwt)
         : await apps()
             .admin.auth()
-            ._verifyAuthBlockingToken(req.body.data.jwt);
+            ._verifyAuthBlockingToken(req.body.data.jwt, 'run.app');
```
taeold, line 844: q: I was checking to see when `_verifyAuthBlockingToken` function was introduced, and it seems like it's a pretty recent addition.

Because we support multiple version of the admin sdk (right now, all the way back to v8.0, though that will change in our next breaking release), I think we need to make sure that this code doesn't crash when using an older version.

e.g. we do something like that for app check here:

https://github.com/firebase/firebase-functions/blob/76e0afd442e141ad0016d01545e5883e27712c3f/src/common/providers/https.ts#L675-L678

taeold, line 844: The second argument to 

https://github.com/firebase/firebase-admin-node/blob/6b9760797c801e4f2e23c0e2fd4c595ae302f107/src/auth/token-verifier.ts#L414-L417  seems to be `isEmulator` and not `audience`. Did this change in some recent version of the admin sdk?

taeold: huh that's confusing. Do you know if the difference is between `getAuth` vs `admin.auth()`?

taeold: way ahead of me.

taeold (approved):

src/common/providers/identity.ts
colerogers: I believe they both return the `Auth` class, I'm pretty sure `token-verifier` is not directly exported

PR #1144, opened 2022-06-17
Rework apps, update admin, remove lodash

inlined:
This change involved a number  of small things that kinda got smushed together. Sorry.

1. I changed the "apps" library. It originally was supposed to hold a refcounted copy of apps logged in with user auth overrides. Ripping all that code out felt good! (including ripping out `before` and `after` in `makeCloudFunction`. Importantly, we will now share the default app if it already exists when we need it. This should help consolidate network connections/caches. OTOH, it could lead to subtle bugs. Push back if you think this optimization is too dangerous.
2. Since we updated the admin SDK version in `launch.next` already, I switched over to the v10 syntax so that we load fewer files on demand.
3. I cleaned up lodash in a few files and decided to pull on the thread until lodash is removed from the SDK entirely.
4. In order to efficiently implement a lodash method, I had to raise the typescript target to ES2019, which is covered by our new supported engines (14 & 16)


src/common/app.ts
```diff
@@ -0,0 +1,67 @@
...
+} from 'firebase-admin/app';
+import { firebaseConfig } from '../config';
+
+const APP_NAME = '__FIREBASE_FUNCTIONS_SDK__';
+
+let cache: App;
+export function getApp(): App {
+  // if (typeof cache === 'undefined') {
+  //   cache = getAppNamed(/* default */);
+  // }
```
colerogers, line 38: remove these comments

colerogers (changes requested):
Couple leftover comments

src/index.ts
```diff
@@ -55,6 +52,8 @@ export {
   logger,
 };
 
+export const app = { setEmulatedAdminApp };
```
taeold, line 55: sigh I wish I knew why this function is needed for the emulator to work.

(I have a theory that this is just a historical artifact before admin sdk natively supported the emulator)

src/common/app.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * This function allows the Firebase Emulator Suite to override the FirebaseApp instance
+ * used by the Firebase Functions SDK. Developers should never call this function for
+ * other purposes.
+ * N.B. For clarity for use in testing this name has no mention of emulation, but
+ * it must be exported from index as app.setEmulatedAdminApp or we break the emulator.
+ * We can remove this export when:
+ * A) We complete the new emulator and no longer depend on monkeypatching
```
taeold, line 58: actively researching this. I _think_ recent versions of the admin sdk doesn't rely on this monkey patching, and I'll try to confirm/deny this.

taeold (approved):

inlined:
@colerogers You left this with a change request. Does uncommenting those lines fix everything?

colerogers (approved):

PR #1153, opened 2022-06-21
Enhance RTDB v2 Triggers

colerogers:
- Export `Change` from `common/change.ts` since we're using it in both v1 & v2 functions
- Yank `reinterpretCast` from `Change` namespace
- Rename the triggers `onValue*`


src/v2/core.ts
```diff
@@ -112,3 +112,99 @@ export interface CloudFunction<EventType extends CloudEvent<unknown>> {
    */
   run(event: EventType): any | Promise<any>;
 }
+
+/**
+ * The Functions interface for events that change state, such as
+ * Realtime Database or Cloud Firestore `onWrite` and `onUpdate`.
+ *
+ * For more information about the format used to construct `Change` objects, see
+ * [`cloud-functions.ChangeJson`](/docs/reference/functions/cloud_functions_.changejson).
```
taeold, line 121: This seems like a wrong link yes?

taeold (commented):
Q: This is verbatim copy of the existing code right? Do we worry about having to worry about divergence of the code or is that actually desirable in this case?

src/v2/core.ts
colerogers: Updated the TSDOC to just link to `ChangeJson`

colerogers:
@taeold Good point, I know we duplicated a bunch of the Event Handler options for better docs so I thought we should also do that here. Ended up moving it the common code path

spec/common/core.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      });
+      expect(created instanceof Change).to.equal(true);
+      expect(created.before).to.deep.equal({ foo: 'bar' });
+      expect(created.after).to.deep.equal({ foo: 'faz' });
+    });
+
+    it('should apply the customizer function to `before` and `after`', () => {
+      function customizer<T>(input: any) {
+        input.another = 'value';
+        // _.set(input, 'another', 'value');
```
inlined, line 101: please remove

src/common/core.ts
```diff
@@ -0,0 +1,112 @@
...
+   */
+  before?: any;
+  /**
+   * @hidden
+   * Comma-separated string that represents names of fields that changed.
+   */
+  fieldMask?: string;
+}
+
+export namespace Change {
```
inlined, line 55: The industry has moved away from using namespaces in TypeScript. Let's just make `fromJSON` and `fromObject` static members and `applyFieldMask` can be a top-level method.

src/common/core.ts
```diff
@@ -0,0 +1,112 @@
+// The MIT License (MIT)
```
inlined, line 1: Let's make this `src/common/change.ts` (or `src/common/utilities/change.ts` after merging the v1/v2 partition PR)

src/providers/database.ts
```diff
@@ -22,19 +22,19 @@
...
 } from '../cloud-functions';
+import { Change } from '../common/core';
 import { DataSnapshot } from '../common/providers/database';
 import { firebaseConfig } from '../config';
 import { DeploymentOptions } from '../function-configuration';
 import { normalizePath } from '../utilities/path';
 import { applyChange } from '../utils';
 
-export { DataSnapshot };
+export { DataSnapshot, Change };
```
inlined, line 37: I don't think we actually exported Change from the RDTB namespace in the past, did we? IIRC it was just exported in `index.ts`. We should put it in `v1/index.ts` and `v2/index.ts` as well as `v2/core.ts`

src/common/core.ts
colerogers: changed to `src/common/change.ts`

src/common/core.ts
colerogers: SG, updated the class

src/common/change.ts
```diff
@@ -0,0 +1,109 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/**
+ * `ChangeJson` is the JSON format used to construct a Change object.
+ */
+export interface ChangeJson {
```
inlined, line 26: Should this be @internal? Are users actually constructing a ChangeJson?

inlined (approved):

inlined:
Trying to make sure we get this in, sync `launch.next` and then I'll start doing some surgery. Committing rather than blocking on a comment about whether a field should be made internal. Let's talk tomorrow.

PR #1155, opened 2022-06-28
Make event params strongly typed

inlined:
Implements go/cf3-typed-params.

WIP: currently only common and v2 are implemented.

Had to update `prettier` to the next major version so it could read TS 4.1 features without erroring out. I created  #1160 as a base for this change to make it easier to separate mechanical from hand-crafted changes. If this is approved, however, I'll commit both changes to `launch.next`


spec/v2/providers/database.spec.ts
```diff
@@ -437,12 +444,18 @@ describe('database', () => {
       const func = database.onValueCreated(
         {
           ref: '/foo/{path=**}/{bar}/',
-          instance: 'my-instance',
+          instance: '{instance}',
```
colerogers, line 447: Capture syntax cannot be used for `instance`

spec/v2/providers/database.spec.ts
```diff
@@ -453,11 +466,10 @@ describe('database', () => {
         labels: {},
         eventTrigger: {
           eventType: database.createdEventType,
-          eventFilters: {
-            instance: 'my-instance',
-          },
+          eventFilters: {},
           eventFilterPathPatterns: {
             ref: 'foo/{path=**}/{bar}',
+            instance: '{instance}',
```
colerogers, line 472: same here

taeold (approved):
Coolest PR of the year

PR #1168, opened 2022-07-14
WIP: present parameters in place of literal values for parameterized fields

Berlioz:
So what I'm pretty sure is going on is that users can set Functions options in two ways: either by presenting a GlobalOptions object to setGlobalOptions, or by presenting an extension type of GlobalOptions as the parameter to an individual function declaration. The fields from GlobalOptions are just copied into ManifestEndpoint, which is the type the CLI sees, as part of the discovery process.

Therefore, changing the type signatures of GlobalOptions(/its various extensions) and ManifestEndpoint _should_ be enough to allow users to provide parameterized values of their configuration fields.

So far, this PR does _not_ include:
- the trigger type specific parameterized fields
- full adjustments to the comments for the changed fields
- any tests whatsoever
- sanity checking on whether provided strings are actually CEL (but shouldn't we just validate that in the CLI?)
but I wanted to see whether I was totally barking up the wrong tree first.

### DIFF
--- a/src/runtime/manifest.ts
+++ b/src/runtime/manifest.ts
@@ -27,12 +27,12 @@ export interface ManifestEndpoint {
   entryPoint?: string;
   region?: string[];
   platform?: string;
-  availableMemoryMb?: number;
-  maxInstances?: number;
-  minInstances?: number;
-  concurrency?: number;
+  availableMemoryMb?: number | string;
+  maxInstances?: number | string;
+  minInstances?: number | string;
+  concurrency?: number | string;
   serviceAccountEmail?: string;
-  timeoutSeconds?: number;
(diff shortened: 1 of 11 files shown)


inlined (changes requested):
Please do not allow generic strings. In this codebase the customer actually entered params or expressions. That means we can add things like base types for Expression<T> that are real _classes_, not just a silly annotation for `T | string | null`. 

For now, let's create an Expression<T> type, make all params inherit from Expression<> and take T | Expression<T> | null on all these fields (I'm OK making it a Field<T> shorthand if you want, but you'll still need things like Expression<T> for arrays).

Berlioz:
This PR is enough up the wrong tree that I'm closing it and sending #1171 for review instead.

PR #1169, opened 2022-07-14
Update typedoc from 0.21.2 to ^0.23.7.

kaibolay:
This should fix:

```
$ npm run build:pack

> firebase-functions@3.22.0 build:pack
> rm -rf lib && npm install && tsc -p tsconfig.release.json && npm pack

npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: typedoc@0.21.2
npm ERR! Found: typescript@4.6.3
npm ERR! node_modules/typescript
npm ERR!   dev typescript@"^4.3.5" from the root project
npm ERR!   typescript@"~4.6.3" from @microsoft/api-extractor@7.23.0
npm ERR!   node_modules/@microsoft/api-extractor
npm ERR!     dev @microsoft/api-extractor@"^7.18.7" from the root project
npm ERR!   4 more (ts-node, tslint, tsutils, tsutils)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer typescript@"4.0.x || 4.1.x || 4.2.x || 4.3.x" from typedoc@0.21.2
npm ERR! node_modules/typedoc
npm ERR!   dev typedoc@"0.21.2" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: typescript@4.3.5
npm ERR! node_modules/typescript
npm ERR!   peer typescript@"4.0.x || 4.1.x || 4.2.x || 4.3.x" from typedoc@0.21.2
npm ERR!   node_modules/typedoc
npm ERR!     dev typedoc@"0.21.2" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

PR #1170, opened 2022-07-15
Fix issues with outdated package-lock.json file.

taeold:
Our `package-lock.json` file is outdated and our dependencies are a bit jumbled up.

I didn't notice any issue when using `nodejs14` but when using `nodejs16` reveals several issues that I've tried to address in the PR:

1. Update `typedoc` dependency to avoid impossible dep resolution (first reported https://github.com/firebase/firebase-functions/pull/1169)

2. Regenerate `package-lock.json`. Looks like it's "behind" package.json. I took this opportunity to use `lockfileVersion: 2` which is the default if you use `nodejs16`

3. Using `firebase-admin` v10.3.0 broke some unit test we had. Fixed in 2ee9549120ce0ec3a069c657d47c08c68ee1b53f.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,5779 @@
 {
   "name": "firebase-functions",
   "version": "3.22.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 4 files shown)


kaibolay (approved):
Thanks!

PR #1171, opened 2022-07-19
Insert Expression and Field types into GlobalOptions and inheriting types

Berlioz:
By changing the types we allow inside the GlobalOptions parameter of the various CF3 setup functions, we automatically plumb them through to the ManifestSpec. Derived from #1168, which was busted enough I decided to just make a new branch.

So far, this PR still does not include:

~~the trigger type specific parameterized fields~~
full adjustments to the comments for the changed fields
~~any tests whatsoever~~ but it seems like we're going to rely more on integration tests here
sanity checking on whether provided strings are actually CEL (but shouldn't we just validate that in the CLI?)
but I wanted to see whether I was totally barking up the wrong tree first.


src/v2/expressions.ts
```diff
@@ -0,0 +1,24 @@
...
+  return arg as unknown as string;
+}
+
+/**
+ * Sanity check on whether a CEL expression is actually evaluatable
+ * @internal
+ */
+export function IsValid<Expression>(arg: Expression): boolean {
+  return true;
+}
```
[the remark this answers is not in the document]
Berlioz: Ack.

spec/v2/providers/tasks.spec.ts
```diff
@@ -112,6 +113,64 @@ describe('onTaskDispatched', () => {
...
+        rateLimits: {
+          maxConcurrentDispatches: FromParam<number>('MAXCONCURRENTDISPATCHES'),
+          maxDispatchesPerSecond: FromParam<number>('MAXDISPATCHESPERSECOND'),
+        },
+        invoker: 'private',
+      },
+      () => {}
+    );
+
+    expect(result.__trigger).to.deep.equal({
```
taeold, line 136: I wouldn't worry about __trigger annotation for v2.

src/v2/expressions.ts
```diff
@@ -0,0 +1,54 @@
+/*
```
taeold, line 1: Why is param expression specific to v2?

src/v2/expressions.ts
```diff
@@ -0,0 +1,54 @@
...
+  return `{{ params.${paramName} }}`;
+}
+
+/**
+ * Creates an Expression that will cause an Options field to take on a runtime-defined
+ * value obtained by evaluating the provided CEL expression.
+ * The CLI does not currently have the capability to evaluate arbitrary CEL expressions.
+ * You almost certainly want to use FromParam() instead.
+ */
+export function FromCEL<T extends string | number | boolean>(
```
taeold, line 35: q: why is this fn necessary?

src/v2/options.ts
```diff
@@ -217,8 +224,22 @@ export function getGlobalOptions(): GlobalOptions {
  * Additional fields that can be set on any event-handling Cloud Function.
  */
 export interface EventHandlerOptions extends GlobalOptions {
+  eventType?: string;
+
+  eventFilters?: Record<string, string | Expression<string>>;
+  eventFilterPathPatterns?: Record<string, string | Expression<string>>;
```
taeold, line 230: what's going on here? We don't expect users to set these options no?

src/v2/options.ts
```diff
@@ -217,8 +224,22 @@ export function getGlobalOptions(): GlobalOptions {
...
+  retry?: Field<boolean>;
+
+  /** Region of the EventArc trigger. */
+  region?: Field<string>;
+
+  /** The service account that EventArc should use to invoke this function. Requires the P4SA to have ActAs permission on this service account. */
+  serviceAccount?: string | null;
+
+  /** The name of the channel where the function receives events. */
+  channel?: string;
```
taeold, line 242: I'm also not sure where these things are coming from and how it relates to the changes we are making. Especially `region` and `serviceAccount` - they are not the same thing as the ones we have defined in `GlobalOptions` and doesn't seem correct to me.

src/v2/providers/tasks.ts
```diff
@@ -33,21 +33,68 @@ import {
...
 } from '../../common/providers/tasks';
 import * as options from '../options';
 import { HttpsFunction } from './https';
+import { Expression, Field } from '../expressions';
 
-export { AuthData, RateLimits, Request, RetryConfig };
+export { AuthData, Request };
+
+/** How congestion control should be applied to the function. */
+export interface ParameterizedRateLimits {
```
taeold, line 45: Why do we do this instead of wrapping RetryConfig/RateLimits in `../../common/providers/tasks` instead?

Is params only supported in v2?

taeold (changes requested):

src/v2/expressions.ts
```diff
@@ -0,0 +1,24 @@
...
+export function ExprString<Expression>(arg: Expression): string {
+  return arg as unknown as string;
+}
+
+/**
+ * Sanity check on whether a CEL expression is actually evaluatable
+ * @internal
+ */
+export function IsValid<Expression>(arg: Expression): boolean {
+  return true;
```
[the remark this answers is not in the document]
Berlioz: Yeah, was a stub in that commit.

spec/v2/providers/tasks.spec.ts
colerogers, line 136: you can yank this, we don't use `__trigger` anymore

src/v2/core.ts
```diff
@@ -27,16 +27,17 @@
...
 export { Change };
 
 /** @internal */
 export interface TriggerAnnotation {
   platform?: string;
-  concurrency?: number;
-  minInstances?: number;
-  maxInstances?: number;
-  availableMemoryMb?: number;
+  concurrency?: number | Expression<number>;
```
colerogers, line 37: We don't use `TriggerAnnotation` anymore, so you can revert these changes

colerogers (changes requested):

src/v2/options.ts
Berlioz: ...I wasn't aware of that, actually, but having actually looked at the relevant code paths, i think you're right.

src/v2/expressions.ts
Berlioz: The code that currently allows users to define params in their functions code lives in the v2 namespace, so...I guess I just assumed?

spec/runtime/loader.spec.ts
```diff
@@ -313,7 +313,14 @@ describe('loadStack', () => {
       {
         name: 'has params',
         modulePath: './spec/fixtures/sources/commonjs-params',
-        expected: { ...expected, params: [{ name: 'FOO', type: 'string' }] },
+        expected: { ...expected, params: [
+          { name: 'FOO', type: 'string' },
+          { name: 'BAR', type: 'string', default: '{{ params.FOO }}', label: 'asdf'},
+          { name: 'BAZ', type: 'string', input: {type: 'select', select: [{value: 'a'}, {value: 'b'}]}},
```
inlined: you, me, and @mbleigh really need to get on the same page for {type: 'select'} vs selectInput: { }.
We cannot release params until we are consistent here.

spec/v2/params.spec.ts
```diff
@@ -239,11 +239,13 @@ describe('params', () => {
 
   describe('Param', () => {
     describe('#toSpec()', () => {
+      /*
```
inlined, line 242: remove the comments or remove the code.

src/v2/params/types.ts
```diff
@@ -21,14 +21,31 @@
 // SOFTWARE.
 
 /** @hidden */
-type ParamValueType = 'string' | 'list' | 'boolean' | 'int' | 'float' | 'json';
+type ParamValueType = 'string' | 'list' | 'boolean' | 'int' | 'float' | 'json' | 'secret';
```
inlined: cut json for now

src/v2/params/types.ts
```diff
@@ -54,19 +71,41 @@ export class Param<T = unknown> {
...
-        | { toString?: () => string }
-        | undefined)?.toString?.();
-    }
 
-    return out as ParamSpec<string>;
+    return out as ParamSpec<T>;
+  }
+}
+
+export class SecretParam extends Param<string> {
```
inlined, line 121: I personally wouldn't make SeecretParam inherit from Param<string> because Param<string> is an Expression<string> and SecretParam is not.

src/v2/expressions.ts
```diff
@@ -0,0 +1,54 @@
+/*
+ * A CEL expression which can be evaulated during function deployment, and
+ * resolved to a value of the generic type parameter: i.e, you can pass
+ * an Expression<number> as the value of an option that normally accepts numbers.
+ */
+export type Expression<T extends string | number | boolean> = string; // eslint-disable-line
```
inlined, line 6: In the Functions codebase an Expression<T> is not the same thing as T. Expression<T> is a base class of Param<T> but can (eventually) do a few extra things:

1. BooleanExpressions should subclass Expression<T> and adds the ternary operator to create an Expression<V>
2. All (non-boolean?) expressions should support an equality condition to make a BooleanExpression

src/v2/expressions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type Field<T extends string | number | boolean> =
+  | T
+  | Expression<T>
+  | null;
```
inlined, line 11: We've talked about not using `null` and instead invent `options.ResetToDefault` which is defined to have a `toJSON` which returns `null`

src/v2/expressions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Creates an Expression that will cause an Options field to take on a runtime-defined
+ * value equal to a CF3 Param defined by the functions in v2/params.ts.
+ */
+export function FromParam<T extends string | number | boolean>(
```
inlined, line 17: People should not need this.

src/v2/expressions.ts
inlined: +1. Remove or `@internal` this too.

src/v2/options.ts
```diff
@@ -76,6 +77,12 @@ const MemoryOptionToMB: Record<MemoryOption, number> = {
   '32GiB': 32768,
 };
 
+function isMemoryOption(
+  arg: MemoryOption | Expression<number>
+): arg is MemoryOption {
+  return arg in MemoryOptionToMB;
```
inlined, line 83: Isn't this going to fail for an Expression<number>?

src/v2/options.ts
inlined: I think we may actually need to support expressions here. For example, a bucket or RTDB collection could be an expression.

src/v2/options.ts
```diff
@@ -246,8 +267,8 @@ export function optionsToTriggerAnnotations(
     opts,
     'availableMemoryMb',
     'memory',
-    (mem: MemoryOption) => {
-      return MemoryOptionToMB[mem];
+    (mem: MemoryOption | Expression<number>): number | string => {
+      return isMemoryOption(mem) ? MemoryOptionToMB[mem] : ExprString(mem);
```
inlined, line 271: Make `Expression` implement `toString` and use that. We need `toString` in order to make string interpolation work anyway.

src/v2/providers/tasks.ts
inlined: I'm in favor of letting v2 functions be configured with Params as well.

src/v2/providers/tasks.ts
```diff
@@ -33,21 +33,68 @@ import {
...
+   * The minimum time to wait between attempts. If left unspecified
+   * will default to 100ms.
+   */
+  minBackoffSeconds?: Field<number>;
+}
 
 export interface TaskQueueOptions extends options.EventHandlerOptions {
   /** How a task should be retried in the event of a non-2xx return. */
-  retryConfig?: RetryConfig;
+  retryConfig?: ParameterizedRetryConfig;
```
inlined, line 94: +1 to Daniel's comments. Just modify the normal `RetryConfig` and `RateLimits` to support Expressions.

inlined (commented):
You're going to hate me for this but:

1. Please make an actual Expression _class_ that has things like `.val()` and `.toString()` and `.eq(test): BooleanExpression` and a `BooleanExpression` class that implements `then<T>(trueVal: T, falseVal: T): Expression<T>`
2. In customer-facing SDKs I've been stricter about using typedefs for public types since they can worsen code complete. This is, for example, why I ripped out all typedefs for the handler function signatures in all of our handler code; it simply wasn't helping customers. I think `Field<T>` needs to be `T | params.Expression<T> | ResetValue`

src/v2/expressions.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  toCEL() {
+    return `{{ ${this.toString()} }}`;
+  }
+}
+
+export class IfElseExpression<
+  T extends string | number | boolean | string[]
+> extends Expression<T> {
+  test: Expression<boolean>;
```
TheIronDev, line 25: Nit: this can be cleaned up a good bit with go/tsstyle#parameter-properties

src/v2/expressions.ts
```diff
@@ -0,0 +1,82 @@
...
+export abstract class Expression<
+  T extends string | number | boolean | string[]
+> {
+  rawValue: string;
+
+  get val(): any {
+    return this.rawValue;
+  }
+
+  toCEL() {
```
TheIronDev, line 17: Nit: requesting docs on this function

src/v2/expressions.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.ifTrue = ifTrue;
+    this.ifFalse = ifFalse;
+  }
+
+  toString() {
+    return `${this.test} ? ${this.ifTrue} : ${this.ifFalse}`;
+  }
+}
+
+export class CompareExpression<
```
TheIronDev, line 41: Nit: requesting docs on the functions in this class

src/v2/expressions.ts
```diff
@@ -0,0 +1,82 @@
+import { Param } from './params/types';
+
+/*
+ * A CEL expression which can be evaulated during function deployment, and
```
TheIronDev, line 4: typo nit: evaluated

src/v2/expressions.ts
TheIronDev: + other classes

src/v2/expressions.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  toString() {
+    return `${this.test} ? ${this.ifTrue} : ${this.ifFalse}`;
+  }
+}
+
+export class CompareExpression<
+  T extends string | number | boolean | string[]
+> extends Expression<boolean> {
+  cmp: '==' | '>' | '>=' | '<' | '<=';
```
TheIronDev, line 44: Maybe add a type to encapsulate this?

src/v2/providers/database.ts
```diff
@@ -102,7 +103,7 @@ export interface ReferenceOptions extends options.EventHandlerOptions {
    * Amount of memory to allocate to a function.
    * A value of null restores the defaults of 256MB.
    */
-  memory?: options.MemoryOption | null;
+  memory?: options.MemoryOption | Expression<number> | null;
```
taeold, line 106: huh I forgot why we repeat this option on each and every v2 provider function 🤔

taeold (commented):
Just a few comments.

spec/fixtures/sources/commonjs-params/index.js
```diff
@@ -1,18 +1,26 @@
 const functions = require("../../../../src/index");
 const functionsv2 = require("../../../../src/v2/index");
-const { defineString } = require("../../../../src/v2/params");
+const params = require("../../../../src/v2/params");
 
-defineString("FOO");
+params.defineString("BORING");
+params.defineString("FOO", {textInput: {validationRegex:"\w+"}})
+params.defineString("BAR", {default: "{{ params.FOO }}", label: "asdf"});
```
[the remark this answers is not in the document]
inlined: We do explicitly support defining a param with a default that is another _expression_ (can be a param or can be an operation on a param).

spec/fixtures/sources/commonjs-params/index.js
```diff
@@ -1,18 +1,26 @@
...
+exports.v1http = functions.https.onRequest({}, (req, resp) => {
     resp.status(200).send("PASS");
 });
 
 exports.v1callable = functions.https.onCall(() => {
     return "PASS";
 });
 
-exports.v2http = functionsv2.https.onRequest((req, resp) => {
+exports.v2http = functionsv2.https.onRequest({}, (req, resp) => {
```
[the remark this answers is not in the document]
inlined: +1. If this is optional, it's weird and should be removed. If it's mandatory, we should fix a bug somewhere.

spec/runtime/loader.spec.ts
```diff
@@ -320,7 +322,7 @@ describe('loadStack', () => {
             {
               name: 'FOO',
               type: 'string',
-              input: { type: 'text', validationRegex: 'w+' },
+              textInput: { validationRegex: 'w+' },
```
inlined, line 325: also... and please don't kill me here, it didn't go from `{input: { type: 'text', other: field }}` to `{textInput: {other: field}}` it went to `{input: { text: {other: field }}}`. In proto, input will be a message that has a oneof for the input types.

spec/runtime/manifest.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+      specVersion: 'v1alpha1',
+    };
+    const expected = {
+      endpoints: {
+        v2http: {
+          platform: 'gcfv2',
+          entryPoint: 'v2http',
+          labels: {},
+          httpsTrigger: {},
+          concurrency: '{{ asdf }}',
```
inlined, line 34: shouldn't this and below be `params.asdf`?

src/function-configuration.ts
```diff
@@ -72,11 +74,11 @@ export const INGRESS_SETTINGS_OPTIONS = [
  * Scheduler retry options. Applies only to scheduled functions.
  */
 export interface ScheduleRetryConfig {
-  retryCount?: number;
-  maxRetryDuration?: string;
-  minBackoffDuration?: string;
-  maxBackoffDuration?: string;
-  maxDoublings?: number;
+  retryCount?: number | Expression<number> | null;
```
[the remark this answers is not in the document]
inlined: No. I want to err on the side of readability. If a user looks up `retryCount` and sees `number | Expression<number> | null` they probably have a better understanding of the world from this API than if they saw `Field<number>` because in the `Field` case, they _must_ look up another page of documentation.

src/runtime/manifest.ts
```diff
@@ -20,24 +20,25 @@
...
-export interface ManifestEndpoint {
+export type ManifestEndpoint = {
   entryPoint?: string;
   region?: string[];
   platform?: string;
-  availableMemoryMb?: number;
-  maxInstances?: number;
-  minInstances?: number;
-  concurrency?: number;
+  availableMemoryMb?: number | Expression<number>;
```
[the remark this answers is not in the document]
inlined: We can have a pedantic pass for resettable fields once that API proposal passes.

src/runtime/manifest.ts
```diff
@@ -93,3 +94,24 @@ export interface ManifestStack {
...
+ * Returns the JSON representation of a ManifestStack, which has CEL
+ * expressions in its options as object types, with its expressions
+ * transformed into the actual CEL strings.
+ * @internal
+ */
+export function stackToWire(stack: ManifestStack): Object {
+  let wireStack = stack as any;
+  let traverse = function traverse(obj: Object) {
+    for (const [key, val] of Object.entries(obj)) {
+      if (val instanceof Expression) {
```
inlined, line 108: if you rename `toCEL` as `toJSON` then `JSON.stringify` will call it for you automatically.

src/v2/expressions.ts
inlined: go/cf3-params-detailed is the doc that formally proposes ditching v2, but it seems easier for Victor to do that when he reaches a stopping point than having some PRs in v2 and some in root. Besides, technically that should be done in the launch.next branch, right? Technically we're changing an export path.

src/v2/expressions.ts
inlined: ¯\_(ツ)_/¯ I'm less concerned with that when the only consumer of this API is in the same file

src/v2/expressions.ts
```diff
@@ -1,54 +1,104 @@
...
 /**
- * Casts an Expression to its literal string value, for use by __getSpec()
- * @internal
+ * Creates an Expression that will cause an Options field to take on a runtime-defined
+ * value based on the value of a CF3 Param defined via the helpers in the v2/params
+ * namespace.
  */
-export function ExprString<Expression>(arg: Expression): string {
-  return (arg as unknown) as string;
+export class ParamExpression<
```
inlined, line 87: I originally tried this, but I found returning an Expression<T> to be a dead end. For me, at least, I had to create new "Base" types for each expression so I could add the valid operators to them. So for example, a StringParam is a StringExpression, which implements `eq` in addition to being an Expression<string>.  The only time I gave up and returned Expression<T> is the `then` API, but really I should go back and modify that to have N overrides so that each return type is a FooExpression and not an Expression<Foo>.

src/v2/params/index.ts
```diff
@@ -55,6 +54,16 @@ function registerParam(param: Param) {
   declaredParams.push(param);
 }
 
+/**
+ * For testing.
+ * @internal
+ */
+export function clearParams() {
+  while (declaredParams.length > 0) {
```
inlined, line 62: I appreciate that you're trying to avoid letting constness escape the package, but fortunately JavaScript has another horrifying const-breaking method that lets you clear an array in O(1) instead of O(N):

```
declaredParams.splice(0, declaredParams.length);
```

src/v2/params/types.ts
```diff
@@ -69,31 +72,36 @@ export interface SelectOptions<T = unknown> {
...
   constructor(readonly name: string, readonly options: ParamOptions<T> = {}) {}
 
-  get rawValue(): string | undefined {
-    return process.env[this.name];
+  get value(): T {
+    throw new Error('Not implemented');
+  }
+
+  expr() {
+    return new ParamExpression<T>(this);
```
inlined, line 95: Can you explain what punchline you're setting up here with ParamExpression<T>?

I originally intended expr to just be a string representation, so I want to make sure I'm following what's going on here.

src/v2/params/types.ts
```diff
@@ -69,31 +72,36 @@ export interface SelectOptions<T = unknown> {
...
   description?: string;
   type: ParamValueType;
-  input?: ParamInput<T>;
-}
+} & HasAtMostOneInput<T>;
 
 export type ParamOptions<T = unknown> = Omit<ParamSpec<T>, 'name' | 'type'>;
 
-export class Param<T = unknown> {
+export abstract class Param<T extends string | number | boolean | string[]> {
```
inlined, line 85: I want to play with abstract classes a bit to see if it changes my mind about the class hierarchy. I guess I don't mind having all expression types supporting the `eq` method even though I loose my sanity a smidge when I see "if (bool === true)". But Boolean expressions need to be their own type I think because only boolean expressions can implement `then`.  So maybe.... create a boolean expression type that implements `then` and make sure `eq` returns it and that `eq` is on BooleanParam?

src/v2/params/types.ts
```diff
@@ -69,31 +72,36 @@ export interface SelectOptions<T = unknown> {
...
-  get value(): any {
-    return this.rawValue || this.options.default || '';
+  cmp(cmp: '==' | '>' | '>=' | '<' | '<=', rhs: T) {
+    return new CompareExpression<T>(cmp, this.expr(), rhs);
   }
 
+  equals(rhs: T) {
+    return this.cmp('==', rhs);
+  }
   toString() {
```
inlined, line 105: As mentioned before, this versoin of `toString` may not work because you lose easy access to the part of the expression that doesn't have the {{}} escape sequence. That's what `expr` was for. Every Expression subclass implements its own `expr` (operators do so recursively) and then `toString` and `toJSON` are simply \`{{ ${ this.expr() } }}\`

spec/fixtures/sources/commonjs-params/index.js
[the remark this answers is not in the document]
Berlioz: Git oopsie. Corrected.

src/function-configuration.ts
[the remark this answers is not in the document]
taeold: Ah that's a good point.

Berlioz:
That latest commit made a fairly significant change to the underlying types of params/expressions to fit the more recent design doc. I'm going to resolve a ton of comments which basically reduce to "change the types/type signatures".

src/v2/params/types.ts
Berlioz: ack (cf3-params-detailed)

src/v2/params/types.ts
Berlioz: Nothing good, that's what. Got rid of ParamExpression

src/v2/params/index.ts
Berlioz: Huh. Done.

src/v2/expressions.ts
```diff
@@ -3,48 +3,90 @@
...
-export type Expression<T extends string | number | boolean> = string; // eslint-disable-line
+export abstract class Expression<T extends string | number | boolean> {
+  rawValue: string;
 
-export type Field<T extends string | number | boolean> =
-  | T
-  | Expression<T>
-  | null;
+  constructor(celLiteral: string) {
+    this.rawValue = celLiteral;
```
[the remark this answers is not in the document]
Berlioz: Yeah, good point. Done.

src/v2/expressions.ts
```diff
@@ -3,48 +3,90 @@
...
+  }
+
+  get val(): any {
+    return this.rawValue;
```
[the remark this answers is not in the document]
Berlioz: The rawValue thing is also not workable.

src/v2/expressions.ts
```diff
@@ -3,48 +3,90 @@
...
-  return celString;
+export class BooleanExpression extends Expression<boolean> {
+  lhs: ParamRef;
+  rhs: ParamRef | string | number | boolean;
+  cmp: '==' | '<' | '>';
+
+  constructor(lhs:ParamRef, rhs: ParamRef | string | number | boolean, cmp?: '==' | '<' | '>') {
+    const rhsVal = (typeof rhs === "object") ? `params.${rhs.name}` : rhs
+    super(`{{ params.${lhs.name} ${cmp} ${rhsVal} }}`)
+  }
```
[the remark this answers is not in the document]
Berlioz: Outdated.

src/v2/expressions.ts
```diff
@@ -3,48 +3,90 @@
...
+  toString() {
+    return this.rawValue;
+  }
+
+  toJSON() {
+    return this.toString();
+  }
+}
+
+export class ParamRef {
```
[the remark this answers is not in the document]
Berlioz: Outdated.

spec/runtime/loader.spec.ts
```diff
@@ -313,7 +315,42 @@ describe('loadStack', () => {
...
         modulePath: './spec/fixtures/sources/commonjs-params',
-        expected: { ...expected, params: [{ name: 'FOO', type: 'string' }] },
+        expected: {
+          ...expected,
+          params: [
+            { name: 'BORING', type: 'string' },
+            {
+              name: 'FOO',
+              type: 'string',
+              textInput: { validationRegex: 'w+' },
```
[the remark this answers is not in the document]
Berlioz: Oops. Fixed.

src/v2/expressions.ts
Berlioz: Fixed.

src/v2/expressions.ts
Berlioz: Done.

src/v2/expressions.ts
Berlioz: Done.

src/v2/options.ts
Berlioz: Outdated.

src/bin/firebase-functions.ts
```diff
@@ -63,7 +64,7 @@ if (process.env.FUNCTIONS_CONTROL_API === 'true') {
     try {
       const stack = await loadStack(functionsDir);
       res.setHeader('content-type', 'text/yaml');
-      res.send(JSON.stringify(stack));
+      res.send(JSON.stringify(stackToWire(stack)));
```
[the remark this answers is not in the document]
Berlioz: ....you know. Now that you mention it. I can't unsee it now.

src/runtime/manifest.ts
```diff
@@ -20,24 +20,25 @@
...
 // SOFTWARE.
 
+import { Expression } from '../v2';
 import { ParamSpec } from '../v2/params/types';
 
 /**
  * An definition of a function as appears in the Manifest.
  */
-export interface ManifestEndpoint {
+export type ManifestEndpoint = {
```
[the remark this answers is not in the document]
Berlioz: Uh, it wasn't, I originally did something really dumb with another type that extended ManifestEndpoint (which does require the change), reverted it, and forgot to revert this.

src/runtime/manifest.ts
```diff
@@ -52,32 +53,32 @@ export interface ManifestEndpoint {
...
 
   eventTrigger?: {
-    eventFilters: Record<string, string>;
-    eventFilterPathPatterns?: Record<string, string>;
+    eventFilters: Record<string, string | Expression<string>>;
+    eventFilterPathPatterns?: Record<string, string | Expression<string>>;
     channel?: string;
     eventType: string;
-    retry: boolean;
+    retry: boolean | Expression<string>;
```
TheIronDev, line 60: Shouldn't this be `Expression<boolean>`?

src/v2/expressions.ts
```diff
@@ -0,0 +1,100 @@
+/*
+ * A CEL expression which can be evaluated during function deployment, and
+ * resolved to a value of the generic type parameter: i.e, you can pass
+ * an Expression<number> as the value of an option that normally accepts numbers.
+ */
+export abstract class Expression<
```
TheIronDev, line 6: In addition to this, I wonder if things would be cleaner if the two following existed

```
export type ExpressionOrType<T> = T | Expression<T>;
export type NullableExpressionOrType<T> = ExpressionOrType<T> | null;
```

The reason I bring this up is because I saw an earlier instance of `boolean | Expression<string>` that may not have been intentional. This hardens the typing slightly.

WDYT?

src/v2/expressions.ts
TheIronDev: yeah its definitely a very optional nit.

src/v2/expressions.ts
```diff
@@ -0,0 +1,100 @@
...
+        return left === this.rhs;
+      case '>':
+        return left > this.rhs;
+      case '>=':
+        return left >= this.rhs;
+      case '<':
+        return left < this.rhs;
+      case '<=':
+        return left <= this.rhs;
+    }
```
TheIronDev, line 90: Nit: shouldn't there be a default case (even if just to throw an error) ?

src/v2/params/index.ts
```diff
@@ -54,6 +55,32 @@ function registerParam(param: Param) {
...
+/**
+ * For testing.
+ * @internal
+ */
+export function clearParams() {
+  declaredParams.splice(0, declaredParams.length);
+}
+
+/**
+ * Declare a secret param, that will persist values only in Cloud Secret Manager.
```
TheIronDev, line 67: Nit: should start as third-person:

`Declares`

taeold (approved):
Minor nits on memory option, but I don't think that should be solved in the PR. this PR has stayed unmerge for too long!

src/v2/params/index.ts
```diff
@@ -54,6 +56,32 @@ function registerParam(param: Param) {
   declaredParams.push(param);
 }
 
+/**
+ * For testing.
+ * @internal
+ */
+export function clearParams() {
+  declaredParams.splice(0, declaredParams.length);
```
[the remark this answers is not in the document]
Berlioz: It would be if `declaredParams` wasn't declared const.

src/v2/params/index.ts
```diff
@@ -54,6 +56,32 @@ function registerParam(param: Param) {
...
+ * Secrets are stored interally as bytestrings. Use ParamOptions.`as` to provide type
+ * hinting during parameter resolution.
+ *
+ * @param name The name of the environment variable to use to load the param.
+ * @param options Configuration options for the param.
+ * @returns A Param with a `string` return type for `.value`.
+ */
+export function defineSecretParam(
+  name: string
+  //options: ParamOptions<string> = {}
```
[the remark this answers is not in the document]
Berlioz: fixed.

src/runtime/manifest.ts
Berlioz: Yes. Fixed.

src/v2/expressions.ts
Berlioz: Yup.

src/v2/params/index.ts
Berlioz: Done.

src/v2/params/index.ts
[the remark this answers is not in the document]
taeold: hmm it's starting to sound like declaredParams isn't a constant variable afterall...

PR #1172, opened 2022-07-19
Updating layout to point to new _project.yaml file in c/461672987.

egilmorez:


egilmorez:
This PR is rendered unnecessary by updates to api-documenter that effectively obsolete the TypeDoc process.

We should probably clean it out after we feel comfortable that api-documenter is going to work going forward.

PR #1173, opened 2022-07-19
Pin typescript and typedoc to compatible versions.

taeold:
We rely on a now deprecated feature of typedoc for theming v1 reference docs. Unfortunately, the new version of typedoc completely remove support for theming the way it is done today.

Old typedoc version is compatible w/ TS version upto 4.4.0 (latest stable version is at 4.7.0 today), so here we pin both TS and typedoc to the compatible version.

In theory we can get out of pinning TS version by migrating v1 typedoc to api-documenter toolchain that we are already using for v2 reference docs. I don't think we can prioritize that work right now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,8 +56,8 @@
         "tslint-config-prettier": "^1.18.0",
         "tslint-no-unused-expression-chai": "^0.1.4",
         "tslint-plugin-prettier": "^2.0.1",
-        "typedoc": "0.23.7",
-        "typescript": "^4.3.5",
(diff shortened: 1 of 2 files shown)


colerogers (approved):

egilmorez (approved):
Thanks for sending!

inlined (approved):

inlined:
I don't *think* we require features more advanced than 4.4? Have we tried pushing this on `launch.vnext` to be sure? `ParamsOf` is pretty advanced.

PR #1175, opened 2022-07-20
Update api-documenter to the latest version supporting new project header

taeold:
Let's `npm run docgen:v1` away!


egilmorez (approved):
Thanks!

PR #1177, opened 2022-07-21
Add v2 Schedule Triggers

colerogers:
Adds schedule triggers to the v2 namespace.

- Replaces pub/sub with an http underlying function
- Updated syntax to match with the other v2 triggers

This change does not add the `invoker` property to the container contract.

```
export const sch = v2.scheduler.onSchedule("* * * * *", () => {});

export const sch = v2.scheduler.onSchedule(
  {
    schedule: "* * * * *",
    timeZone: "utc",
  },
  () => {}
);
```


Berlioz:
Note to self: some of the types in this trigger are also going to need the Field/Expression treatment.

colerogers:
@Berlioz I can update this PR to take the work off your plate, what example should I be looking at?

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import * as express from 'express';
+import { logger } from '../..';
```
taeold, line 24: nit* this is equivalent to 

```
import * as logger from "../../logger"
```

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+import { HttpsFunction } from './https';
+
+/** @hidden */
+interface ScheduleArgs {
+  schedule: string;
+  timeZone?: timezone;
+  retryCount?: number;
+  maxRetryDuration?: string;
+  minBackoffDuration?: string;
+  maxBackoffDuration?: string;
```
taeold, line 38: Not the best time to have this discussion, but I'm surprised that this is `*Duration` instead of `*Seconds`. That's what we did for Task Queue Functions (even though the underlying Tasks API uses `*Duration`).

On one hand, I'd like consistency across v2. On the other hand, this breaks consistency w/ v1 scheduled function! What do you think we should do?

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * Handler for scheduled functions. Triggered whenever the associated
+ * scheduler job sends a http request.
+ * @param args - Either a schedule or an object containing function options.
+ * @param handler - A function to execute when triggered.
+ * @returns A function that you can export and deploy.
+ */
+export function onSchedule(
+  args: string | ScheduleOptions,
+  handler: (req: express.Request) => any | Promise<any>
```
taeold, line 136: Should `handler` signature be a little more specific? e.g. `(req ...) => void | Promise<void>`

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+  const separatedOpts = getOpts(args);
+
+  const func: any = async (
+    req: express.Request,
+    res: express.Response
+  ): Promise<any> => {
+    try {
+      await handler(req);
+    } catch (err) {
+      logger.warn((err as Error).message);
```
taeold, line 147: why not `logger.error`?

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    res.setHeader('Content-Type', 'application/json');
```
taeold, line 150: q: why is this header required? Does Cloud Scheduler require it?

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+    res.status(200).send();
```
taeold, line 151: why do we always respond w/ 200? Doesn't this mean that we won't retry execution w/ errors?

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+  delete (opts as any).schedule;
+  delete (opts as any).timeZone;
+  delete (opts as any).retryCount;
+  delete (opts as any).maxRetryDuration;
+  delete (opts as any).minBackoffDuration;
+  delete (opts as any).maxBackoffDuration;
+  delete (opts as any).maxDoublings;
+
+  scheduleArgs.opts = opts;
+  return scheduleArgs as ScheduleArgs;
```
taeold, line 70: I'm not 100% sure if we need this separation logic. Can't we instead add `ScheduleOptions` to list of type accepted by `optionsToEndpoint`? Does that lead to some other error I'm not seeing here?

https://github.com/firebase/firebase-functions/blob/9b6da2d8d12c70914e5ca6258c50b56239e4ad88/src/v2/options.ts#L291

src/v2/providers/schedule.ts
colerogers: I believe this is from vscode's auto import

src/v2/providers/schedule.ts
colerogers: Hmm that's an interesting point. A counter argument is that the user could have a string value that's not in seconds. The API uses `1h` as an example in this case. IDK if we should name the interface something that could limit the customers choices. But on the other hand, we try to have an opinionated SDK.  🤷‍♂️ 

Also, I'm not sure what you mean by breaking consistency w/ v1 scheduled functions since they use the `Duration` naming too.

src/v2/providers/schedule.ts
colerogers: Yeah, void makes sense in this context

src/v2/providers/schedule.ts
colerogers: The current behavior for v1 is that a function error will not retry the execution. A retry will occur when the scheduler job fails. I opted for consistency and just return a 200. Let me know if you think we should change this up for v2.

src/v2/providers/schedule.ts
colerogers: I don't think it's required, just leftover cruft, I'll remove it

src/v2/providers/schedule.ts
taeold: Meant that `*Seconds` breaks consistency w/ Task Queue API. Lots of parameters here look like TQ API's parameter - I would find it weird to see TQ and Scheduled function use slightly different parameters for configuration.

src/v2/providers/schedule.ts
taeold: I think it's also good to reduce code duplication! One thing that stuck out to me here is the number of times we have to repeat scheduled function configs.

One trick that I'm seeing in other namespace is that we just cast `ScheduledOptions` as `GlobalOptions` and pass that to `options.optionsToEndpoint`. Would that work here too? Did you think that was gross?

src/v2/providers/schedule.ts
taeold: Hm following footsteps of v1 scheduled function seems reasonable, though I would guess having scheduled functions automatically retry on transient errors to be pretty useful. This is something we explicitly set for event functions via `failurePolicy` and I can see it being applied here too.

But if you think that is a feature that doesn't need to be implemented v2 scheduled function, that sounds fine with me. Maybe we'll get a FR for it one day, and we can implement it then?

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,198 @@
...
+    'maxRetryDuration',
+    'minBackoffDuration',
+    'maxBackoffDuration',
+    'maxDoublings'
+  );
+  // TODO(colerogers): add invoker to scheduleTrigger
+  // and the container contract if we want to
+  // support enterprise customers and change the
+  // behavior of using the default compute service agent as
+  // the function invoker from cloud scheduler.
```
taeold, line 186: q: is this a quick follow up or something we want to punt until we have explicit request?

(I imagine it will be a pita to test this internally :()

src/runtime/manifest.ts
```diff
@@ -63,7 +63,7 @@ export interface ManifestEndpoint {
 
   scheduleTrigger?: {
     schedule?: string;
-    timezone?: string;
+    timeZone?: string;
```
taeold, line 66: hmm this seems like a tricky migration - would changing possibly break users trying to deploy v1 scheduled function w/ CLI that relies/copies `timezone` property instead?

colerogers: Nope, v1 schedule functions use the correct camel case `timeZone` - https://github.com/firebase/firebase-functions/blob/9b6da2d8d12c70914e5ca6258c50b56239e4ad88/src/providers/pubsub.ts#L151

This field wasn't actually directly used in v1 at all. The container contract spec & the CLI both have this being camel case, so this would have been a bug if we left it lowercased

src/v2/providers/schedule.ts
colerogers: I was leaving this a follow up since I didn't want to wait for another API review since this is modifying the container contact. But I can block on that if you think we should launch with custom invoker support

src/runtime/manifest.ts
taeold: Oh this is **fixing* a typo on `manifest.ts` 🤦🏼. 

Sigh I wish TypeScript helped but looks like TS is not going to help w/ situations like this ([playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgILIN4Chm+SOAWwgH4AuZAZzClAHMc8xhiAtAexFIutpAYC+WLKEixEKAEKZGuAsXJUa9WcmbEAXp25K+g4Qk7VkcCugC8mNSwgcuFAESV2xa661cHyIQBsIYZAAjCmlLOABuZAB6KPx2ZGgodigSIA)).

src/v2/providers/schedule.ts
taeold: sgtm. Maybe we can create a ticket/github issue instead. I know that is as good as throwing a ticket down a black hole, but I hope it's more discoverable in the future.

src/v2/providers/schedule.ts
colerogers: After internal discussion, I'll change from *Duration to *Seconds for the v2 api

src/v2/providers/schedule.ts
colerogers: SG, created an internal issue

src/v2/providers/schedule.ts
colerogers: updated the PR to just cast instead of creating a new object & deleting fields

src/v2/providers/schedule.ts
colerogers: SG, updated to send a 500 if we catch an error.

integration_test/functions/src/index.ts
```diff
@@ -1,9 +1,9 @@
 import { PubSub } from '@google-cloud/pubsub';
-import { GoogleAuth } from 'google-auth-library';
 import { Request, Response } from 'express';
 import * as admin from 'firebase-admin';
 import * as functions from 'firebase-functions';
 import * as fs from 'fs';
+import { GoogleAuth } from 'google-auth-library';
```
colerogers, line 6: linter caught this

src/v2/providers/schedule.ts
```diff
@@ -0,0 +1,182 @@
...
+    maxBackoffSeconds: args.maxBackoffSeconds,
+    maxDoublings: args.maxDoublings,
+    opts: args as options.GlobalOptions,
+  };
+}
+
+/** The Cloud Function type for Schedule triggers. */
+export interface ScheduleFunction extends HttpsFunction {
+  __requiredAPIs?: ManifestRequiredAPI[];
+  run(data: express.Request): void | Promise<void>;
```
inlined, line 66: Do we really need to take an express request? Do we really pass that to customers? This makes mocking/testing difficult.

inlined: Looking at the code here, it looks like we can start by making `run` take no arguments. Maybe we can take a look at the HTTP request that Cloud Scheduler sends to see if there's anything we'd like to expose (e.g. scheduled deliver time and retry count) but we should try to do something smarter than just forwarding a hard-to-mock object.

inlined (changes requested):

package.json
```diff
@@ -66,7 +66,8 @@
     "./v2/alerts/crashlytics": "./lib/v2/providers/alerts/crashlytics.js",
     "./v2/eventarc": "./lib/v2/providers/eventarc.js",
     "./v2/identity": "./lib/v2/providers/identity.js",
-    "./v2/database": "./lib/v2/providers/database.js"
+    "./v2/database": "./lib/v2/providers/database.js",
+    "./v2/schedule": "./lib/v2/providers/schedule.js"
```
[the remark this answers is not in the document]
colerogers: In the proposal, we had `v2/schedule`, I'll make the change and run it through API council with the *Seconds change

src/v2/providers/schedule.ts
colerogers: Per our discussion in chat, I've changed this to a `ScheduleEvent` object

src/v2/providers/schedule.ts
taeold: I'd vote for keeping v1 the same.

taeold (approved):

taeold:
Force merging - @inlined's comments have all been addressed.

kbi-daniel:
The new V2 onSchedule cloud function does not seem to have a way to set timeoutSeconds or memory options.

For instance, my V1 scheduled cloud function currently sets both using runWith():
`exports.function = functions.runWith({timeoutSeconds: 300, memory: '512MB'}).pubsub.schedule('0 22 * * 0-6').onRun(() => {...});`

However, the currently listed ScheduleOptions listed in the V2 documentation does not list either of these properties:
https://firebase.google.com/docs/functions/beta/reference/firebase-functions.scheduler.scheduleoptions

Are these properties going to be added to ScheduleOptions?

taeold:
Hey @kbi-daniel! The ScheduleOptions extends the base [GlobalOptions](https://firebase.google.com/docs/functions/beta/reference/firebase-functions.globaloptions.md#globaloptions_interface), so options like timeout and memory should be available. Eg.

```js
exports.v2schedule = onSchedule(
    {
        // ScheduleOptions
        schedule: "every 2 minutes",
        maxBackoffSeconds: 20,
        // GlobalOptions
        memory: "1GiB",
        timeoutSeconds: 50,
    }, (event) => {
    console.log(event);
});

```

kbi-daniel:
Thanks @taeold.  Worked as expected!

PR #1179, opened 2022-07-25
Fix integration tests

taeold:
One notable change made here is that we generate and include ID tokens in requests to http/callable functions since we aren't allowed to run public functions on internal GCP project

### DIFF
--- a/integration_test/functions/src/index.ts
+++ b/integration_test/functions/src/index.ts
@@ -1,5 +1,5 @@
 import { PubSub } from '@google-cloud/pubsub';
-import  { GoogleAuth } from "google-auth-library";
+import { GoogleAuth } from 'google-auth-library';
 import { Request, Response } from 'express';
 import * as admin from 'firebase-admin';
 import * as functions from 'firebase-functions';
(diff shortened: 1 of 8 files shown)


integration_test/functions/src/index.ts
```diff
@@ -42,31 +45,36 @@ async function callV2HttpsTrigger(
   data: any,
   accessToken: string
 ) {
-  let resp = await fetch(
+  const resp0 = await fetch(
```
colerogers, line 48: nit: I personally would name these responses something more descriptive, like `fnLookupResp` & `fnInvokeResp`

colerogers (approved):

integration_test/functions/src/index.ts
```diff
@@ -22,17 +23,19 @@ const firebaseConfig = JSON.parse(process.env.FIREBASE_CONFIG);
...
-    {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json',
-      },
-      body: JSON.stringify({ data }),
-    }
+  const url = `https://${REGION}-${firebaseConfig.projectId}.cloudfunctions.net/${name}`;
+  const client = await new GoogleAuth().getIdTokenClient(
+    '32555940559.apps.googleusercontent.com'
```
inlined, line 28: What is this hard-coded project-specific string?

inlined: Doing some research, this is the CLI appID? Whatever it is, can't we assume that the default compute service account has invoker permissions and use the ADC?

PR #1181, opened 2022-07-27
Fix bug where callable function skipped unrecognized auth headers 

taeold:
We have a weird edge case where an authorization in form we don't recognize will completely skip the auth check.

The fix here applies 2 changes:

1) We allow 'Bearer <TOKEN>' format to be case insensitive. 'bearer <TOKEN>' also works.

2) We reject other authorization header. e.g. 'Beaver <token>' is rejected.

### DIFF
--- a/src/common/providers/https.ts
+++ b/src/common/providers/https.ts
@@ -638,7 +638,7 @@ export async function checkAuthToken(
   if (!authorization) {
     return 'MISSING';
   }
-  const match = authorization.match(/^Bearer (.*)$/);
+  const match = authorization.match(/^Bearer (.*)$/i);
   if (match) {
     const idToken = match[1];
     try {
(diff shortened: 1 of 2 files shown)


src/common/providers/https.ts
```diff
@@ -638,7 +638,7 @@ export async function checkAuthToken(
   if (!authorization) {
     return 'MISSING';
   }
-  const match = authorization.match(/^Bearer (.*)$/);
+  const match = authorization.match(/^Bearer (.*)$/i);
   if (match) {
```
inlined, line 642: nit: maybe instead of indenting the whole function in the general case of "match" we can instead early exit with "invald" if `!match` and simplify the whole function.

inlined (approved):

Issue #1135, opened 2022-06-08
sessionClaims content not getting added to the decoded token


patzj:
### Related issues

N/A

### [REQUIRED] Version info

**node:** 16.14.0

**firebase-functions:** 3.21.2

**firebase-tools:** 10.7.1

**firebase-admin:** 10.2.0

### [REQUIRED] Test case

https://cloud.google.com/identity-platform/docs/blocking-functions#setting_custom_and_session_claims

#### Web App
```js
export default function Home() {
  const { data: user } = useUser();
  const auth = getAuth();

  function signIn() {
    const provider = new GoogleAuthProvider();
    signInWithPopup(auth, provider).then(async (value) => {
      const idTokenResult = await value.user.getIdTokenResult(true);
      console.log(idTokenResult);
    });
  }

  if (!user) {
    return <button onClick={signIn}>Sign In</button>;
  }

  return (
    <div>
      <span>{user.displayName}</span>&nbsp;
      <button onClick={() => signOut(auth)}>Sign Out</button>
    </div>
  );
}
```

#### Cloud Function

```ts
export const authUserBeforeCreate = functions
  .region(region)
  .auth.user()
  .beforeCreate((user, context) => {
  // codes
});

export const authUserBeforeSignIn = functions
  .region(region)
  .auth.user()
  .beforeSignIn(async (_, context) => {
    return {
      displayName: "Raging Tomato",
      sessionClaims: { signInIpAddress: context.ipAddress },
    };
  });
});
```

### [REQUIRED] Steps to reproduce

The project was originally a GCP project and is using Identity Platform but I had to setup Firebase in it for various reasons.

1. Deploy a `beforeSignIn` Cloud Function via `firebase-tools`
2. In the web app, sign in the user using `signInWithPopup` from `firebase/auth`
3. Print the `idTokenResult`
```ts
export const authUserBeforeCreate = functions
  .region(region)
  .auth.user()
  .beforeCreate((user, context) => {
  // codes
});

export const authUserBeforeSignIn = functions
  .region(region)
  .auth.user()
  .beforeSignIn(async (_, context) => {
    return {
      displayName: "Raging Tomato",
      sessionClaims: { signInIpAddress: context.ipAddress },
    };
  });
});
```

### [REQUIRED] Expected behavior

The `sessionsClaims` I returned from the `beforeSignIn` Cloud Function should be included in the decoded `idToken`

### [REQUIRED] Actual behavior

[...]

PR #1184, opened 2022-07-28
Modify getUpdateMask to no longer ignore customClaims and sessionClaims

Metevier:
### Description

Attempts to fix: #1135 - sessionClaims content not getting added to the decoded token 

From my personal testing, adding `sessionClaims` or `customerClaims` through a blocking function does not work. 
The emulator code is expecting both of these keys to show up in the `updateMask` of the blocking function's response.

https://github.com/firebase/firebase-tools/blob/30771f6498f32791cd2e88ac142611df58d32b7e/src/emulator/auth/operations.ts#L3083-L3117

I of course do not know what the internal firebase auth platform is expecting to receive in the `updateMask`, so these changes may not be the correct approach. 

The test that is removed in the PR notes that the updateMask "should return empty on only customClaims and sessionClaims", and the consuming code will actually return an empty response if only the claims values are set: 

https://github.com/firebase/firebase-functions/blob/cc18326068278dcf2f5e15bf38d8ca5f5aaa9ff2/src/common/providers/identity.ts#L863-L872

I do not see any mention of this requirement in the documentation: https://firebase.google.com/docs/auth/extend-with-blocking-functions 

I am happy to update these fixes if the internal firebase auth implementation is expecting a different response (which will also require an update to the firebase-tools emulators).

### DIFF
--- a/spec/common/providers/identity.spec.ts
+++ b/spec/common/providers/identity.spec.ts
@@ -765,19 +765,6 @@ describe('identity', () => {
       expect(identity.getUpdateMask()).to.eq('');
     });
 
-    it('should return empty on only customClaims and sessionClaims', () => {
-      const response = {
-        customClaims: {
-          claim1: 'abc',
-        },
-        sessionClaims: {
-          claim2: 'def',
-        },
-      };
-
-      expect(identity.getUpdateMask(response)).to.eq('');
-    });
-
     it('should return the right claims on a response', () => {
       const response = {
         displayName: 'john',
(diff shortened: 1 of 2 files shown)


gorkamolero:
Same happening for customClaims on beforeCreate function. It doesn't work. Can we get some help?

taeold:
@Metevier Thank you for your contribution and sorry we missed this PR.

We just merged an exact same fix this morning - https://github.com/firebase/firebase-functions/pull/1199. We should've caught your PR request and merged it a month ago. I apologize for completely messing up here.

PR #1187, opened 2022-08-08
Update `InAppFeedbackPayload`

jladieu:
* Add `feedbackConsoleUri`
* Clarify fast expiry of `screenshotUris`

### DIFF
--- a/src/v2/providers/alerts/appDistribution.ts
+++ b/src/v2/providers/alerts/appDistribution.ts
@@ -53,6 +53,8 @@ export interface InAppFeedbackPayload {
   ['@type']: 'type.googleapis.com/google.events.firebase.firebasealerts.v1.AppDistroInAppFeedbackPayload';
   /** Resource name. Format: `projects/{project_number}/apps/{app_id}/releases/{release_id}/feedbackReports/{feedback_id}` */
   feedbackReport: string;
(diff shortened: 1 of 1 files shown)


src/v2/providers/alerts/appDistribution.ts
```diff
@@ -71,8 +73,8 @@ export interface InAppFeedbackPayload {
   buildVersion: string;
   /** Text entered by the tester */
   text: string;
-  /** URIs to download screenshot(s) */
-  screenshotUris?: string[];
+  /** URIs to download screenshot(s). These URIs are fast expiring. */
```
inlined, line 76: It's not enough to just change the type annotation. ESF won't send down any value in JSON if the array is empty. We're going to need to go into the actual handler code and say `event.data.payload.screenshotUris = event.data.payload.screenshotUris ?? []`

jladieu: Oh dang.  I'm not actually familiar enough with this codebase to make that change with confidence, so I thought it'd be a good idea to decouple it, so I just split out the change from optional to non-optional for this PR and put that in a separate https://github.com/firebase/firebase-functions/pull/1188

colerogers (approved):
looks good, thanks!

jladieu:
Thank you so much!

PR #1188, opened 2022-08-10
Make `screenshotUris` non-optional.

jladieu:
If omitted, they should be defaulted to an empty list.

### DIFF
--- a/src/v2/providers/alerts/appDistribution.ts
+++ b/src/v2/providers/alerts/appDistribution.ts
@@ -72,7 +72,7 @@ export interface InAppFeedbackPayload {
   /** Text entered by the tester */
   text: string;
   /** URIs to download screenshot(s) */
-  screenshotUris?: string[];
+  screenshotUris: string[];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


inlined (commented):
Starting on line 332, edit the following code:

```
  const func = (raw: CloudEvent<unknown>) => {
    return handler(raw as AppDistributionEvent<InAppFeedbackPayload>);
  };
  ```
  
  to be:
  
  ```
  const func = (raw: CloudEvent<unknown>) => {
    const event = raw as AppDistributionEvent<InAppFeedbackPayload>;
    // Consolidate the case of empty array and null array
    event.data.payload.screenshotUris = event.data.payload.screenshotUris || [];
    return handler(event);
  };
  ```

File: src/v2/providers/alerts/appDistribution.ts [hunk 1]
```diff
@@ -330,7 +330,10 @@ export function onInAppFeedbackPublished(
   const [opts, appId] = getOptsAndApp(appIdOrOptsOrHandler);
 
   const func = (raw: CloudEvent<unknown>) => {
-    return handler(raw as AppDistributionEvent<InAppFeedbackPayload>);
+    const event = raw as AppDistributionEvent<InAppFeedbackPayload>;
```
jladieu, line 333: @inlined Thanks for the coaching, fix applied.

inlined (approved):

PR #1189, opened 2022-08-12
fix: prioritize boolean cors value over debug flags

stfsy:
### Description

As discussed in https://github.com/firebase/firebase-tools/issues/4862 it's not possible to disable cors in v2 functions when running in the emulator, because the emulator enables the debug feature for cors overriding `opts.cors`.

If we could prioritize the boolean flag over the debug feature, we could have both: The ability to disable cors and also prioritize the debug flag over user provided configuration.


src/v2/providers/https.ts
```diff
@@ -219,7 +219,7 @@ export function onRequest(
     opts = optsOrHandler as HttpsOptions;
   }
 
-  if (isDebugFeatureEnabled('enableCors') || 'cors' in opts) {
+  if (opts.cors !== false && (isDebugFeatureEnabled('enableCors') || 'cors' in opts)) {
```
taeold, line 222: oof lots conditions here that's making my head spin 🤯 

What if we do this in longer form:

```suggestion
  if (isDebugFeatureEnabled('enableCors') || 'cors' in opts) {
    let origin = opts.cors;
    if (isDebugFeatureEnabled('enableCors')) {
      // Respect `cors: false` to turn off cors even if debug feature is enabled.
      origin = opts.cors === false ? false : true 
    }
  }
```

taeold (changes requested):

stfsy:
@taeold updated as per suggestion and rebased

taeold (approved):
LGTM - please fix formatting!

Thank you for an awesome contribution!

stfsy:
@taeold anything else I can do?

stfsy:
@taeold got it now, had to run `npm run format:fix` to fix the formatting error. I committed and pushed the changes just a few seconds ago.

colerogers:
@stfsy I fixed your merge conflicts but it looks like you need to run `npm run format` one more time before we can merge this.

stfsy:
@colerogers thank you very much 🙂 I ran `npm run format` and commited+pushed the results

PR #1192, opened 2022-08-19
Automatically add trace property to v2 function logs

taeold:
All traffic to Cloud Run revisions are populated with `traceparent` and `X-Cloud-Trace-Context` header [doc](https://cloud.google.com/run/docs/trace) and Functions Framework helpfully processes and injects the traceparent on cloudevent payload (See https://github.com/GoogleCloudPlatform/functions-framework/issues/47).

We take advantage of this and automatically inject `trace` property to all logs generated from `functions.logger`. All logs emitted from functions logger then are correlated with requests - and it even works when concurrency > 1.

Traced logs are awesome with Cloud Logging:

image

And looks great in Cloud Trace too!

image


TheIronDev (approved):
Lgtm!

yblok13:
Trace property is not automatically added to logger.write().

PR #1197, opened 2022-08-23
More ESLint Migration

taeold:
Turns out I missed most of the files for migration from prettier to eslint in https://github.com/firebase/firebase-functions/pull/1194 due to poorly configured `.eslintignore` - it ignored all files under `src/v1` and `src/v2`!

Correctly configure eslint and run `npm run format`. Again, I had to manually fixup issues like:

1. Remove empty function definition (prefer ()=>undefined over ()=>{})
1. Remove unused arguments from function definitions


taeold:
Again, I'm going to merge this PR without waiting for a review. I don't expect a human to be able to review this PR. Instead, we will rely on integration tests and bugbash to help surface any accidental regression.

PR #1198, opened 2022-08-23
Fix integration tests

taeold:
Integration test caught some bugs!

1. From https://github.com/firebase/firebase-functions/pull/1155 - Some internal type must be exposed - otherwise compiling user projects fails w/ error message like below:

```
Error: node_modules/firebase-functions/lib/common/params.d.ts(10,13): error TS2314: Generic type 'Extract' requires 2 type argument(s).
Error: node_modules/firebase-functions/lib/common/params.d.ts(10,21): error TS2304: Cannot find name 'Split'.
Error: node_modules/firebase-functions/lib/common/params.d.ts(10,27): error TS2304: Cannot find name 'NullSafe'.
```

2. From https://github.com/firebase/firebase-functions/pull/1181 - There is no way to call private callable functions since GCF requrie GCP identity token as authorization header while callable function requires Firebase auth tokens. Disabling those test cases in the integration test until we can mint a GCP project w/ permission to create public functions.

### DIFF
--- a/integration_test/functions/src/index.ts
+++ b/integration_test/functions/src/index.ts
@@ -20,6 +20,8 @@ import * as testLab from "./v1/testLab-utils";
 const firebaseConfig = JSON.parse(process.env.FIREBASE_CONFIG);
 admin.initializeApp();
 
+// Re-enable no-unused-var check once callable functions are testable again.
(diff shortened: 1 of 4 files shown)


TheIronDev (approved):
LGTM

PR #1199, opened 2022-08-23
Fixes a bug that disallowed setting customClaims and/or sessionClaims in blocking functions

colerogers:
- Adds `customClaims` and `sessionClaims` to the updateMask.


taeold (approved):
It sounds like pulling customClaims and sessionClaims out was intentional at some point - what changed?

colerogers:
@taeold Originally, we had the update mask function recurse into keys and stopped if we hit custom or session claims. We changed behavior & rewrote the code, but never removed the the if check and tests 🤦‍♂️

PR #1201, opened 2022-08-24
Update Change docs

TheIronDev:
Update Change docs

### DIFF
--- a/src/common/change.ts
+++ b/src/common/change.ts
@@ -39,6 +39,7 @@ export interface ChangeJson {
   fieldMask?: string;
 }
 
+/** @internal */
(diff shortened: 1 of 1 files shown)


src/common/change.ts
```diff
@@ -34,13 +34,11 @@ export interface ChangeJson {
    */
   before?: any;
   /**
-   * @hidden
    * Comma-separated string that represents names of fields that changed.
    */
   fieldMask?: string;
 }
 
-/** @hidden */
```
taeold, line 43: should this be @internal instead?

taeold (approved):

src/common/change.ts
TheIronDev: Yeah I think it should! 👍

PR #1208, opened 2022-08-24
Update firestore docs

TheIronDev:
Update v1 Firestore documentation

### DIFF
--- a/src/v1/providers/firestore.ts
+++ b/src/v1/providers/firestore.ts
@@ -31,11 +31,8 @@ import * as logger from "../../logger";
 import { CloudFunction, Event, EventContext, makeCloudFunction } from "../cloud-functions";
 import { DeploymentOptions } from "../function-configuration";
 
-/** @hidden */
 export const provider = "google.firestore";
-/** @hidden */
 export const service = "firestore.googleapis.com";
-/** @hidden */
 export const defaultDatabase = "(default)";
 let firestoreInstance: any;
 export type DocumentSnapshot = firestore.DocumentSnapshot;
@@ -52,19 +49,16 @@ export function document<Path extends string>(path: Path) {
   return _documentWithOptions(path, {});
 }
 
-/** @hidden */
 // Multiple namespaces are not yet supported by Firestore.
 export function namespace(namespace: string) {
   return _namespaceWithOptions(namespace, {});
 }
 
-/** @hidden */
 // Multiple databases are not yet supported by Firestore.
 export function database(database: string) {
   return _databaseWithOptions(database, {});
 }
 
-/** @hidden */
 export function _databaseWithOptions(
   database: string = defaultDatabase,
   options: DeploymentOptions
@@ -72,18 +66,15 @@ export function _databaseWithOptions(
   return new DatabaseBuilder(database, options);
 }
 
-/** @hidden */
 export function _namespaceWithOptions(namespace: string, options: DeploymentOptions) {
   return _databaseWithOptions(defaultDatabase, options).namespace(namespace);
 }
 
-/** @hidden */
 export function _documentWithOptions<Path extends string>(path: Path, options: DeploymentOptions) {
   return _databaseWithOptions(defaultDatabase, options).document(path);
 }
 
 export class DatabaseBuilder {
-  /** @hidden */
   constructor(private database: string, private options: DeploymentOptions) {}
 
   namespace(namespace: string) {
@@ -96,7 +87,6 @@ export class DatabaseBuilder {
 }
 
 export class NamespaceBuilder {
-  /** @hidden */
   constructor(
     private database: string,
     private options: DeploymentOptions,
@@ -142,7 +132,6 @@ function _getValueProto(data: any, resource: string, valueFieldName: string) {
   return proto;
 }
 
-/** @hidden */
 export function snapshotConstructor(event: Event): DocumentSnapshot {
   if (!firestoreInstance) {
     firestoreInstance = firestore.getFirestore(getApp());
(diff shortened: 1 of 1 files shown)


src/v1/providers/firestore.ts
```diff
@@ -52,38 +49,32 @@ export function document<Path extends string>(path: Path) {
...
 }
 
-/** @hidden */
 // Multiple databases are not yet supported by Firestore.
 export function database(database: string) {
   return _databaseWithOptions(database, {});
 }
 
-/** @hidden */
 export function _databaseWithOptions(
```
TheIronDev, line 62: IMO `_databaseWithOptions`, `_namespaceWithOptions`, `_documentWithOptions` can all be possibly removed as an export. `functions-builder` appears to use them, but we can clone-n-own in the separate file.

Thoughts

src/v1/providers/firestore.ts
```diff
@@ -31,11 +31,8 @@ import * as logger from "../../logger";
 import { CloudFunction, Event, EventContext, makeCloudFunction } from "../cloud-functions";
 import { DeploymentOptions } from "../function-configuration";
 
-/** @hidden */
 export const provider = "google.firestore";
```
TheIronDev, line 34: Maybe this is a personal nit but, shouldn't these three const strings be uppercased?

src/v1/providers/firestore.ts
```diff
@@ -96,7 +87,6 @@ export class DatabaseBuilder {
 }
 
 export class NamespaceBuilder {
```
TheIronDev, line 89: firestore.ts is the only thing use this. Does this need to be exported still?

Should it be @internal instread?

src/v1/providers/firestore.ts
```diff
@@ -175,7 +163,6 @@ function changeConstructor(raw: Event) {
 }
 
 export class DocumentBuilder<Path extends string> {
```
TheIronDev, line 165: Same thing with DocumentBuilder. This class is only used in this file. Remove `export` or `@internal`?

src/v1/providers/firestore.ts
taeold: Please mark this internal instead.

Let's try to keep the work to minimum - we don't necessarily have to fix things.

src/v1/providers/firestore.ts
taeold: Please mark this @internal instead.

if you mean variable name should be caps, maybe! if you mean the string value, i don't think so (they might map to eventflow event type)

src/v1/providers/firestore.ts
taeold: I think this object ultimately gets exposed on customer code.

Marking it @internal might break `tsc` on customer code 🤔

src/v1/providers/firestore.ts
taeold: Same thing - this is probably exposed to customer and must be exported and not be internal. (I could be wrong though).

taeold (changes requested):

src/v1/providers/firestore.ts
```diff
@@ -52,38 +49,32 @@ export function document<Path extends string>(path: Path) {
   return _documentWithOptions(path, {});
 }
 
-/** @hidden */
 // Multiple namespaces are not yet supported by Firestore.
 export function namespace(namespace: string) {
```
taeold, line 53: If not hidden, can we try to documen this?

We can also consult w/ @egilmorez when he's back to document this section if you aren't sure how to document this.

src/v1/providers/firestore.ts
TheIronDev: 🤔  Leaving it then..

src/v1/providers/firestore.ts
TheIronDev: I don't have enough context on this..

src/v1/providers/firestore.ts
TheIronDev: Marked as internal it is!

src/v1/providers/firestore.ts
TheIronDev: renamed to caps and added `@internal`

src/v1/providers/firestore.ts
```diff
@@ -31,9 +31,12 @@ import * as logger from "../../logger";
 import { CloudFunction, Event, EventContext, makeCloudFunction } from "../cloud-functions";
 import { DeploymentOptions } from "../function-configuration";
 
-export const provider = "google.firestore";
-export const service = "firestore.googleapis.com";
-export const defaultDatabase = "(default)";
+/** @internal */
+export const PROVIDER = "google.firestore";
```
taeold, line 35: "technically" renaming an expose variable name is a breaking change.

Truthfully, I don't really care for that kind of breaking changes, but I'd err on the safe side and not change the casing here unless there are other practical purpose I'm missing.

src/v1/providers/firestore.ts
taeold: Let's leave it blank for now.

taeold (approved):
one minor nit on casing. otherwise lgtm.

src/v1/providers/firestore.ts
TheIronDev: Fair point, reverted this. 👍

src/v1/providers/firestore.ts
TheIronDev: kko!

src/v1/providers/firestore.ts
TheIronDev: SGTM!

PR #1216, opened 2022-08-31
Add v2 Remote Config Triggers

colerogers:
Adds remote config triggers to CF3v2

```
import { onConfigUpdated } from 'firebase-functions/v2/remoteConfig';

export const fn = onConfigUpdated((event) => { console.log("hello world"); });
```


TheIronDev (approved):
LGTM!

taeold (approved):
Loving the doc comments. Do you want to mark these triggers as @alpha until we have samples/docs/CLI support merged and released?

taeold (approved):
One more thing - do you want to add integration test target for this (in this PR or later in another PR).

colerogers:
@taeold - added the `@alpha` tags & since this change requires CLI changes to be released, I'll follow up with a int test in another PR

PR #1218, opened 2022-08-31
Move params as top level import

taeold:
Instead of:

```
require("firebase-functions/v2/params");
```

We now have:

```
require("firebase-functions/params");
```

A breaking change indeed.


Berlioz:
You seem to have missed src/v2/providers/identity.ts. based on CI, but otherwise LGTM.

Berlioz (approved):

PR #1223, opened 2022-09-06
Adds performance monitoring triggers to v2 alerts

colerogers:
Adds Firebase Performance Monitoring triggers for performance threshold alerts to the v2 namespace.

```
// listen on all apps in a project
export const fn = onThresholdAlertPublished((event: PerformanceEvent<ThresholdAlertPayload>) => { … });

// listen on a specific app with a supplied appId
export const fn2 = onThresholdAlertPublished("1234567", (event: PerformanceEvent<ThresholdAlertPayload>) => { … });

// listen on a all apps with an options object
export const fn3 = onThresholdAlertPublished(
  {
    region: "us-central1",
  },
  (event: PerformanceEvent<ThresholdAlertPayload>) => { … }
);

// listen on a specific app with a supplied appId in the options object
export const fn4 = onThresholdAlertPublished(
  {
    appId: "1234567",
    region: "us-central1",
  },
  (event: PerformanceEvent<ThresholdAlertPayload>) => { … }
);
```

### DIFF
--- a/src/v2/providers/alerts/performance.ts
+++ b/src/v2/providers/alerts/performance.ts
@@ -158,15 +158,13 @@ export function onThresholdAlertPublished(
 export function getOptsAndApp(
   appIdOrOpts: string | PerformanceOptions
 ): [EventHandlerOptions, string | undefined] {
-  let opts: EventHandlerOptions;
-  let appId: string | undefined;
-  if (typeof appIdOrOpts === 'string') {
-    opts = {};
-    appId = appIdOrOpts;
-  } else {
-    appId = appIdOrOpts.appId;
-    opts = { ...appIdOrOpts };
-    delete (opts as any).appId;
+  if (typeof appIdOrOpts === "string") {
+    return [{}, appIdOrOpts];
   }
+
+  const opts: EventHandlerOptions = { ...appIdOrOpts };
+  const appId: string | undefined = appIdOrOpts.appId;
+  delete (opts as any).appId;
+
   return [opts, appId];
 }
--- a/src/v2/providers/alerts/performance.ts
+++ b/src/v2/providers/alerts/performance.ts
@@ -47,9 +47,9 @@ export interface ThresholdAlertPayload {
   /* The unit for the alert threshold (e.g. "percent", "seconds") */
   thresholdUnit: string;
   /* The percentile of the alert condition, can be 0 if percentile is not applicable to the alert condition; range: [0, 100] */
-  conditionPercentile: number;
+  conditionPercentile?: number;
   /* The app version this alert was triggered for, can be empty if the alert is for a network request (because the alert was checked against data from all versions of app) or a web app (where the app is versionless) */
-  appVersion: string;
+  appVersion?: string;
   /* The value that violated the alert condition (e.g. "76.5", "3") */
   violationValue: number;
   /* The unit for the violation value (e.g. "percent", "seconds") */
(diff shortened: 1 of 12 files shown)


src/v2/providers/alerts/performance.ts
```diff
@@ -0,0 +1,165 @@
...
+ */
+
+import { CloudEvent, CloudFunction } from "../../core";
+import { EventHandlerOptions } from "../../options";
+import { FirebaseAlertData, getEndpointAnnotation } from ".";
+import { wrapTraceContext } from "../../trace";
+
+/**
+ * The internal payload object for a performance threshold alert.
+ * Payload is wrapped inside a `FirebaseAlertData` object.
```
TheIronDev, line 35: nit: `{@link FirebaseAlertData}`

src/v2/providers/alerts/performance.ts
```diff
@@ -0,0 +1,165 @@
...
+  func.__endpoint = getEndpointAnnotation(opts, thresholdAlert, appId);
+
+  return func;
+}
+
+/**
+ * Helper function to parse the function opts and appId.
+ * @internal
+ */
+export function getOptsAndApp(
```
TheIronDev, line 151: Nit: you can simplify this a bit:

```
export function getOptsAndApp(
  appIdOrOpts: string | PerformanceOptions
): [EventHandlerOptions, string | undefined] {
  if (typeof appIdOrOpts === "string") {
    return [{}, appIdOrOpts];
  }
  
  const opts: EventHandlerOptions = { ...appIdOrOpts };
  const appId: string | undefined = appIdOrOpts.appId;
  delete (opts as any).appId;
  
  return [opts, appId];
}

```

src/v2/providers/alerts/performance.ts
```diff
@@ -0,0 +1,165 @@
...
+  appIdOrOpts: string | PerformanceOptions
+): [EventHandlerOptions, string | undefined] {
```
TheIronDev, line 153: Not really a nit or complaint, but curious why returning populated array instead of an object containing the fields?

```
[EventHandlerOptions, string | undefined]
vs
{opts: EventHandlerOptions, appId?: string}
```

TheIronDev (approved):
LGTM

src/v2/providers/alerts/performance.ts
colerogers: looks good, thanks for the improvement!

colerogers: Just b/c it was only a couple fields, I did a similar thing for the other alert triggers

taeold (approved):

src/v2/providers/alerts/performance.ts
```diff
@@ -0,0 +1,165 @@
...
+export function onThresholdAlertPublished(
+  opts: PerformanceOptions,
+  handler: (event: PerformanceEvent<ThresholdAlertPayload>) => any | Promise<any>
+): CloudFunction<PerformanceEvent<ThresholdAlertPayload>>;
+
+/**
+ * Declares a function that can handle receiving performance threshold alerts.
+ * @param appIdOrOptsOrHandler - A specific application, options, or an event-handling function.
+ * @param handler - Event handler which is run every time a threshold alert is received.
+ * @returns A function that you can export and deploy.
```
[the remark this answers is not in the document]
colerogers: Oh good catch, but I think that remote config should actually be changed. Looking at all the other alert functions (& most other v2 functions) we say 'A function...'.

I remember @egilmorez giving us guidance here a while back, but not sure which way

PR #1227, opened 2022-09-08
Bring SDK Expression types into alignment with CLI + unit tests for Expression

Berlioz:
Writing tests revealed, uh. A number. Of bugs in the current Expression implementation:

0) (big feature we forgot) **a param's default needs to be an Expression in addition to a literal**
1) (feature alignment) comparisons should be able to accept another Expression as the rhs
2) (feature alignment) ternary operators should be able to provide Expressions as the if_true/if_false values
3) (bug) ternary operators should not be forced to be of the same type as the params being compared (i.e `{{ params.INT2 == params.INT2 ? params.STRING : "foo" }}` was illegal)
4) (bug) boolean params would always evaluate to true no matter what string value the CLI passed in
5) (feature alignment) we should support != as a comparison

### DIFF
--- a/spec/params/params.spec.ts
+++ b/spec/params/params.spec.ts
@@ -1,6 +1,16 @@
 import { expect } from "chai";
 import * as params from "../../src/params";
 
+describe("Params spec extraction", () => {
+  it("converts Expressions in the param default to strings", () => {
+    const bar = params.defineInt("BAR");
+    expect(
+      params.defineString("FOO", { default: bar.notEquals(22).then("asdf", "jkl;") }).toSpec()
+        .default
+    ).to.equal(`{{ params.BAR != 22 ? "asdf" : "jkl;" }}`);
+  });
+});
+
 describe("Params value extraction", () => {
   beforeEach(() => {
     process.env.A_STRING = "asdf";
--- a/spec/params/params.spec.ts
+++ b/spec/params/params.spec.ts
@@ -1,0 +1,202 @@
+import { expect } from "chai";
+import * as params from "../../src/params";
+
+describe("Params value extraction", () => {
+  beforeEach(() => {
+    process.env.A_STRING = "asdf";
+    process.env.SAME_STRING = "asdf";
+    process.env.DIFF_STRING = "jkl;";
+    process.env.AN_INT = "-11";
+    process.env.SAME_INT = "-11";
+    process.env.DIFF_INT = "22";
+    process.env.PI = "3.14159";
+    process.env.TRUE = "true";
+    process.env.FALSE = "false";
+  });
+
+  afterEach(() => {
+    params.clearParams();
+    delete process.env.A_STRING;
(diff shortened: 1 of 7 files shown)


spec/params/params.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+    const sameInt = params.defineInt("SAME_INT");
+    const diffInt = params.defineInt("DIFF_INT");
+    expect(int.equals(sameInt).value()).to.be.true;
+    expect(int.equals(-11).value()).to.be.true;
+    expect(int.equals(diffInt).value()).to.be.false;
+    expect(int.equals(22).value()).to.be.false;
+  });
+
+  it("can use all the comparison operators when explicitly requested", () => {
+    const jkl = params.defineString("DIFF_STRING");
```
taeold, line 85: does comparison operator support comparing params?

e.g. does this work?

```
const asdf = params.defineString("A_STRING")
expect(jkl.cmp(">", asdf).value()).to.be.true;
```

spec/params/params.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+    expect(boolParam.value()).to.be.true;
+
+    const floatParam = params.defineFloat("PI");
+    expect(floatParam.value()).to.equal(3.14159);
+
+    const falseParam = params.defineBoolean("FALSE");
+    expect(falseParam.value()).to.be.false;
+  });
+
+  it("falls back on the javascript zero values in case of type mismatch", () => {
```
taeold, line 58: hmm fallbacks are tricky. We don't expect these to happen in real life no? (b/c cli will ensure that the values exist)

src/params/types.ts
```diff
@@ -183,26 +214,54 @@ export abstract class Param<T extends string | number | boolean | string[]> exte
...
       name: this.name,
-      ...this.options,
+      ...otherOptions,
       type: (this.constructor as typeof Param).type,
     };
 
-    return out as ParamSpec<T>;
+    if (paramDefault instanceof Expression) {
+      out.default = paramDefault.toCEL();
+    } else if (paramDefault) {
```
taeold, line 260: what is `paramDefault = false`? Should this condition be instead `paramDefault !== undefined` or something?

taeold (approved):
few comments, mostly good to my eyes.

src/params/types.ts
Berlioz: yeah, that's what I was meaning to capture

spec/params/params.spec.ts
Berlioz: Correct.

spec/params/params.spec.ts
Berlioz: It does (that was the impact of changing the `rhs` type in a ComparisonExpression to be `rhs: T | Expression<T>`) but you're right in that I should add tests

PR #1231, opened 2022-09-12
WIP: Create an InternalExpression type which allows precanned access to env.FIREBASE_CONFIG

Berlioz:
WIP because the tests don't exist, because the test file i really should be adding to them is in #1227.

### DIFF
--- a/src/params/types.ts
+++ b/src/params/types.ts
@@ -41,6 +41,33 @@ export abstract class Expression<T extends string | number | boolean | string[]>
   }
 }
 
+/**
+ * A CEL expression which represents an internal Firebase variable. This class
+ * cannot be instantiated by developers, but we provide several canned instances
+ * of it to make available params that will never have to be defined at
+ * deployment time, and can always be read from process.env.
+ */
+export class InternalExpression extends Expression<string> {
+  constructor(
+    private readonly name: string,
+    private readonly getter: (env: NodeJS.ProcessEnv) => string
+  ) {
+    super();
+  }
+
+  value(): string {
+    return this.getter(process.env) || "";
+  }
+
+  toString(): string {
+    return `params.${this.name}`;
+  }
+
+  toCEL(): string {
+    throw new Error("An InternalExpression should never be marshalled for wire transmission.");
+  }
+}
+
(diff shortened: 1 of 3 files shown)


src/params/types.ts
```diff
@@ -41,6 +41,33 @@ export abstract class Expression<T extends string | number | boolean | string[]>
   }
 }
 
+/**
+ * A CEL expression which represents an internal Firebase variable. This class
+ * cannot be instantiated by developers, but we provide several canned instances
+ * of it to make available params that will never have to be defined at
+ * deployment time, and can always be read from process.env.
+ */
+export class InternalExpression extends Expression<string> {
```
taeold, line 50: Should this be marked @hidden or @beta or @alpha or whatever so we don't expose it to our end user?

I wish we can use @internal, but I'm afraid that might break user code that tries to reference variables with that type 🤔

inlined: No, it should be @internal and the variable should be declared as an Expression<string>. Type erasure is great for hiding implementation details and not tying our hands to them.

Berlioz: That's a good point. Done.

inlined:
Unfortunately the part you're trying not to do is the part I specifically want to do. I want people to be able to say:

```
const bucket = defineResource("BUCKET", {
  input: {
    resourceSelector: { type: "storage.googleapis.com/Bucket" },
   },
   default: params.storageBucket,
});
```

That means you're going to need to support both a `toCEL` and a CEL parser for something like `{{ params.FIREBASE_CONFIG.bucket }}` in the CLI

inlined:
Hmmm... on second thought I can see how (for now) you can turn that into literal values in the CEL expression?... there may be less work than anticipated (e.g. not extending the parser) but the user journey is what I'm hoping for.

inlined:
I think your magic transformation of FIREBASE_CONFIG to params.DEFAULT_BUCKET et. al. might be on the right track. The extensions team is looking at adding a _bunch_ of new expressions, and I think I've negotiated that they will not be published as environment variables (since we'll use up the user's list). So I think we should strongly consider having a list of magic envs in the CLI that we have values for but don't write to .env files. Everything you've done here is a good candidate.

PR #1233, opened 2022-09-15
Remove unusable list param definition function from autocompletion

Berlioz:


src/params/index.ts
```diff
@@ -134,6 +134,7 @@ export function defineFloat(name: string, options: ParamOptions<number> = {}): F
 /**
  * Declare a list param (array of strings).
  *
+ * @hiddden
```
taeold, line 137: I don't think hidden works 🤔. You need to make it @internal or just get rid of the the implementation.

taeold (approved):

inlined:
Quick notes:

1. `@hidden` hides the API from VS Code's code complete while allowing the syntax
2. `@alpha` hides the API from API docs (I believe)
3. `@internal` drops the API from module exports

If we depend on the API inside the `firebase-functions` binary, I don't recall if we can actually use `@internal`. Test before you release.

Berlioz:
We're axing the function entirely, not making it as hidden, but that explanation was helpful.

PR #1236, opened 2022-09-19
Converting alert type and app id to camel case in the CloudEvent

colerogers:
Fixes a bug where the SDK expected camel cased fields, but eventarc did not, so we convert them before passing the event to the end users code.

### DIFF
--- a/spec/v2/providers/alerts/alerts.spec.ts
+++ b/spec/v2/providers/alerts/alerts.spec.ts
@@ -1,4 +1,5 @@
 import { expect } from 'chai';
+import { CloudEvent } from '../../../../src/v2';
 import * as options from '../../../../src/v2/options';
 import * as alerts from '../../../../src/v2/providers/alerts';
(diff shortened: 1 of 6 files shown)


taeold (approved):

marcgovi:
Adding a note here that CloudEvent specs explicitly mention that extension context attributes (eg: appid and alerttype for Firebase Alerts) must be lowercase https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#extension-context-attributes.

I'm worried that the firebase functions is converting them to camel case format instead before delivering the CloudEvent object to the customer function.

colerogers:
@marcgovi We had a discussion offline, but posting here for visibility. 

Basically, we try to match the TS/JS norms in the Functions SDK to give the best possible developer experience. In this case, converting fields to camelcase in the event signature. We have a history of changing the event with other integrations too, for example with RTDB, we convert fields to camelcase and also create a custom `ref` object so customers don't have to import the Admin SDK directly. We think these quality of life improvements make a difference to developers and should be included in the SDK

PR #1241, opened 2022-09-22
Adds v2 test lab triggers

colerogers:
This change adds test lab triggers to the v2 codebase and sneaks in some remote config changes. For remote config, we update the tsdoc strings for reference docs and remove @alpha tags since the public docs have been published. 

Test Lab functions:
```
import {
  onTestMatrixCompleted,
  TestMatrixCompletedData
} from "firebase-functions/v2/testLab";

// passing in a handler only
export const fn1 = onTestMatrixCompleted((event) => { ... });

// declaring options
export const fn2 = onTestMatrixCompleted(
  { region: "us-central1", retry: true },
  (event) => {  ...  }
);

```

### DIFF
--- a/src/v2/providers/testLab.ts
+++ b/src/v2/providers/testLab.ts
@@ -23,6 +23,7 @@
 import { ManifestEndpoint } from "../../runtime/manifest";
 import { CloudEvent, CloudFunction } from "../core";
 import { EventHandlerOptions, getGlobalOptions, optionsToEndpoint } from "../options";
+import { wrapTraceContext } from "../trace";
 
 /** @internal */
 export const eventType = "google.firebase.testlab.testMatrix.v1.completed";
@@ -189,7 +190,7 @@ export function onTestMatrixCompleted(
   const specificOpts = optionsToEndpoint(optsOrHandler);
 
   const func: any = (raw: CloudEvent<unknown>) => {
-    return handler(raw as CloudEvent<TestMatrixCompletedData>);
+    return wrapTraceContext(handler(raw as CloudEvent<TestMatrixCompletedData>));
   };
   func.run = handler;
 
--- a/src/v2/providers/testLab.ts
+++ b/src/v2/providers/testLab.ts
@@ -20,111 +20,194 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-/* Possible test states for a test matrix. */
+import { ManifestEndpoint } from "../../runtime/manifest";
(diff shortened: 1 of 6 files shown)


src/v2/providers/testLab.ts
```diff
@@ -0,0 +1,213 @@
...
+ * @alpha
+ */
+export function onTestMatrixCompleted(
+  optsOrHandler:
+    | EventHandlerOptions
+    | ((event: CloudEvent<TestMatrixCompletedData>) => any | Promise<any>),
+  handler?: (event: CloudEvent<TestMatrixCompletedData>) => any | Promise<any>
+): CloudFunction<CloudEvent<TestMatrixCompletedData>> {
+  if (typeof optsOrHandler === "function") {
+    handler = optsOrHandler as (event: CloudEvent<TestMatrixCompletedData>) => any | Promise<any>;
```
TheIronDev, line 184: Any chance we can change this to not be any | Promise ? (same with others)

src/v2/providers/testLab.ts
```diff
@@ -0,0 +1,213 @@
...
+    optsOrHandler = {};
+  }
+
+  const baseOpts = optionsToEndpoint(getGlobalOptions());
+  const specificOpts = optionsToEndpoint(optsOrHandler);
+
+  const func: any = (raw: CloudEvent<unknown>) => {
+    return handler(raw as CloudEvent<TestMatrixCompletedData>);
```
taeold, line 192: can we add `wrapTraceContext`

taeold (approved):

src/v2/providers/testLab.ts
TheIronDev: I guess my main question was along the lines of if we could avoid the `any`.

But if theres prior art, and would require api review again... I'm alright leave it as-is.

src/v2/providers/testLab.ts
```diff
@@ -0,0 +1,213 @@
...
+// SOFTWARE.
+
+import { ManifestEndpoint } from "../../runtime/manifest";
+import { CloudEvent, CloudFunction } from "../core";
+import { EventHandlerOptions, getGlobalOptions, optionsToEndpoint } from "../options";
+
+/** @internal */
+export const eventType = "google.firebase.testlab.testMatrix.v1.completed";
+
+/** Possible test states for a test matrix. */
```
TheIronDev, line 30: Btw I love all the documentation!!!

TheIronDev:
I love all the thoroough documentation!!!

TheIronDev (approved):

src/v2/providers/testLab.ts
colerogers: Added

PR #1245, opened 2022-09-28
Hide documentation for in-app feedback

kaibolay:
should be reverted once the feature is launched

### DIFF
--- a/src/v2/providers/alerts/appDistribution.ts
+++ b/src/v2/providers/alerts/appDistribution.ts
@@ -53,6 +53,8 @@ export interface NewTesterDevicePayload {
 /**
  * The internal payload object for receiving in-app feedback from a tester.
  * Payload is wrapped inside a `FirebaseAlertData` object.
+ *
+ * @internal
  */
 export interface InAppFeedbackPayload {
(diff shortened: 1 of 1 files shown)


norealityagent (approved):

src/v2/providers/alerts/appDistribution.ts
```diff
@@ -53,6 +53,8 @@ export interface NewTesterDevicePayload {
 /**
  * The internal payload object for receiving in-app feedback from a tester.
  * Payload is wrapped inside a `FirebaseAlertData` object.
+ *
+ * @internal
```
taeold, line 57: I'd consider using `@alpha`.

`@internal`: typescript tag. Removes the function entirely out of build output.
`@alpha`: api-extractor tag used for reference doc generation. Keeps the function in build output allowing us to test it. No reference doc will be generated

taeold (approved):

src/v2/providers/alerts/appDistribution.ts
kaibolay: Thanks for the tip! Done.

PR #1246, opened 2022-09-29
Fix bug where function configuration with null couldn't be deployed.

taeold:
```
> (typeof null === "object")
true
```


Berlioz (approved):
Yikes. +1

PR #1247, opened 2022-09-30
Allow actual RegExp objects in the regex field of a textInput

Berlioz:
I'm not....proud of this hack, but I was writing up some docs for params and frankly I couldn't get the slash escapes right in my head when typing them out, which means it's _really_ not reasonable to expect users to get them right either. Much better to just let them use the same JS regexps that they're used to using, and can copy-paste test.

Is this going to require APi review? The actual wire format hasn't changed.

### DIFF
--- a/src/runtime/manifest.ts
+++ b/src/runtime/manifest.ts
@@ -102,9 +102,11 @@ export interface ManifestStack {
  * @internal
  */
 export function stackToWire(stack: ManifestStack): Record<string, unknown> {
-  for (const param of stack.params) {
-    if ("text" in param.input && param.input.text.validationRegex instanceof RegExp) {
-      param.input.text.validationRegex = param.input.text.validationRegex.source;
+  if (stack.params) {
+    for (const param of stack.params) {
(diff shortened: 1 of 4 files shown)


src/runtime/manifest.ts
```diff
@@ -102,6 +102,14 @@ export interface ManifestStack {
  * @internal
  */
 export function stackToWire(stack: ManifestStack): Record<string, unknown> {
+  if (stack.params) {
```
Berlioz, line 105: In retrospect, I think there's a good argument to be made that this ugly special case should probably live in the generic toSpec() implementation in params/types.ts?

taeold: nice "hack" with super great usability upgrade. Do you want to do that here or on a separate PR.

taeold (approved):

Berlioz:
1) it turns out that we do not need API review to do it like this for the slightly hilarious reason that i misread the design doc and we were always supposed to do it like this
2) the active ingredient of this patch has indeed been moved into `toSpec()` as we discussed

PR #1249, opened 2022-09-30
allow Expressions in the top-level configuration of v1 functions

Berlioz:


taeold (approved):

Berlioz:
Gaaaack. Wrong PR. At least launch.next being unprotected meant I could revert the click immediately.

PR #1250, opened 2022-10-03
Add RESET_VALUE configuration value as sentinel value to reset a config option.

taeold:
Instead of allowing `null`s to reset a configuration value, we are introducing a new sentinel value, `RESET_VALUE` to make it clear the intention of resetting the config value to platform default:

```js
import * as functions from "firebase-functions/v1";
import { onRequest } from "firebase-functions/v2/https";
import { RESET_VALUE } from "firebase-functions/v2/options";

export const v1Req = functions
  .runWith({
    minInstances: functions.RESET_VALUE,
    memory: functions.RESET_VALUE,
    serviceAccount: RESET_VALUE
  }).https.onRequest(
    (req, res) => {
      res.sendStatus(200)
    },
);

export const v2Req = onRequest(
  {
   memory: RESET_VALUE,
   minInstances: RESET_VALUE,
  },
  (req, res) => {
    res.sendStatus(200)
  },
);
```

Majority of this PR concerns fixing/updating typings on various configuration options.

Recall that in v2, we inlined/repeat all function options in each respective providers to make reference doc more clear. This unfortunately bloated up this PR since I had to repeat the changes in each and every provider.

### DIFF
--- a/src/runtime/manifest.ts
+++ b/src/runtime/manifest.ts
@@ -38,10 +38,10 @@ export interface ManifestEndpoint {
   timeoutSeconds?: number | Expression<number> | ResetValue;
   vpc?: {
     connector: string | Expression<string> | ResetValue;
-    egressSettings?: string;
+    egressSettings?: string | Expression<string> | ResetValue;
   };
-  ingressSettings?: string | ResetValue;
-  serviceAccountEmail?: string;
+  ingressSettings?: string | Expression<string> | ResetValue;
+  serviceAccountEmail?: string | Expression<string> | ResetValue;
   cpu?: number | "gcf_gen1";
   labels?: Record<string, string>;
   environmentVariables?: Record<string, string>;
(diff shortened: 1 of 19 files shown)


src/runtime/manifest.ts
```diff
@@ -30,19 +31,19 @@ export interface ManifestEndpoint {
...
+  availableMemoryMb?: number | Expression<number> | ResetValue;
+  maxInstances?: number | Expression<number> | ResetValue;
+  minInstances?: number | Expression<number> | ResetValue;
+  concurrency?: number | Expression<number> | ResetValue;
+  timeoutSeconds?: number | Expression<number> | ResetValue;
   vpc?: {
-    connector: string | Expression<string>;
-    egressSettings?: string;
+    connector: string | Expression<string> | ResetValue;
+    egressSettings?: string | Expression<string> | ResetValue;
```
TheIronDev, line 41: was making this an `Expression` intentional?

src/runtime/manifest.ts
```diff
@@ -30,19 +31,19 @@ export interface ManifestEndpoint {
...
   };
+  ingressSettings?: string | Expression<string> | ResetValue;
+  serviceAccountEmail?: string | Expression<string> | ResetValue;
+  cpu?: number | "gcf_gen1";
   labels?: Record<string, string>;
-  ingressSettings?: string;
```
TheIronDev, line 45: Was removal intentional?

src/v2/providers/tasks.ts
```diff
@@ -144,9 +147,6 @@ export interface TaskQueueOptions extends options.EventHandlerOptions {
    * Secrets to bind to a function.
    */
   secrets?: string[];
-
-  /** Whether failed executions should be delivered again. */
-  retry?: boolean;
```
TheIronDev, line 149: Removal intentional?

TheIronDev (approved):
Had a few questions to make sure some changes were intentional, otherwise LGTM

src/v2/providers/tasks.ts
taeold: Thanks for asking! Yes - it's intentional. I don't think this function make sense for task queue functions; we already have `retryConfig` setup and largely ignore this config option.

src/runtime/manifest.ts
taeold: Yes - just reordering a bit here.

colerogers (approved):

Issue #1248, opened 2022-09-30
ObjectMetadata property "name" is optional


xxRockOnxx:
### Related issues

None

### [REQUIRED] Version info

**node:**

**firebase-functions:** 3.24.0

**firebase-tools:**

**firebase-admin:**

### [REQUIRED] Test case

```
functions.storage.object().onFinalize((object) => {
  if (object?.name) {
  }
})
```

### [REQUIRED] Steps to reproduce

- Open editor/IDE
- Copy code from test case
- Check if property `name` is optional

### [REQUIRED] Expected behavior

property `name` is required

Image taken from https://firebase.google.com/docs/reference/functions/firebase-functions.storage.objectmetadata.md#storageobjectmetadata_interface
![image](https://user-images.githubusercontent.com/11251134/193204859-3490ff45-7510-4655-8f68-b3354ce2e9cd.png)

### [REQUIRED] Actual behavior

property `name` is optional

https://github.com/firebase/firebase-functions/blob/v3.24.0/src/providers/storage.ts#L263-L265

Is this intentional? Is there really times where `name` is non-existent?

Or is the docs correct and this is just a typo/mistake? I could send a PR since it's just a removal of `?` from the type.

### Were you able to successfully deploy your functions?

Yes

inlined:
the bucket and object name will always be present. We will fix the typing for this in upcoming SDK 4.0.0

taeold:
Fixed per https://github.com/firebase/firebase-functions/pull/1251

PR #1251, opened 2022-10-03
Document some storage fields as required

inlined:
Documents some fields as required in response to #1248.

Consulted with an old contact from the v1/EventFlow days to get this minimum list of fields which are always present.


TheIronDev (approved):
Sweet!

PR #1256, opened 2022-10-05
Allow SecretParams to be directly registered in v1/v2 options

Berlioz:
open questions:
1) do we actually not have tests corresponding to spec/v1/function-builder for the options struct in v2? or did i just not find them?
2) do we want to introduce a helper type for (string | SecretParam)[]? My gut instinct is no since it's a user-facing composite type.


taeold (approved):

taeold:
1. looks like we don't have any spec for `optionsToEndpoint` but instead they are implicitly tested by each provider spec.
2. Agreed

PR #1257, opened 2022-10-10
Print a warning when Expression.value() is invoked during discovery

Berlioz:
This is the neatest way I can think of to implement this off the top of my head, which means there's an excellent chance that there's a neater way to do it that I didn't think of.

### DIFF
--- a/src/params/types.ts
+++ b/src/params/types.ts
@@ -35,7 +35,7 @@ export abstract class Expression<T extends string | number | boolean | string[]>
         `Parametrized expression ${this.toString()}.value() invoked during Function discovery. This is usually a mistake.`
       );
       logger.warn(
-        `To configure a Function with an expression object, use it directly without calling .value().`
+        `To configure a function with a parameter, use it directly without calling .value().`
       );
     }
     return this.runtimeValue();
--- a/src/params/types.ts
+++ b/src/params/types.ts
@@ -20,6 +20,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+import * as logger from "../logger";
+
 /*
  * A CEL expression which can be evaluated during function deployment, and
  * resolved to a value of the generic type parameter: i.e, you can pass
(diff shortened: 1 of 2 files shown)


src/params/types.ts
```diff
@@ -28,6 +30,19 @@
...
+        `Parametrized expression ${this.toString()}.value() invoked during Function discovery. This is usually a mistake.`
+      );
+      logger.warn(
+        `To configure a Function with an expression object, use it directly without calling .value().`
+      );
+    }
+    return this.runtimeValue();
+  }
+
+  // @internal
```
taeold, line 44: Do we need @internal tag on all of implementation of `runtimeValue`?

src/params/types.ts
```diff
@@ -28,6 +30,19 @@
 export abstract class Expression<T extends string | number | boolean | string[]> {
   // Returns the Expression's runtime value, based on the CLI's resolution of params.
   value(): T {
+    if (process.env.FIREBASE_DISCOVERY_DIR) {
+      logger.warn(
+        `Parametrized expression ${this.toString()}.value() invoked during Function discovery. This is usually a mistake.`
+      );
+      logger.warn(
+        `To configure a Function with an expression object, use it directly without calling .value().`
```
taeold, line 38:
```suggestion
        `To configure a function with a parameter, use it directly without calling .value().`
```

Some suggestions:

1) Let's avoid using unfamiliar Expression, even if it's technically more correct term.

2) If we can make this message super verbose, can we add an example?

src/params/types.ts
```diff
@@ -69,8 +84,8 @@ export class TernaryExpression<
     this.ifFalse = ifFalse;
   }
 
-  value(): T {
-    return this.test.value() ? valueOf(this.ifTrue) : valueOf(this.ifFalse);
+  runtimeValue(): T {
+    return this.test.runtimeValue() ? valueOf(this.ifTrue) : valueOf(this.ifFalse);
```
taeold, line 88: woah we are renaming `value` to `runtimeValue()`?

src/params/types.ts
Berlioz: Possibly, and while I was down the rabbit hole of looking that up, I also realized that tsc doesn't respect jsdoc annotations unless they're in, well, jsdoc format, so fixing that too.

src/params/types.ts
Berlioz: Not from users' perspective. They'll still call value(), which will use the implementation in the base Expression class, which will do the discovery check and then call runtimeValue()

src/params/types.ts
```diff
@@ -28,6 +30,20 @@
 export abstract class Expression<T extends string | number | boolean | string[]> {
   // Returns the Expression's runtime value, based on the CLI's resolution of params.
   value(): T {
+    if (process.env.FIREBASE_DISCOVERY_DIR) {
+      logger.warn(
+        `${this.toString()}.value() invoked during Function configuration, instead of during runtime.`
```
taeold, line 35:
```suggestion
        `${this.toString()}.value() invoked during function deployment, instead of during runtime.`
```

wdyt?

taeold (approved):

PR #1259, opened 2022-10-12
Remove manifest types from API.

taeold:
Right now, `launch.next` has some issues with inconsistently exposing types that concerns the container contract manifest.

This is sweep to make all Manifest-related types internal. Developers shouldn't need these types for their daily uses anyway.


Berlioz:
Yeah there's just no universe in which anyone should be instantiating any of these types, agreed.

Berlioz (approved):

PR #1260, opened 2022-10-13
Re-export manifest types for use by internal packages.

taeold:
Reverts https://github.com/firebase/firebase-functions/pull/1259 - instead do it the "right" way.


Berlioz (approved):
+1

PR #1261, opened 2022-10-13
Update release script to support making release candidates

taeold:
Adding a new flag to allow us to create a release candidates with versions like `4.0.0-rc.0`.

Instructions added on go/publish-via-cloud-build.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -88,10 +88,10 @@ npm run build:release
 echo "Ran publish build."
 
 echo "Making a $VERSION version..."
-if [[ $PRE_RELEASE == "" ]]; then
-  npm version $VERSION
-else
+if [[ $PRE_RELEASE != "" ]]; then
   npm version pre$VERSION --preid=rc
+else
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -88,13 +88,13 @@ npm run build:release
 echo "Ran publish build."
 
 echo "Making a $VERSION version..."
-# TODO: Remove the following command.
-#   npm version command had previously failed claiming unclean git repo, and we don't know why.
-echo "DEBUG: Running git status to show dirty files..."
-git status
-npm version $VERSION
+if [[ $PRE_RELEASE == "" ]]; then
```
colerogers, line 91: Should we switch this if statement around since below we make the inverted check? Might be easier readability if they're all the same direction

colerogers (approved):

scripts/publish.sh
taeold: goooood call.

tfang17:
This is breaking deploys when upgrading.

PR #1267, opened 2022-10-18
Fix wrong paths to package.json typesVersions

### LINKED ISSUE #1266: package.json typesVersions contains start slash for some providers
### [REQUIRED] Version info
4.0.0

**node:16.10.0**

**firebase-functions: 4.0.0**

**firebase-tools:**

**firebase-admin:11.2.0**

### [REQUIRED] Test case

Before I could import `ObjectMetadata` by using `import { ObjectMetadata } from 'firebase-functions/lib/providers/storage'`

but in 4.0.0 it's gone (which is good, never liked to do deep imports)

I looked at the package.json of 4.0.0 and It should be possible to import via the `firebase-functions/v1/storage`
but If you look at a couple of the paths are wrong.

remoteConfig
storage
tasks
testLab

contain a / in the beginning, whilst all others don't
I think this is what is causing the problem.

```
"v1/firestore": [
  "./lib/v1/providers/firestore"
],
"v1/pubsub": [
  "./lib/v1/providers/pubsub"
],
"/v1/remoteConfig": [
  "./lib/v1/providers/remoteConfig"
],
"/v1/storage": [
  "./lib/v1/providers/storage"
],
"/v1/tasks": [
  "./lib/v1/providers/tasks"
],
"/v1/testLab": [
  "./lib/v1/providers/testLab"
**],**
```

to be able to import from

`import { ObjectMetadata } from 'firebase-functions/v1/storage'`

to get it to work know I need to add a double / at the beginning

`import { ObjectMetadata } from 'firebase-functions//v1/storage'`

leon:
typedVersions should not contain slash at the start of the path.

Fixes #1266

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,16 +89,16 @@
       "v1/pubsub": [
         "./lib/v1/providers/pubsub"
       ],
-      "/v1/remoteConfig": [
+      "v1/remoteConfig": [
         "./lib/v1/providers/remoteConfig"
       ],
-      "/v1/storage": [
+      "v1/storage": [
         "./lib/v1/providers/storage"
       ],
-      "/v1/tasks": [
+      "v1/tasks": [
         "./lib/v1/providers/tasks"
       ],
-      "/v1/testLab": [
+      "v1/testLab": [
         "./lib/v1/providers/testLab"
       ],
       "v2": [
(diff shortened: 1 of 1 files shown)


inlined (approved):

inlined:
Thank you for contributing to our SDK!

PR #1281, opened 2022-10-26
Give secret params a `.value()` method

### LINKED ISSUE #1264: Environment Secrets have no value
### [REQUIRED] Version info

**node:** v16.17.1

**firebase-functions:** 4.0.0

**firebase-tools:** 11.4.4

**firebase-admin:** 11.1.0

### [REQUIRED] Test case
Firebase init
install latest version of admin/tools and functions.  (as my versions)
import defineSecret 
`import { defineSecret } from "firebase-functions/params";`
define functions to run with that secret
`runWith({secrets:[my_secret]})`
try to get the value inside the function

```js
import { defineSecret } from "firebase-functions/params";

// Defined in GCloud or not, we just don't have access to its value!
const my_secret = defineSecret("my_secret");
// Remember to define runWith the secrets, (here only one in the array) 
// as it should only be pupulated at runtime from GCloud
export const helloWorld = functions.runWith({secrets:[my_secret]}).https.onRequest((request, response) => {
  my_secret.value(); // Property 'value' does not exist on type 'SecretParam'.ts(2339)
});
```
now
`my_secret.value();`
is undefined. SecretParam (the thing defineSecret returns) only has a .name parameter, no .value()

Expected to be able to reference .value() method as per the documentation
https://firebase.google.com/docs/functions/config-env

cannot access the value :(
**Property 'value' does not exist on type 'SecretParam'.ts(2339)*

no

jhuleatt:
### Description

Fixes #1264 

### Code sample

```ts
const { defineSecret } = require('firebase-functions/params');
const discordApiKey = defineSecret('DISCORD_API_KEY');

export const postToDiscord = functions.runWith({ secrets: [discordApiKey] }).https.onRequest(
  (req, res) => {
    const apiKey = discordApiKey.value();
    //…
```

https://firebase.google.com/docs/functions/config-env#secret_parameters


taeold (approved):

taeold:
Can we add a changelog? I think this is a bugfix.

jhuleatt:
Thanks, I added a changelog!

A couple of questions:

1. Do we need to add any guards to prevent this param type from being used outside of runtime, since secret values are only available to the function at runtime? Or does not inheriting from `Parameter` solve this for us?
2. Should the `.value` function throw an error if the secret value isn't available to the function (for example, if a developer forgot to add the `secrets` config array to the function)?

src/params/types.ts
```diff
@@ -345,7 +345,13 @@ export class SecretParam {
 
   /** @internal */
   runtimeValue(): string {
-    return process.env[this.name] || "";
+    const val = process.env[this.name];
+    if (!val) {
```
Berlioz, line 349: couldn't someone deliberately set a secret to the empty string, which is falsy?

Berlioz (approved):
Other than one nit, LGTM.

src/params/types.ts
jhuleatt: Changed from a falsy check to `val === undefined`

Berlioz:
LGTM

taeold (approved):

PR #1283, opened 2022-10-27
Support for writing params of type List during discovery

Berlioz:


spec/params/params.spec.ts
```diff
@@ -172,6 +195,18 @@ describe("Params value extraction", () => {
...
+  it("can test list params for equality but not < or >", () => {
+    const p1 = params.defineList("LIST");
+    const p2 = params.defineList("ESCAPED_LIST");
+
+    expect(p1.equals(p1).value()).to.be.true;
+    expect(p1.notEquals(p1).value()).to.be.false;
+    expect(p1.equals(p2).value()).to.be.false;
+    expect(p1.notEquals(p2).value()).to.be.true;
+
+    expect(() => p1.greaterThan(p1).value()).to.throw;
```
inlined, line 207: This is why I wish we had done operators in the type system with strong types for each expression type =/

src/params/types.ts
```diff
@@ -159,6 +164,7 @@ type ParamValueType = "string" | "list" | "boolean" | "int" | "float" | "secret"
 type ParamInput<T> =
   | { text: TextInput<T> }
   | { select: SelectInput<T> }
+  | { selectList: SelectListInput }
```
inlined, line 167: I thought this was called MultiSelect in the spec since that name has been common since the days of jQuery.

inlined (commented):
Let's discuss offline whether this should be a JSON array, as presented, or a string + delimiter, as originally designed

spec/params/params.spec.ts
```diff
@@ -92,6 +104,17 @@ describe("Params value extraction", () => {
 
     const stringToBool = params.defineBoolean("A_STRING");
     expect(stringToBool.value()).to.equal(false);
+
+    const listToInt = params.defineInt("LIST");
+    expect(listToInt.value()).to.equal(0);
+  });
+
+  it("falls back on the javascript zero values in case a list param's is unparsable as string[]", () => {
+    const notAllStrings = params.defineList("BAD_LIST");
```
taeold, line 113: no action required - but i can see this being pretty frustrating if we silent fall back to an empty list if the list isn't made up of all string elements :/.

I don't think we expect this to happen in real life since environment variable will almost always be strings anyway?

taeold (commented):
Can we add a changelog entry?

taeold (approved):

PR #1295, opened 2022-10-31
removed duplicate key in package.json typesVersions

YannicEl:
### Description

It looks like my previous PR #1287  and this commit #1267 had a merge conflict that resulted in a duplicate key in the ```typesVersions``` object in ```package.json```

In this PR I deleted the wrong path for the ```v1/https``` key.


YannicEl:
fixed in 4.0.2

PR #1298, opened 2022-11-03
Allow v1 and v2 functions to set the Omit option on a function

Berlioz:
Can't merge until API review

### DIFF
--- a/src/v2/providers/alerts/alerts.ts
+++ b/src/v2/providers/alerts/alerts.ts
@@ -87,7 +87,7 @@ export interface FirebaseAlertOptions extends options.EventHandlerOptions {
   appId?: string;
 
   /**
-   * If true (or resolving to true), do not deploy this function.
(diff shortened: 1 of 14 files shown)


src/v2/providers/alerts/alerts.ts
```diff
@@ -86,6 +86,11 @@ export interface FirebaseAlertOptions extends options.EventHandlerOptions {
   /** Scope the function to trigger on a specific application. */
   appId?: string;
 
+  /**
+   * If true (or resolving to true), do not deploy this function.
```
taeold, line 90: I think we can leave it as just `If true` - since that's how we phrase any other config value.

taeold (approved):

src/v2/providers/alerts/alerts.ts
Berlioz: Done.

PR #1301, opened 2022-11-07
Send the port on which the server was opened back to the parent process

akongara-goog:
### Description

Send the port on which the server was opened back to the parent process, if this process was opened with an (node) IPC channel. If no such channel exists, nothing is changed.

### Code sample

```
      childProcess.on("message", (port: Serializable) => {
        if (typeof port === "number") {
          // port is the port number of the child process.
        } else {
          // handle error
        }
      });
```

### DIFF
--- a/src/bin/firebase-functions.ts
+++ b/src/bin/firebase-functions.ts
@@ -78,4 +78,9 @@ if (process.env.PORT) {
 }
 
 console.log("Serving at port", port);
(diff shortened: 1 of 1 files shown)


taeold:
Can you share some context for this change?

taeold (commented):
Can you share some context for this change?

PR #1305, opened 2022-11-08
Fix schedule function deployment

colerogers:
- Fixes schedule function deployment by always returning a retryConfig from init schedule trigger
- Fixes a bug in task queues where some properties of retryConfig and rateLimits were not resettable if the user supplied any options
- Fixes a bug in testLab where the reset-able endpoint options weren't being added in

### DIFF
--- a/src/v1/function-builder.ts
+++ b/src/v1/function-builder.ts
@@ -274,7 +274,7 @@ export function region(
  *
  * Value must not be null.
  */
-export function runWith(runtimeOptions: RuntimeOptions): FunctionBuilder {
+export function runWith(runtimeOptions: DeploymentOptions): FunctionBuilder {
   if (assertRuntimeOptionsValid(runtimeOptions)) {
     return new FunctionBuilder(runtimeOptions);
   }
@@ -313,7 +313,7 @@ export class FunctionBuilder {
    *
    * Value must not be null.
    */
-  runWith(runtimeOptions: RuntimeOptions): FunctionBuilder {
+  runWith(runtimeOptions: DeploymentOptions): FunctionBuilder {
     if (assertRuntimeOptionsValid(runtimeOptions)) {
       this.options = {
         ...this.options,
--- a/src/v1/function-builder.ts
+++ b/src/v1/function-builder.ts
@@ -274,7 +274,7 @@ export function region(
  *
  * Value must not be null.
  */
-export function runWith(runtimeOptions: DeploymentOptions): FunctionBuilder {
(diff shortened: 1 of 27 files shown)


src/v2/providers/tasks.ts
```diff
@@ -251,8 +251,21 @@ export function onTaskDispatched<Args = any>(
     taskQueueTrigger: initTaskQueueTrigger(options.getGlobalOptions(), opts),
   };
 
-  copyIfPresent(func.__endpoint.taskQueueTrigger, opts, "retryConfig");
-  copyIfPresent(func.__endpoint.taskQueueTrigger, opts, "rateLimits");
+  copyIfPresent(
+    func.__endpoint.taskQueueTrigger.retryConfig,
+    opts.retryConfig || {},
```
taeold, line 256: q: why do we need `|| {}`? If `retryConfig` doesn't exist, then `copyIfPresent` would be a no-op right?

src/v1/function-builder.ts
```diff
@@ -274,7 +274,7 @@ export function region(
  *
  * Value must not be null.
  */
-export function runWith(runtimeOptions: RuntimeOptions): FunctionBuilder {
+export function runWith(runtimeOptions: DeploymentOptions): FunctionBuilder {
```
taeold, line 277: hmm this would mean things like `region` would also be suggested in `runWith` which shouldn't be.

I wonder I made the mistake of setting `preserveExternalChanges` Deployment Option instead of Runtime Option. It doesn't sound right that `preserveExternalChanges` is a runtime option, but then that's the only way I can abuse `runWith` tag 😢

colerogers: Hmm yeah I figured this was a bug since I couldn't figure out a way to use `preserveExternalChanges` without making this change. Also, the v1 docs show `preserveExternalChanges` in the runtime options section - https://firebase.google.com/docs/functions/manage-functions#set_runtime_options

Should we just dump `preserveExternalChanges` into `RuntimeOptions` instead? That way we wouldn't have to fix any docs

src/v2/providers/tasks.ts
colerogers: Ahh thanks for the catch, fixed!

src/v1/function-builder.ts
taeold: let's dump it in RuntimeOptions. I think runWith is already a dumping of options.

taeold (approved):
lgtm - would consider moving the preserveExternalChanges option to runtime options instead.

TheIronDev (approved):

PR #1313, opened 2022-11-16
Allow emitting eventarc channels as params

Berlioz:
As far as interesting PRs go, this ain't it, chief.

Though I have to say it's a little surprising that we define channel as a possible field on the base EventOptions class, despite it only being meaningful to have in one type of Event trigger option? That seems like not how classes should work, unless I'm really missing something about Typescript?


colerogers (approved):

taeold (approved):
> As far as interesting PRs go, this ain't it, chief.Though I have to say it's a little surprising that we define channel as a possible field on the base EventOptions class, despite it only being meaningful to have in one type of Event trigger option? That seems like not how classes should work, unless I'm really missing something about Typescript?

Yeah maybe custom event should've been it's own trigger type, like scheduledTrigger or taskQueueTrigger.

PR #1321, opened 2022-12-02
Fix typo: sections -> seconds.

kaibolay:


egilmorez (approved):
Good catch!  Let's fix this viral copy & paste error, thanks!

PR #1329, opened 2022-12-20
Mark vpc connector as accepting string param type

Berlioz:


taeold (approved):
Can we add a changelog?

PR #1334, opened 2023-01-05
Corrected config-env link to be URL

Mitchman215:
### Description

Fixing a small documentation link error:
The old relative link of `/docs/functions/config-env` doesn't work outside of this repo (for example when using this project as a dependency). Also, markdown syntax didn't seem to be properly rendered on the [official API documentation page](https://firebase.google.com/docs/reference/functions/firebase-functions), so I changed it to just be the URL.


taeold (approved):
🙏🏼

PR #1345, opened 2023-02-01
Patch that fixes emulated v1 callable functions that are in a monorepo project

colerogers:
Surgical patch that will unblock v1 callable users that use monorepo setups in the emulator. Address https://github.com/firebase/firebase-tools/issues/5210

### DIFF
--- a/src/common/providers/https.ts
+++ b/src/common/providers/https.ts
@@ -665,11 +665,14 @@ function wrapOnCallHandler<Req = any, Res = any>(
 
       const context: CallableContext = { rawRequest: req };
 
-      // This code is needed to fix v1 callable functions in the emulator with a monorepo setup.
+      // TODO(colerogers): yank this when we release a breaking change of the CLI that removes
+      // our monkey-patching code referenced below and increases the minimum supported SDK version.
+      //
+      // Note: This code is needed to fix v1 callable functions in the emulator with a monorepo setup.
       // The original monkey-patched code lived in the functionsEmulatorRuntime
       // (link: https://github.com/firebase/firebase-tools/blob/accea7abda3cc9fa6bb91368e4895faf95281c60/src/emulator/functionsEmulatorRuntime.ts#L480)
       // and was not compatible with how monorepos separate out packages (see https://github.com/firebase/firebase-tools/issues/5210).
-      if (process.env.FUNCTIONS_EMULATOR && handler.length === 2) {
+      if (isDebugFeatureEnabled("skipTokenVerification") && handler.length === 2) {
         const authContext = context.rawRequest.header(CALLABLE_AUTH_HEADER);
         if (authContext) {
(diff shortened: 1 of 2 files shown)


TheIronDev (approved):

TheIronDev:
Nonblocking question:

are there tests for this?
Is it a lot of effort to add tests for this?

src/common/providers/https.ts
```diff
@@ -661,6 +664,29 @@ function wrapOnCallHandler<Req = any, Res = any>(
       }
 
       const context: CallableContext = { rawRequest: req };
+
+      // This code is needed to fix v1 callable functions in the emulator with a monorepo setup.
+      // The original monkey-patched code lived in the functionsEmulatorRuntime
+      // (link: https://github.com/firebase/firebase-tools/blob/accea7abda3cc9fa6bb91368e4895faf95281c60/src/emulator/functionsEmulatorRuntime.ts#L480)
+      // and was not compatible with how monorepos separate out packages (see https://github.com/firebase/firebase-tools/issues/5210).
+      if (process.env.FUNCTIONS_EMULATOR && handler.length === 2) {
```
taeold, line 672: nit* I'd prefer re-using `isDebugFeatureEnabled("skipTokenVerification")` logic instead of explicitly checking for existence of a super-special environment variable.

src/common/providers/https.ts
```diff
@@ -661,6 +664,29 @@ function wrapOnCallHandler<Req = any, Res = any>(
       }
 
       const context: CallableContext = { rawRequest: req };
+
+      // This code is needed to fix v1 callable functions in the emulator with a monorepo setup.
+      // The original monkey-patched code lived in the functionsEmulatorRuntime
+      // (link: https://github.com/firebase/firebase-tools/blob/accea7abda3cc9fa6bb91368e4895faf95281c60/src/emulator/functionsEmulatorRuntime.ts#L480)
+      // and was not compatible with how monorepos separate out packages (see https://github.com/firebase/firebase-tools/issues/5210).
```
taeold, line 671: Can we add a TODO with clarification on when we can remove this patch of code? The programmer in me wants to get rid of this code ASAP.

taeold (approved):

src/common/providers/https.ts
colerogers: SGTM, updated

src/common/providers/https.ts
colerogers: Added a TODO explaining when to yank this code

colerogers:
> Nonblocking question:
> 
> are there tests for this? Is it a lot of effort to add tests for this?

Good question - I'm planning on putting a test case for this in `firebase-tools` since integration tests there will use the emulator. But this change in the SDK needs to be released first before we can reference it there.

PR #1374, opened 2023-04-13
Support consumeAppCheckToken option for callable functions

taeold:


src/common/providers/https.ts
```diff
@@ -51,8 +51,24 @@ export interface Request extends express.Request {
  * The interface for AppCheck tokens verified in Callable functions
  */
 export interface AppCheckData {
+  /**
+   * The App ID that App Check token belonged to.
```
weixifan, line 55: Please mention attestation somehow -- maybe:
> The app ID of the Firebase app attested by this App Check token.

src/v2/options.ts
```diff
@@ -200,10 +200,27 @@ export interface GlobalOptions {
...
-   * When false, requests with invalid tokens set event.app to undefiend.
+   * When false, requests with invalid tokens set event.app to undefined.
    */
   enforceAppCheck?: boolean;
 
+  /**
+   * Determines whether Firebase App Check token is consumed on request. Defaults to false.
+   *
+   * @remarks
+   * Set this to true to enable the App Check replay protection feature by consuming App Check token on callable request.
```
weixifan, line 211: nit: consuming *the* App Check token ...

src/v2/options.ts
```diff
@@ -200,10 +200,27 @@ export interface GlobalOptions {
...
+   * Tokens that are found to be already consumed will return app data as alreadyConsumed.
```
weixifan, line 212: Maybe something more explicit like

> will have their `app` context's `alreadyConsumed` field marked as `true`.

src/v2/options.ts
```diff
@@ -200,10 +200,27 @@ export interface GlobalOptions {
...
+   *
+   * Tokens are only considered to be consumed by calling the VerifyAppCheckToken method and setting this value to true;
```
weixifan, line 214: In this case there isn't a method per se, so let's say

> Tokens are only considered to be consumed if it is sent to the App Check backend by setting this option to `true`; other uses of the token do not consume it.

src/v2/providers/https.ts
```diff
@@ -167,6 +167,23 @@ export interface CallableOptions extends HttpsOptions {
    * When false, requests with invalid tokens set event.app to undefiend.
    */
   enforceAppCheck?: boolean;
+
+  /**
+   * Determines whether Firebase App Check token is consumed on request. Defaults to false.
+   *
+   * @remarks
+   * Set this to true to enable the App Check replay protection feature by consuming App Check token on callable request.
+   * Tokens that are found to be already consumed will return app data as alreadyConsumed.
```
weixifan, line 176: Similarly here.

(How come we have two copies of this?)

src/v2/options.ts
weixifan, line 212: We might also want to mention that setting this field to `true` does not affect the behaviour of `enforceAppCheck`.

> This option does not affect the `enforceAppCheck` option. Setting the latter to `true` will cause your Callable Function to auto-respond with a 401 error on invalid tokens. Tokens that are found to be already consumed (but are otherwise valid) will not trigger such an auto-response; instead, in the `event.app` context, the `alreadyConsumed` field will be set to `true`, which your business logic can inspect and react to.

src/common/providers/https.ts
```diff
@@ -51,8 +51,24 @@ export interface Request extends express.Request {
...
   appId: string;
+  /**
+   * Decoded App Check token.
+   */
   token: DecodedAppCheckToken;
+  /**
+   * Indicates if the token has been consumed.
+   *
+   * @remarks
+   * If `false`, App Check has not validated the token, and will be marked as consumed for future use.
```
weixifan, line 66: Let's consolidate our phrasing a bit here:

> If `false`, this is the first time the App Check backend has seen this token. That token will be marked as `already_consumed` for all future invocations of this Callable Function (provided that the `consumeAppCheckToken` option continues to be set to `true`).

taeold: reworded

src/common/providers/https.ts
```diff
@@ -607,23 +628,48 @@ export async function checkAuthToken(
...
+      const decodedToken = unsafeDecodeAppCheckToken(appCheckToken);
       appCheckData = { appId: decodedToken.app_id, token: decodedToken };
+      if (options.consumeAppCheckToken) {
+        appCheckData.alreadyConsumed = false;
+      }
     } else {
-      appCheckData = await getAppCheck(getApp()).verifyToken(appCheck);
+      const appCheck = getAppCheck(getApp());
+      if (options.consumeAppCheckToken) {
+        if (appCheck.verifyToken.length === 1) {
```
lahirumaramba, line 651: I might be wrong here but if `appCheck.verifyToken` is `undefined`, accessing `.length` throws, doesn't it?
Would it be better to do the following instead?

```
if (typeof appCheck.verifyToken === "function") { 
```

lahirumaramba: I just realized that you are checking the number of parameters here. If we are confident that the Admin SDK used always has the appCheck API, please ignore my comment above.

lahirumaramba (approved):

src/common/providers/https.ts
taeold: Think defensive coding is better here!

PR #1380, opened 2023-04-18
Export params package in v1/v2 namespace to improve reference docs.

taeold:


egilmorez (approved):
Verified working in local test, thanks!

PR #1384, opened 2023-04-26
Fix typo on fire alert event type.

taeold:


src/v2/providers/alerts/alerts.ts
```diff
@@ -70,7 +70,7 @@ export type AlertType =
   | "crashlytics.velocity"
   | "crashlytics.newAnrIssue"
   | "billing.planUpdate"
-  | "billing.automatedPlanUpdate"
+  | "billing.planAutomatedUpdate"
```
colerogers, line 73: Is the backend changed to this?

colerogers: Oh it was me! - go/eafxg

colerogers (approved):

src/v2/providers/alerts/alerts.ts
taeold: yep. we actually have it right in the billing.ts.

PR #1423, opened 2023-06-13
Add features to task queue functions

blidd-google:
Since CF3 released support for Task Queue Functions at Google I/O 2022, users have made several feature requests. This PR augments the existing `TaskContext` interface to enable the following features:
* [Headers passed to TaskQueue HTTPS handlers](https://cloud.google.com/tasks/docs/creating-http-target-tasks#handler) are now available to the context/event.
* Users can name their tasks by including an `id` to `TaskOptions` when enqueueing tasks.
* Users can cancel their tasks by providing the task `id` to the new `delete` API.

See https://github.com/firebase/firebase-admin-node/pull/2216 for the sister changes in the Admin SDK.

### DIFF
--- a/src/common/providers/tasks.ts
+++ b/src/common/providers/tasks.ts
@@ -88,22 +88,72 @@ export interface TaskContext {
    * The result of decoding and verifying an ODIC token.
    */
   auth?: AuthData;
+
+  /**
+   * The name of the queue.
+   * Populated via the X-CloudTasks-QueueName header.
+   */
+  queueName: string;
+
+  /**
+   * The "short" name of the task, or, if no name was specified at creation, a unique
+   * system-generated id.
+   * This is the my-task-id value in the complete task name, ie, task_name =
+   * projects/my-project-id/locations/my-location/queues/my-queue-id/tasks/my-task-id.
+   * Populated via the X-CloudTasks-TaskName header.
+   */
+  id: string;
+
+  /**
+   * The number of times this task has been retried.
+   * For the first attempt, this value is 0. This number includes attempts where the task failed
+   * due to 5XX error codes and never reached the execution phase.
+   * Populated via the X-CloudTasks-TaskRetryCount header.
+   */
+  retryCount: number;
+
+  /**
+   * The total number of times that the task has received a response from the handler.
+   * Since Cloud Tasks deletes the task once a successful response has been received, all
+   * previous handler responses were failures. This number does not include failures due to 5XX
+   * error codes.
+   * Populated via the X-CluodTasks-TaskExecutionCount header.
+   */
+  executionCount: number;
+
+  /**
+   * The schedule time of the task, as an RFC 3339 string in UTC time zone
+   * Populated via the X-CloudTasks-TaskETA header, which uses seconds since January 1 1970.
+   */
+  scheduledTime: string;
+
+  /**
+   * The HTTP response code from the previous retry.
+   * Populated via the X-CloudTasks-TaskPreviousResponse header
+   */
+  previousResponse?: number;
+
+  /**
+   * The reason for retrying the task.
+   * Populated via the X-CloudTasks-TaskRetryReason header.
+   */
+  retryReason?: string;
+
+  /**
+   * Raw request headers.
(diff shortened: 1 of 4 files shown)


inlined (approved):

src/common/providers/tasks.ts
```diff
@@ -88,22 +88,72 @@ export interface TaskContext {
...
+   * The total number of times that the task has received a response from the handler.
+   * Since Cloud Tasks deletes the task once a successful response has been received, all
+   * previous handler responses were failures. This number does not include failures due to 5XX
+   * error codes.
+   * Populated via the X-CluodTasks-TaskExecutionCount header.
+   */
+  executionCount: number;
+
+  /**
+   * The schedule time of the task, as an RFC 3339 string in UTC time zone
```
taeold, line 125:
```suggestion
   * The schedule time of the task, as an RFC 3339 string in UTC time zone.
```

src/common/providers/tasks.ts
```diff
@@ -88,22 +88,72 @@ export interface TaskContext {
...
+   * Populated via the X-CloudTasks-TaskETA header, which uses seconds since January 1 1970.
+   */
+  scheduledTime: string;
+
+  /**
+   * The HTTP response code from the previous retry.
+   * Populated via the X-CloudTasks-TaskPreviousResponse header
```
taeold, line 132: Would suggest putting the additional comments like this under `@remark` as done here https://github.com/firebase/firebase-functions/blob/master/src/common/providers/https.ts#L148

Or to put it as a single sentence. FYI everything under `@remark` tag won't show up in the reference docs.

src/common/providers/tasks.ts
```diff
@@ -119,7 +169,26 @@ export function onDispatchHandler<Req = any>(
...
+      for (const [key, value] of Object.entries(req.headers)) {
+        if (!Array.isArray(value)) {
+          headers[key] = value;
+        }
+      }
+
+      const context: TaskContext = {
+        queueName: req.header("X-CloudTasks-QueueName"),
+        id: req.header("X-CloudTasks-TaskName"),
+        retryCount: Number(req.header("X-CloudTasks-TaskRetryCount")),
```
taeold, line 182: nit* Can we make sure this won't below up if the request header isn't missing for some reason (e.g. when we invoke it in the emulator).

Would prefer having safe default behavior in case these headers are missing for whatever reason (maybe log a warning when that happens?)

taeold (approved):
lgtm minor nits.

src/common/providers/tasks.ts
```diff
@@ -88,22 +88,72 @@ export interface TaskContext {
...
+  retryReason?: string;
+
+  /**
+   * Raw request headers.
+   */
+  headers?: Record<string, string>;
 }
 
 /**
  * The request used to call a Task Queue function.
```
egilmorez, line 149: Somewhat out of scope here, but "Task Queue" does not need caps.

src/common/providers/tasks.ts
```diff
@@ -88,22 +88,72 @@ export interface TaskContext {
    * The result of decoding and verifying an ODIC token.
    */
   auth?: AuthData;
+
+  /**
+   * The name of the queue.
+   * Populated via the X-CloudTasks-QueueName header.
```
egilmorez, line 94: Doc strings all look good!

Only question for me is whether "X-CloudTasks-QueueName" should be backticked as a literal.  Or is the idea here that we are suggesting variables, where "QueueName" will actually be a unique name, and X maybe a number?

egilmorez: Yup, per offline, it sounds like we should backtick it for code font as in https://firebase.google.com/docs/functions/callable-reference#request_format_headers and elsewhere.

src/common/providers/tasks.ts
taeold: oh i thought this would blow up:

```
> Number(undefined)
NaN
```

Not sure if I like `NaN` but not as bad as I thought.

taeold (approved):

PR #1429, opened 2023-06-24
Remove param tag that does not exist

manicmaniac:
### Description

The JS doc of `defineSecret()` says it receives `options` argument but that is not true.

So I removed the line including `@param options`.


taeold (approved):
Great find!

Issue #1426, opened 2023-06-16
DataSnapshot.val() throws error for nested child which doesn't exist 


bezysoftware:
### Related issues
This is related to v4 changes here: https://github.com/firebase/firebase-functions/blame/c0fffcc47dee5cca93683f745b619c0d95cf7339/src/common/providers/database.ts#L123

### [REQUIRED] Version info
**node:** v16.15.0
**firebase-functions:** 4.4.1
**firebase-tools:** 12.4.0
**firebase-admin:** 10.2.0

### [REQUIRED] Test case

```
import * as functions from 'firebase-functions';

export let test = functions
    .database
    .ref('/test/{id}')
    .onWrite((event, context) => console.log(event.after.child('nested/property').val()));
```

### [REQUIRED] Steps to reproduce

Insert any value to realtime database at `/test/123/whatever`

### [REQUIRED] Expected behavior

Calling `event.after.child('nested/property').val())` should return `null` (as it does prior to v4)

### [REQUIRED] Actual behavior

```
TypeError: Cannot read properties of null (reading 'property') at DataSnapshot.val (/workspace/node_modules/firebase-functions/lib/common/providers/database.js:104:32)
```

### Were you able to successfully deploy your functions?
Yes, this is a runtime error.

### Footnote
Since the v4 rewrote the `exists()` as a check to the result of `val()` it is also affected. And it's really not clear to me how else to check if the nested property exists.

blidd-google:
Hi @bezysoftware, thanks for finding this issue and bringing it to our attention. PR is out to fix the bug at https://github.com/firebase/firebase-functions/pull/1432/.

bezysoftware:
@blidd-google the PR is still not merged, is there something pending?

inlined:
Sorry for the delay in release. The latest version of the SDK has this fix.

PR #1432, opened 2023-06-30
Fixes access on deeply nested, nonexistent property

blidd-google:
When attempting to access a deeply nested property on a RTDB node, we are currently throwing an error whereas the expected behavior is to return `null`. This PR corrects the behavior.

Fixes https://github.com/firebase/firebase-functions/issues/1426.


timothegenzmer:
We have recently upgraded to 4.4.1 and are facing the exact same issue after updating.
I would love to see this merged and releases.
This is currently preventing us from updating

colerogers (approved):

bezysoftware:
Any updates on this? It's preventing us to update to latest version

PR #1446, opened 2023-07-18
Add trace property to logs created via logger.write()

yblok13:
### Description

Calling `logger.write()` does not automatically add the traceId in logs like calling `logger.log()`, `logger.error()`, etc.
Since the latter functions call `logger.write()` anyways, I moved the snippet of code setting the traceId to the `write` function.

### Code sample

```
const { logger } = require("firebase-functions/v2/logger");

logger.write("TEST LOG", { severity: "NOTICE" });
// Prior to change, traceId was not set.
```


jiangbo0216:
same as here, Can we expedite the processing?

jiangbo0216:
What is the reason behind this modification?

yblok13:
With console, you can only use debug, log, warning, and error. Logger uses much more fine tuned verbose levels. However, using those verbose levels does not attach the stack trace in Google Cloud Logging. This pull request adds the stack trace to all verbose levels.

lox:
It also makes it impossible to get trace information on critical+ log levels.

taeold:
Very reasonable change. thanks for the contribution

taeold (approved):

taeold (approved):

PR #1458, opened 2023-08-23
supporting recaptcha verdict for auth blocking functions

Xiaoshouzi-gh:
### Description
Adding recaptcha support for auth blocking function beforeCreate and beforeSignIn

### Code sample
```
exports.checkrecaptcha = beforeUserSignedIn(async (event) => {
  if (event.additionalUserInfo.recaptchaScore > 0.6) {
    return {
      recaptchaPassed: true
    }
  } else {
    return {
      recaptchaPassed: false
    }
  }
```

### DIFF
--- a/spec/common/providers/identity.spec.ts
+++ b/spec/common/providers/identity.spec.ts
@@ -26,6 +26,7 @@ import * as identity from "../../../src/common/providers/identity";
 
 const EVENT = "EVENT_TYPE";
 const now = new Date();
+const TEST_NAME = "John Doe";
 
 describe("identity", () => {
   describe("userRecordConstructor", () => {
@@ -232,14 +233,14 @@ describe("identity", () => {
   describe("parseProviderData", () => {
     const decodedUserInfo = {
       provider_id: "google.com",
-      display_name: "John Doe",
+      display_name: TEST_NAME,
       photo_url: "https://lh3.googleusercontent.com/1234567890/photo.jpg",
       uid: "1234567890",
       email: "user@gmail.com",
     };
     const userInfo = {
       providerId: "google.com",
-      displayName: "John Doe",
+      displayName: TEST_NAME,
       photoURL: "https://lh3.googleusercontent.com/1234567890/photo.jpg",
       uid: "1234567890",
       email: "user@gmail.com",
@@ -340,12 +341,12 @@ describe("identity", () => {
       uid: "abcdefghijklmnopqrstuvwxyz",
       email: "user@gmail.com",
       email_verified: true,
-      display_name: "John Doe",
+      display_name: TEST_NAME,
       phone_number: "+11234567890",
       provider_data: [
         {
           provider_id: "google.com",
-          display_name: "John Doe",
+          display_name: TEST_NAME,
           photo_url: "https://lh3.googleusercontent.com/1234567890/photo.jpg",
           email: "user@gmail.com",
           uid: "1234567890",
@@ -366,7 +367,7 @@ describe("identity", () => {
           provider_id: "password",
           email: "user@gmail.com",
           uid: "user@gmail.com",
-          display_name: "John Doe",
+          display_name: TEST_NAME,
         },
       ],
       password_hash: "passwordHash",
@@ -407,11 +408,11 @@ describe("identity", () => {
       phoneNumber: "+11234567890",
       emailVerified: true,
       disabled: false,
-      displayName: "John Doe",
+      displayName: TEST_NAME,
       providerData: [
         {
           providerId: "google.com",
-          displayName: "John Doe",
+          displayName: TEST_NAME,
           photoURL: "https://lh3.googleusercontent.com/1234567890/photo.jpg",
           email: "user@gmail.com",
           uid: "1234567890",
@@ -435,7 +436,7 @@ describe("identity", () => {
         },
         {
           providerId: "password",
-          displayName: "John Doe",
+          displayName: TEST_NAME,
           photoURL: undefined,
           email: "user@gmail.com",
           uid: "user@gmail.com",
@@ -489,8 +490,9 @@ describe("identity", () => {
   });
 
   describe("parseAuthEventContext", () => {
+    const TEST_RECAPTCHA_SCORE = 0.9;
     const rawUserInfo = {
-      name: "John Doe",
+      name: TEST_NAME,
       granted_scopes:
         "openid https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/userinfo.profile",
       id: "123456789",
@@ -516,6 +518,7 @@ describe("identity", () => {
         user_agent: "USER_AGENT",
         locale: "en",
         raw_user_info: JSON.stringify(rawUserInfo),
+        recaptcha_score: TEST_RECAPTCHA_SCORE,
       };
       const context = {
         locale: "en",
@@ -534,6 +537,7 @@ describe("identity", () => {
           profile: rawUserInfo,
           username: undefined,
           isNewUser: false,
+          recaptchaScore: TEST_RECAPTCHA_SCORE,
         },
         credential: null,
         params: {},
@@ -563,6 +567,7 @@ describe("identity", () => {
         oauth_refresh_token: "REFRESH_TOKEN",
         oauth_token_secret: "OAUTH_TOKEN_SECRET",
         oauth_expires_in: 3600,
+        recaptcha_score: TEST_RECAPTCHA_SCORE,
       };
       const context = {
         locale: "en",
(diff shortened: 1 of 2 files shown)


Xiaoshouzi-gh:
cc @prameshj @eromney

src/common/providers/identity.ts
```diff
@@ -640,9 +647,38 @@ function parseAdditionalUserInfo(decodedJWT: DecodedPayload): AdditionalUserInfo
...
     username,
     isNewUser: decodedJWT.event_type === "beforeCreate" ? true : false,
+    recaptchaScore: decodedJWT.recaptcha_score,
   };
 }
 
+/** Helper to generate payload to GCIP from client request.
+ * @internal
+ */
+export function generateRequestPayload(
```
blidd-google, line 657: *nit: I think the name "RequestPayload" also makes sense since the payload will be sent to GCIP, and honestly both make sense to me, but for the sake of consistency with existing naming conventions in the SDK let's call the payload a "ResponsePayload".

src/common/providers/identity.ts
```diff
@@ -640,9 +647,38 @@ function parseAdditionalUserInfo(decodedJWT: DecodedPayload): AdditionalUserInfo
     profile,
     username,
     isNewUser: decodedJWT.event_type === "beforeCreate" ? true : false,
+    recaptchaScore: decodedJWT.recaptcha_score,
   };
 }
 
+/** Helper to generate payload to GCIP from client request.
```
blidd-google, line 654: add newline after /**

src/common/providers/identity.ts
```diff
@@ -640,9 +647,38 @@ function parseAdditionalUserInfo(decodedJWT: DecodedPayload): AdditionalUserInfo
...
+ * @internal
+ */
+export function generateRequestPayload(
+  authResponse?: BeforeCreateResponse | BeforeSignInResponse
+): any {
```
blidd-google, line 659: Do we know exactly the shape of the data GCIP expects? If we do then we could statically type the return value here instead of using `any`, otherwise not really a huge deal.

blidd-google (commented):
Just a few minor things, but looks great and thanks for your work on this! Also, please add an entry to the changelog documenting the changes. Will leave this in a commented state until we get the go-ahead from the API council.

src/common/providers/identity.ts
prameshj, line 654: "Helper to generate a response from the blocking  function to the Firebase Auth backend"?

src/common/providers/identity.ts
prameshj: +1 to calling it ResponsePayload.

src/common/providers/identity.ts
```diff
@@ -338,8 +339,13 @@ export interface AuthBlockingEvent extends AuthEventContext {
   data: AuthUserRecord;
 }
 
+/** The base handler response type for beforeCreate and beforeSignIn blocking events*/
+export interface BlockingFunctionResponse {
+  recaptchaPassed?: boolean;
```
prameshj, line 344: Since all fields are optional in BeforeCreateResponse and BeforeSignInResponse, should we put them all here? I understand this will be a much larger refactor, so this change is very optional.

src/common/providers/identity.ts
prameshj: +1, let's avoid 'any' if possible. I guess BeforeSignInResponse is the type to use? (Or if we consolidate all fields into BlockingFunctionsResponse we could use that, in future)

spec/common/providers/identity.spec.ts
```diff
@@ -762,4 +770,38 @@ describe("identity", () => {
       );
     });
   });
+
+  describe("generateRequestPayload", () => {
+    const DISPLAY_NAME_FILED = "displayName";
```
prameshj, line 775: nit: typo "DISPLAY_NAME_FIELD"

spec/common/providers/identity.spec.ts
```diff
@@ -762,4 +770,38 @@ describe("identity", () => {
...
+    const TEST_RESPONSE = {
+      displayName: TEST_NAME,
+      recaptchaPassed: false,
+    } as identity.BeforeCreateResponse;
+
+    const EXPECT_PAYLOAD = {
+      userRecord: { displayName: TEST_NAME, updateMask: DISPLAY_NAME_FILED },
+      recaptchaPassed: false,
```
prameshj, line 783: nit: can we try a test case where recaptchaPassed is true?

src/common/providers/identity.ts
prameshj: Actually, given this is a public interface, I think we should keep it as you have it here. Maybe we can have an internal/hidden interface which consolidates all the fields, mainly as return type for generatePayloadRequest.

src/common/providers/identity.ts
Xiaoshouzi-gh: BeforeSignInResponse is the interface for blocking function returns for developers (e.g. developers can return displayName, emailVeried etc field when they implemented the blocking function).  The requestPayload(responsePayload) here is (will be) a new different interface that follows GCIP backend proto.  They are very similar but nested a bit differently.  This is also the reason why the naming is kind of confusing, it is generated from the blocking function response as a request to GCIP backend.

src/common/providers/identity.ts
Xiaoshouzi-gh: The reason why separating this out is this new interface will be used in the beforeEmailSent trigger.  And in the beforeEmailSent response, `recaptchaPassed` is the only field we support.

src/common/providers/identity.ts
Xiaoshouzi-gh: Updated

src/common/providers/identity.ts
Xiaoshouzi-gh: Done.

spec/common/providers/identity.spec.ts
Xiaoshouzi-gh: Thanks for catching this. updated.

spec/common/providers/identity.spec.ts
Xiaoshouzi-gh: added

src/common/providers/identity.ts
Xiaoshouzi-gh: Added a new interface for this.

spec/common/providers/identity.spec.ts
```diff
@@ -762,4 +770,52 @@ describe("identity", () => {
       );
     });
   });
+
+  describe("generateRequestPayload", () => {
```
prameshj, line 774: "generateResponsePayload"

src/common/providers/identity.ts
prameshj: makes sense.. Since the fields are all optional, technically, we could use the same Response for beforeEmailSent trigger. But what you have is cleaner.

src/common/providers/identity.ts
```diff
@@ -423,9 +429,21 @@ export interface DecodedPayload {
   oauth_refresh_token?: string;
   oauth_token_secret?: string;
   oauth_expires_in?: number;
+  recaptcha_score?: number;
   [key: string]: any;
 }
 
+/** @internal */
```
prameshj, line 436: nit: add a comment on why this is used.

"Internal definition to include all the fields that can be sent as a response from the blocking function to the backend. This is added mainly to have a type definition for 'generateResponsePayload'".

prameshj (approved):
Thanks for making the changes!

src/common/providers/identity.ts
```diff
@@ -673,7 +678,8 @@ export function generateResponsePayload(
     return {};
   }
 
-  const { recaptchaPassed, ...formattedAuthResponse } = authResponse;
+  const { recaptchaActionOverride: recaptchaActionOverride, ...formattedAuthResponse } =
```
blidd-google, line 682: Am I missing something with the destructuring assignment for `recaptchaActionOverride`, or can we just do `const { recaptchaActionOverride, ...formatedAuthResponse } = authResponse;` instead?

blidd-google (approved):
LGTM besides one very minor nit

src/common/providers/identity.ts
```diff
@@ -338,13 +339,19 @@ export interface AuthBlockingEvent extends AuthEventContext {
   data: AuthUserRecord;
 }
 
+/**
+ * The reCACPTCHA action options.
+ */
+export type RecatpchaActionOptions = "ALLOW" | "BLOCK";
```
prameshj, line 345: nit: "Recaptcha" typo. Also in the comment.

prameshj (approved):

PR #1477, opened 2023-10-27
fixes #1474: remove enforceAppCheck from HttpsOptions

### LINKED ISSUE #1474: v2 onRequest does not enforce AppCheck
### Related issues

https://github.com/firebase/firebase-functions/issues/1377

### [REQUIRED] Version info

"firebase-admin": 
"firebase-functions": 

**node:** v18.15.0

**firebase-functions:** "4.4.1",

**firebase-tools:** 12.5.4

**firebase-admin:** "11.11.0",

### [REQUIRED] Test case

```ts
import {onRequest} from 'firebase-functions/v2/https';

onRequest({enforceAppCheck: true}, (req, res) => res.status(200).end());
```

Set up a cloud function using v2 onRequest, with `enforceAppCheck: true`

Calling the function (locally or when deployed) with the HTTP request should be blocked if no app check token is available

Function works fine, same as https://github.com/firebase/firebase-functions/issues/1377 without any token

yes

AmitMY:
fixes #1474

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,3 @@
 - Fix retry in event triggered functions. (#1463)
 - Expose retry configuration in v2 RTDB trigger (#1588)
+- Remove invalid `enforceAppCheck` option for v2 onRequest trigger (#1477)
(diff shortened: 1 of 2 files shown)


taeold (approved):
Thanks!

PR #1518, opened 2024-01-31
Make storage bucket parameterizeable. Create const for bucket picker

inlined:
API spec basically said to parameterize everything so I don't know why this was overlooked. Also, I've added a const BUCKET_PICKER so customers don't have to say `{ input: { resource: { type: "storage.googleapis.com/Bucket" } } }` and can instead say `{ input: BUCKET_PICKER }`. That const might need to go through API council review though.

### DIFF
--- a/src/params/types.ts
+++ b/src/params/types.ts
@@ -21,7 +21,6 @@
 // SOFTWARE.
 
 import * as logger from "../logger";
-import { Resource } from "../v1";
 
 /*
  * A CEL expression which can be evaluated during function deployment, and
--- a/src/params/types.ts
+++ b/src/params/types.ts
@@ -231,7 +231,7 @@ export interface SelectInput<T = unknown> {
 export interface MultiSelectInput {
   multiSelect: {
     options: Array<SelectOptions<string>>;
-  }
+  };
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/params/types.ts
```diff
@@ -205,10 +205,16 @@ export interface TextInput<T = unknown> {
...
+    type: "storage.googleapis.com/Bucket";
   };
 }
 
+export const BUCKET_PICKER: ResourceInput = {
+  resource: {
+    type: "storage.googleapis.com/Bucket",
+  },
+};
+
```
mbleigh, line 217: Is the helper constant even needed given VS Code will autocomplete the enum string for the resource type?

If so, might prefer naming of `SELECT_STORAGE_BUCKET` over `BUCKET_PICKER`.

mbleigh (approved):

src/params/types.ts
inlined: I think it is much less cringe to type a constant rather than a complex structure that has only one possible configuration. There's precedence here with the ServerValues consts/functions in RTDB/Firestore. Switched to your preferred name though.

inlined:
Given the additional functions added, there is no question that this now needs API council review. The review is scheduled for next week.

PR #1519, opened 2024-02-01
Add new 2nd gen Firestore auth context triggers

blidd-google:
Auth context support has only been available for RTDB 1st gen functions — until now! The Firestore and Eventarc teams have recently added support for Firestore Cloud Events to include event actor information, adding four new event types. These changes enable us to offer four new 2nd gen Firestore trigger types whose events will additionally contain the auth context along with the existing event data:

- `onDocumentWrittenWithAuthContext`
- `onDocumentUpdatedWithAuthContext`
- `onDocumentCreatedWithAuthContext`
- `onDocumentDeletedWithAuthContext`

The API is very similar to the existing 2nd gen Firestore triggers API; however, users should take special care to avoid event loss when migrating from the original 2nd gen Firestore trigger to the auth-context trigger. See the [migration guide](link) for more details about best practices. (TODO: link migration guide when available)

### DIFF
--- a/spec/v2/providers/firestore.spec.ts
+++ b/spec/v2/providers/firestore.spec.ts
@@ -558,6 +558,23 @@ describe("firestore", () => {
       expect(func.run(true as any)).to.eq(2);
       expect(func.__endpoint).to.deep.eq(expectedEp);
     });
+
+    it("calls init function", async () => {
+      const event: firestore.RawFirestoreEvent = {
+        ...eventBase,
+        datacontenttype: "application/json",
+        data: {
+          oldValue: null,
+          value: null,
+        },
+      };
+
+      let hello;
+      onInit(() => (hello = "world"));
+      expect(hello).to.be.undefined;
+      await firestore.onDocumentWrittenWithAuthContext("path", () => null)(event);
+      expect(hello).to.equal("world");
+    });
   });
 
   describe("onDocumentCreatedWithAuthContext", () => {
@@ -604,6 +621,23 @@ describe("firestore", () => {
 
       expect(func.run(true as any)).to.eq(2);
       expect(func.__endpoint).to.deep.eq(expectedEp);
+    });
+
+    it("calls init function", async () => {
+      const event: firestore.RawFirestoreEvent = {
+        ...eventBase,
+        datacontenttype: "application/json",
+        data: {
+          oldValue: null,
+          value: null,
+        },
+      };
+
+      let hello;
+      onInit(() => (hello = "world"));
+      expect(hello).to.be.undefined;
+      await firestore.onDocumentCreatedWithAuthContext("path", () => null)(event);
+      expect(hello).to.equal("world");
     });
   });
 
@@ -652,6 +686,23 @@ describe("firestore", () => {
       expect(func.run(true as any)).to.eq(2);
       expect(func.__endpoint).to.deep.eq(expectedEp);
     });
+
+    it("calls init function", async () => {
+      const event: firestore.RawFirestoreEvent = {
+        ...eventBase,
+        datacontenttype: "application/json",
+        data: {
+          oldValue: null,
+          value: null,
+        },
+      };
+
+      let hello;
+      onInit(() => (hello = "world"));
+      expect(hello).to.be.undefined;
+      await firestore.onDocumentUpdatedWithAuthContext("path", () => null)(event);
+      expect(hello).to.equal("world");
+    });
   });
 
   describe("onDocumentDeletedWithAuthContext", () => {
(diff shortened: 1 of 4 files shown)


src/v2/providers/firestore.ts
```diff
@@ -234,6 +345,48 @@ export function onDocumentUpdated<Document extends string>(
...
+export function onDocumentUpdatedWithAuthContext<Document extends string>(
+  opts: DocumentOptions<Document>,
+  handler: (
+    event: FirestoreEvent<Change<QueryDocumentSnapshot> | undefined, ParamsOf<Document>>
+  ) => any | Promise<any>
+): CloudFunction<FirestoreEvent<Change<QueryDocumentSnapshot> | undefined, ParamsOf<Document>>>;
+
+/**
+ * Event handler which triggers when a document is updated in Firestore.
+ *
```
exaby73, line 377: The following doc was not added here:

```txt
This trigger will also provide the authentication context of the principal who triggered the event.
```

src/v2/providers/firestore.ts
```diff
@@ -153,6 +177,50 @@ export function onDocumentWritten<Document extends string>(
   return onChangedOperation(writtenEventType, documentOrOpts, handler);
 }
 
+/**
+ * Event handler which triggers when a document is created, updated, or deleted in Firestore.
```
egilmorez, line 181: This and most other instances of "which" in these comments is better as "that" (more restrictive, rather than additive). 

https://prowritingaid.com/art/438/-Which--or--That-%3A-Know-When-to-Use-Each.aspx

src/v2/providers/firestore.ts
```diff
@@ -153,6 +177,50 @@ export function onDocumentWritten<Document extends string>(
...
+ * This trigger will also provide the authentication context of the principal who triggered the event.
```
egilmorez, line 182: Suggest "also provides"

Avoid "will" wherever you can.  Please check this globally in this PR, thanks!

egilmorez (commented):
Couple of things for you, thanks!

src/v2/providers/firestore.ts
```diff
@@ -95,6 +115,10 @@ export interface FirestoreEvent<T, Params = Record<string, string>> extends Clou
   namespace: string;
   /** The document path */
   document: string;
+  /** The type of principal that triggered the event */
+  authType?: AuthType;
```
inlined, line 119: These will always be populated in the new event type, right? If so, we should create a separate event type where the fields are omitted in the previous event handlers and guaranteed to exist in the new ones.

src/v2/providers/firestore.ts
```diff
@@ -83,6 +101,8 @@ export type DocumentSnapshot = firestore.DocumentSnapshot;
 /** A Firestore QueryDocumentSnapshot */
 export type QueryDocumentSnapshot = firestore.QueryDocumentSnapshot;
 
+type AuthType = "service_account" | "api_key" | "system" | "unauthenticated" | "unknown";
```
inlined, line 104: Can we document when "unknown" will happen and if there are any assumptions the dev can make (e.g. that it's from an admin context)?

inlined: Also wouldn't hurt to export.

inlined (commented):
Where's the withInit calls? Is this based on an old CL?

src/v2/providers/firestore.ts
blidd-google: @exaby73 actually made the same suggestion when implementing the triggers in the Python SDK, but my reasoning for using the same event type is that we had discussed plans for aliasing `onDocumentX` to `onDocumentXWithAuthContext` in the next major release of the SDK, which may pollute the namespace a bit and cause confusion; a user writing a `onDocumentX` function will need to use the `FirestoreEventWithAuthContext` interface as opposed to just a single `FirestoreEvent` interface. The additional `authId` field is also optional and not guaranteed to exist in the delivered event.

But I also see the argument for having a separate event type. WDYT? Happy to make the changes and coordinate with @exaby73 to restore his old changes if we think the correct approach.

src/v2/providers/firestore.ts
blidd-google: Yep the CL is a few months old, quite a few things have changed in the meantime!

src/v2/providers/firestore.ts
inlined: I'm pretty strongly in favor of fields being non-optional when they will always be provided and omitted when they will never be provided (in TypeScript at least. I'm not the expert for Python)

spec/v2/providers/firestore.spec.ts
```diff
@@ -511,6 +513,194 @@ describe("firestore", () => {
     });
   });
 
+  describe("onDocumentWrittenWithAuthContext", () => {
```
inlined, line 516: Missing `it calls init function` cases everywhere. Because v2 doesn't have a shared `makeCloudFunction` library, I have tests for each event type.

colerogers (commented):
lgtm after addressing @inlined's concerns

src/v2/providers/firestore.ts
```diff
@@ -156,7 +196,58 @@ export function onDocumentWritten<Document extends string>(
...
+export function onDocumentWrittenWithAuthContext<Document extends string>(
+  documentOrOpts: Document | DocumentOptions<Document>,
+  handler: (
+    event: FirestoreAuthEvent<Change<DocumentSnapshot> | undefined, ParamsOf<Document>>
+  ) => any | Promise<any>
+): CloudFunction<FirestoreAuthEvent<Change<DocumentSnapshot> | undefined, ParamsOf<Document>>> {
+  // const fn = (
+  //   event: FirestoreEvent<Change<QueryDocumentSnapshot> | undefined, ParamsOf<Document>> & {
+  //     foo: string;
+  //   }
```
inlined, line 242: Remove dead code

inlined (approved):

wilpar:
migration guide link in the original post is 404.

PR #1531, opened 2024-03-02
Add startup lifecycle hook

inlined:
Adds a new function that can be used at global scope to perform initialization in production that won't be called during discovery. In this function, params will be guaranteed to have actual values.

Usage:

```
const apiKey = params.defineSecret("API_KEY");

onInit(() => {
  someSDK.init(apiKey.value);
});
```


src/common/onInit.ts
```diff
@@ -0,0 +1,39 @@
+import * as logger from "../logger";
+
+let initCallback: (() => unknown) | null = null;
+let didInit = false;
```
blidd-google, line 4: Do we leave globals as camel-cased or in CAPS?

blidd-google (approved):
LGTM

src/common/onInit.ts
inlined: Typically CAPS is for consts, which these aren't

PR #1533, opened 2024-03-04
FIX: Add wrapTraceContext to onChangedOperation to enable traceId on onDocumentUpdated

JonatanSalas:
### Description

When using firebase-functions v2 `traceId` is missing for Firestore triggers like `onDocumentUpdated`. 

After a deep dive into the repository's code I've found that there's a difference. 

`onDocumentCreated` relies in the method `onOperation` which wraps the event handler with the function `wrapTraceContext` that injects the `trace` information.

Screenshot 2024-03-04 at 14 14 44

On the other side, `onDocumentUpdated` relies in the method `onChangedOperation` which executes the event handler without using the `wrapTraceContext` function. **I don't know if this is intentional, but this is a problem of those who are heavy users of firebase functions under the v2 because in previous version the `traceId` was available**. 

Screenshot 2024-03-04 at 14 14 59

I've adapted the code to make `onChangedOperation` use the `wrapTraceContext` function. With that line I think we can solve the issue of missing `traceId` in `onDocumentUpdated`.

### DIFF
--- a/src/v2/providers/firestore.ts
+++ b/src/v2/providers/firestore.ts
@@ -473,7 +473,7 @@ export function onChangedOperation<Document extends string>(
     const event = raw as RawFirestoreEvent;
     const params = makeParams(event.document, documentPattern) as unknown as ParamsOf<Document>;
     const firestoreEvent = makeChangedFirestoreEvent(event, params);
-    return handler(firestoreEvent);
+    return wrapTraceContext(handler)(firestoreEvent);
   };
 
   func.run = handler;


exaby73:
Hey @JonatanSalas. Can you fix the conflicts?

JonatanSalas:
@exaby73 sure!

exaby73:
Hey @JonatanSalas. There still seems to be conflicts on this PR

taeold:
This was fixed in https://github.com/firebase/firebase-functions/pull/1531!

Thank you for your help!

PR #1534, opened 2024-03-04
FEAT: Add logger compat module for v2 restoring fully v1 behavior when using standard console

JonatanSalas:
### Description

In firebase-functions v1 when using the standard `console` you'd visualize logs under cloud run with a `traceId`. 

In v2 that behavior has gone, there's a module to patch the `console` but that method only makes console work again by adjusting the way it logs the user inserted messages. 

Screenshot 2024-03-04 at 14 30 04

If we want to have the trace attached to each log emitted by a function we need to use the firebase logger module which is provided by firebase functions. 

Since the **logger compat** module doesn't restore the behavior under v1, I've made an adaptation to ensure it works as expected and keeps the `traceId`. 

**The code is a draft, I've added a `v2` folder inside the `logger` folder which I think is wrong but I've done so we can start discussing if this adjustment is possible and where it should go.**

### DIFF
--- a/src/logger/v2/compat.ts
+++ b/src/logger/v2/compat.ts
@@ -1,0 +1,44 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017 Firebase
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


taeold:
@JonatanSalas Hi Jonantan. I don't think this will fix the issue as you reported.

IIUC, `compat` module must be explicitly imported in order for console patching to work. And even if it is patched, you still won't be receiving `traceIds` in your 2nd Gen function; this is something that changed in Google Cloud Functions's 2nd Gen execution environment.

I recommend that you use Firebase Function SDK's logger module which automatically injects traceId based on request context.

PR #1536, opened 2024-03-08
CORS origins of an array of one are made a scalar.

inlined:
The access-control-allowed-origins CORS header only allows a single origin or "*" as its response. To support multiple origins, the cors middleware makes this header dynamic based on the origin header of the request when the middleware is configured with anything but a single string.

To help avoid a few edge cases customers may encounter, we can unwrap an array of one element into a scalar to encourage the cors middleware to make the access-control-allowed-origin header static.

As a very minor performance boost, this change also instantiates the cors middleware once and uses it on all requests rather than constructing it dynamically within a request.

### DIFF
--- a/src/v2/providers/https.ts
+++ b/src/v2/providers/https.ts
@@ -265,12 +265,19 @@ export function onRequest(
       // Respect `cors: false` to turn off cors even if debug feature is enabled.
       origin = opts.cors === false ? false : true;
     }
+    // Arrays cause the access-control-allow-origin header to be dynamic based
+    // on the origin header of the request. If there is only one element in the
+    // array, this is unnecessary.
+    if (Array.isArray(origin) && origin.length === 1) {
+      origin = origin[1];
+    }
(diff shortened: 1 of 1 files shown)


colerogers (approved):

src/v2/providers/https.ts
```diff
@@ -265,12 +265,19 @@ export function onRequest(
       // Respect `cors: false` to turn off cors even if debug feature is enabled.
       origin = opts.cors === false ? false : true;
     }
+    // Arrays cause the access-control-allow-origin header to be dynamic based
+    // on the origin header of the request. If there is only one element in the
+    // array, this is unnecessary.
+    if (Array.isArray(origin) && origin.length === 1) {
+      origin = origin[1];
```
BenaymS, line 272: Is this a bug... shouldn't this either be `origin = origin[0]` or `origin = undefined`?

inlined: 🤬 Will fix this week

PR #1543, opened 2024-03-26
Remove unnecessary strict dep on node-fetch

### LINKED ISSUE #1542: Please move `node-fetch` as a `devDependency`
### [REQUIRED] Version info

**node:**

v18.18.0

**firebase-functions:**

v4.8.1

**firebase-tools:**

13.5.2

**firebase-admin:**

11.11.1

### [REQUIRED] Test case

We are trying to upgrade `node-fetch` to v3.X.X, but the `firebase-functions` package has pinned as a dependency v2.X.X.
As far as I can see, you only use `node-fetch` for tests.
Could you please move the `node-fetch` dependency as a `devDependency` so it does not prevent packages that depend on `firebase-functions` from upgrading.

Add `firebase-functions` as a dependency to a project which depends on `node-fetch` v3.X.X

The project cannot use `node-fetch` v3.X.X

If you put `node-fetch` v2.X.X  in `devDependencies` of `firebase-functions`, it will still be used in tests, but it will not prevent other projects from upgrading.

inlined:
Fixes #1542

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,7 +201,6 @@
     "@types/express": "4.17.3",
     "cors": "^2.8.5",
     "express": "^4.17.1",
-    "node-fetch": "^2.6.7",
     "protobufjs": "^7.2.2"
   },
   "devDependencies": {


blidd-google (approved):
Strange...

PR #1544, opened 2024-03-26
Fix bug introduced in #1536

inlined:
Accidentally 1-indexed into an array.

Fixes regression from #1536


colerogers (commented):
😅

colerogers (approved):

bjester:
Second followup here https://github.com/firebase/firebase-functions/pull/1564

PR #1548, opened 2024-04-03
Ugly type safety

inlined:
Here's the best I could do to make everything type safe without adding any extra `any`. Personally, I'm not sure if I care to have this be so complicated or if we should have more `any`s as well. Just if you do choose to shirk type safety, please make sure fields that shouldn't be defined in the non-auth event types don't have the extra fields.

Should possibly extend test for make*Event to test fields other than data.


blidd-google (approved):
made some changes

PR #1594, opened 2024-08-16
Change default entrypoint of the firebase-functions package to v2 instead of v1

taeold:
On the same note of `gcloud functions` commands defaulting to 2nd Gen functions, Firebase Functions SDK will change default entrypoint of the SDK to point to the v2 namespace in the next major release.

Before:
```js

// This is a v1 callable function
exports.v1 = functions.https.onCall((data, context) => {
  // ...
});
```

After:
```js

// This is a v2 callable function
exports.v2 = functions.https.onCall((req) => {
  // ...
});

// Preferred style 
const { onCall } = require('firebase-functions/https');

// This is a v2 callable function
exports.v2 = onCall((req) => {
  // ...
});
```


package.json
```diff
@@ -22,16 +22,34 @@
...
   },
-  "types": "lib/v1/index.d.ts",
+  "types": "lib/v2/index.d.ts",
   "exports": {
-    ".": "./lib/v1/index.js",
+    ".": "./lib/v2/index.js",
     "./logger/compat": "./lib/logger/compat.js",
     "./logger": "./lib/logger/index.js",
     "./params": "./lib/params/index.js",
+    "./core": "./lib/v2/core.js",
```
inlined, line 35: Wow! I can’t believe we missed these! Can we collocate v1 vs v2?

inlined (approved):
Way more files changed than I expected. Thanks!

package.json
taeold: can you clarify what you mean?

PR #1607, opened 2024-09-09
Expose functions.config in the v2 namespace to avoid breaking the Functions Emulator

taeold:
Unforutnately, Functions Emulator monkey-patches the `functions.config()` method. 

Not exposing the `config()` method in the default namespace crashes the functions emulator 😢 .

https://github.com/firebase/firebase-tools/blob/0e2ab54c522d40ec9fe874e108332427e4886e0a/src/emulator/functionsEmulatorRuntime.ts#L717

In the near future, we'll make monkey-patching robust in the emulator to not crash when `functions.config()` method is missing


egilmorez (approved):
LG to this primate.  Thanks!

Issue #1078, opened 2022-04-15
Update Express


Brian-McBride:
https://github.com/firebase/firebase-functions/blob/9a5b23f855f96cfd69e37480b63636cfbe3e39f9/package.json#L169-L171

Express is updated, for some reason the `@types` is current in this lib, but the actual `express` is a couple of versions behind. 

Can we get this bumped up? Using Firebase is causing a lot of dependency conflicts with other packages that are keeping up.

adrian-meditect:
There is currently a vulnerability in this express version caused by a dependency on qs. Please, upgrade asap 🙏 .

Here are the dependency trail details in yarn.lock format:

```
express@^4.17.1:
  version "4.17.2"
  resolved "https://registry.yarnpkg.com/express/-/express-4.17.2.tgz#c18369f265297319beed4e5558753cc8c1364cb3"
  integrity sha512-oxlxJxcQlYwqPWKVJJtvQiwHgosH/LrLSPA+H4UxpyvSS6jC5aH+5MoHFM+KABgTOt0APue4w66Ha8jCUo9QGg==
  dependencies:
    accepts "~1.3.7"
    array-flatten "1.1.1"
    body-parser "1.19.1"
    content-disposition "0.5.4"
    content-type "~1.0.4"
    cookie "0.4.1"
    cookie-signature "1.0.6"
    debug "2.6.9"
    depd "~1.1.2"
    encodeurl "~1.0.2"
    escape-html "~1.0.3"
    etag "~1.8.1"
    finalhandler "~1.1.2"
    fresh "0.5.2"
    merge-descriptors "1.0.1"
    methods "~1.1.2"
    on-finished "~2.3.0"
    parseurl "~1.3.3"
    path-to-regexp "0.1.7"
    proxy-addr "~2.0.7"
    qs "6.9.6"
    range-parser "~1.2.1"
    safe-buffer "5.2.1"
    send "0.17.2"
    serve-static "1.14.2"
    setprototypeof "1.2.0"
    statuses "~1.5.0"
    type-is "~1.6.18"
    utils-merge "1.0.1"
    vary "~1.1.2"
```
<img width="940" alt="Screenshot 2023-10-30 at 22 30 48" src="https://github.com/firebase/firebase-functions/assets/69963302/e882e9aa-2920-4659-bb4a-1d218161a95b">

IchordeDionysos:
It would also be nice to update `@types/express` to the latest version :)
[...]

PR #1624, opened 2024-10-03
Bump express version to at least ^4.19.2.

taeold:
Previous express versions are vulnerable to https://nvd.nist.gov/vuln/detail/CVE-2024-29041.


package.json
```diff
@@ -272,7 +272,7 @@
     "@types/cors": "^2.8.5",
     "@types/express": "4.17.3",
```
Flet, line 273: May want to also update the types to match?

Flet:
This closes #1078

This will also also resolve several other CVEs that we're seeing in our security check on transitive dependencies:

 CVE-2024-45590 (body-parser-1.20.2)
 CVE-2024-45296 (path-to-regexp-0.1.7)
 CVE-2024-43800 (serve-static-1.15.0)
 CVE-2024-43799 (send-0.18.0)
 CVE-2024-43796 (express-4.19.2)
 
 Please take a look when you can! Thanks!

package.json
taeold: Oh interesting - there is no type versions for 4.19.2:

https://www.npmjs.com/package/@types/express?activeTab=versions

I'll bump it to the latest, compatible version which is 4.17.21.

colerogers (approved):

PR #1641, opened 2024-11-20
Add support for firebase-admin v13 (and onward)

taeold:


inlined (approved):

marco-mastella:
the issue is still there

PR #1645, opened 2024-11-21
Remove support for streaming responses for v1 callables.

taeold:


joehan (approved):
LGTM after formatting

PR #1647, opened 2024-11-22
Add GitHub action to autogenerate ref docs on commit to master.

taeold:
Example output: https://github.com/firebase/firebase-functions/actions/runs/11981676046


egilmorez (approved):
Brilliant, thanks!

PR #1650, opened 2024-12-12
Add support for an authPolicy that returns Permission Denied when failed

inlined:
Per upcoming design (and to support `onCallGenkit` callable functions' options can now support an `authPolicy` callback. Adds support for this callback, as well as two helpers: `isSignedIn` and `hasClaim`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,7 +65,7 @@
         "node": ">=14.10.0"
       },
       "peerDependencies": {
-        "firebase-admin": ">=11.0.0"
+        "firebase-admin": "^11.10.0 || ^12.0.0 || ^13.0.0"
       }
     },
     "node_modules/@babel/parser": {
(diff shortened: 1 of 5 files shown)


src/common/providers/https.ts
```diff
@@ -841,6 +842,14 @@ function wrapOnCallHandler<Req = any, Res = any>(
       }
 
       const data: Req = decode(req.body.data);
+      if (options.authPolicy) {
+        // Don't ask me why, but Google decided not to disambiguate between unauthenticated and unauthorized
```
taeold, line 846: is this true? we have permission-denied and unauthenticated:

https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

taeold (approved):

package-lock.json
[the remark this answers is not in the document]
inlined: Yes. I couldn't successfully compile until I ran `npm install` and that changed `package-lock.json`

src/common/providers/https.ts
inlined: I wrote this comment before I realized better. Will delete the comment.

PR #1651, opened 2024-12-13
Use await import instead of require for functions modules that contain top-level awaits

Berlioz:
https://github.com/firebase/firebase-tools/issues/8054 surfaced an issue in which functions containing top-level awaits threw ERR_REQUIRE_ASYNC_MODULE errors during functions discovery, due to changes in Node 22.12 (https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V22.md#requireesm-is-now-enabled-by-default).

Previously, ES modules containing top-level awaits would be imported via awaiting an import() instead of require() in a catch block after the require failed. Extending the catch conditions to cover this new error should deal with current regressions, while we consider if we need to do any other work.

Fixes https://github.com/firebase/firebase-tools/issues/8054

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,6 +32,7 @@ jobs:
         node-version:
           - 18.x
           - 20.x
+          - 22.x
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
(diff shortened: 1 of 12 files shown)


taeold:
Can we add a test case in

https://github.com/firebase/firebase-functions/tree/master/scripts/bin-test/sources?

We will also want to expand our text matrics to test against node 22 so we can actually make sure this issue is fixed:

https://github.com/firebase/firebase-functions/blob/9ed934d3dc4c2c257ac1968a65f6913c73ca779f/.github/workflows/test.yaml#L54

taeold (approved):
Add Changelog?

PR #1652, opened 2024-12-13
Update streaming callable API

inlined:
Changes the (not yet documented) streaming API for v2 Callable functions.

1. `CallableProxyResponse` is now `CallableResponse` (to match `CallableRequest`)
2. `acceptsStreaming` has been moved to `CallableRequest` because it is a request property
3. `CallableResponse.write` has been renamed `CallableResponse.sendChunk` partially to match Genkit and partially to avoid confusion where "write" acts very differently (accepts raw data, has a callback)
4. `CallableResponse.sendChunk` returns a Promise so that errors are no longer swallowed
5. Adds a `stream` method though its implementation will come at a later date


taeold (approved):

taeold (commented):
Changelog?

PR #1655, opened 2024-12-17
Adds onCallGenkit

inlined:
Adds support for a callable function specialization just for Genkit!

This design chooses to implement callable functions by locally redefining necessary interfaces to avoid a dependency on Genkit that is otherwise only needed for type definitions. Once RC.4 is released (fixing compiler errors in firebase-functions when taking a dependency on genkit) we can add genkit as a dev dependency and add a "succeeds if compiles" test to detect changes in the Genkit API.

### DIFF
--- a/src/v2/providers/https.ts
+++ b/src/v2/providers/https.ts
@@ -48,10 +48,6 @@ import * as options from "../options";
 import { withInit } from "../../common/onInit";
 import * as logger from "../../logger";
 
-// Optional dependencies:
-import type { Action } from "genkit";
-import type { ZodAny } from "zod";
-
 export { Request, CallableRequest, FunctionsErrorCode, HttpsError };
 
 /**
@@ -273,7 +269,7 @@ export interface CallableFunction<T, Return, Stream = unknown> extends HttpsFunc
   stream(
     request: CallableRequest<T>,
     response: CallableResponse<Stream>
-  ): { stream: AsyncIterator<Stream>; output: Return };
+  ): { stream: AsyncIterable<Stream>; output: Return };
 }
 
 /**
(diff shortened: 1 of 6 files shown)


inlined:
Tested manually, though it requires a CLI with https://github.com/firebase/firebase-tools/pull/8127

Code:

```typescript
import { defineSecret } from 'firebase-functions/params';
import { onCallGenkit } from "firebase-functions/https";
import { gemini15Flash, googleAI } from "@genkit-ai/googleai";
import { genkit, z } from "genkit";

const apiKey = defineSecret("GOOGLE_GENAI_API_KEY")

const ai = genkit({
  plugins: [ googleAI(), ],
  model: gemini15Flash,
});

export const jokeTeller = ai.defineFlow({
   name: "jokeTeller",
   inputSchema: z.string().nullable(),
   outputSchema: z.string(),
   streamSchema: z.string(),
}, async (type, { sendChunk }) => {
   const { stream, response } = ai.generateStream(`Tell me a ${type ?? "dad"} joke.`);
   for await (const chunk of stream) {
      sendChunk(chunk.text);
   }
   return (await response).text;
});

export const tellJoke = onCallGenkit({ secrets: [apiKey] }, jokeTeller);
```

Tests:

```bash
❯ curl -X POST -H "Content-Type: application/json" -d '{"data": "software"}' https://us-central1-inlined-junkdrawer.cloudfunctions.net/tellJoke
{"result":"Why do programmers prefer dark mode?\n\nBecause light attracts bugs!\n"}%   

❯ curl -X POST -H "Content-Type: application/json" -H "Accept: text/event-stream" -d '{"data": "software"}' https://us-central1-inlined-junkdrawer.cloudfunctions.net/tellJoke
data: {"message":"Why"}
data: {"message":" do programmers prefer dark mode?\n\nBecause light attracts bugs!\n"}
data: {"result":"Why do programmers prefer dark mode?\n\nBecause light attracts bugs!\n"}
```

Additionally tested that a second test without using onCallGenkit was able to compile successfully without taking a dependency on Genkit.

src/v2/providers/https.ts
```diff
@@ -493,3 +494,94 @@ export function onCall<T = any, Return = any | Promise<any>, Stream = unknown>(
...
+): CallableFunction<ActionInput<A>, Promise<ActionOutput<A>>, ActionStream<A>> {
+  let opts: CallableOptions<ActionInput<A>>;
+  if (arguments.length === 2) {
+    opts = optsOrAction as CallableOptions<ActionInput<A>>;
+  } else {
+    opts = {};
+    action = optsOrAction as A;
+  }
+  if (!opts.secrets?.length) {
+    logger.debug(
```
taeold, line 562: q: is this visible during `firebase deploy`?

taeold (approved):
This looks awesome!

src/v2/providers/https.ts
inlined: Only if you do --debug I think. Can figure out for sure later, but I have to run and want to let you cut a release so I can build upon it over the weekend.

inlined: Alternatively, we could push this into the CLI where it notices a genkit onCall with no secrets?

PR #1661, opened 2025-01-13
Fixed issue calling DataSnapshot methods with null data

rhodgkins:
### Description

See https://github.com/firebase/firebase-functions-test/pull/254

### DIFF
--- a/src/common/providers/database.ts
+++ b/src/common/providers/database.ts
@@ -128,7 +128,7 @@ export class DataSnapshot implements database.DataSnapshot {
     let source = this._data;
     if (parts.length) {
       for (const part of parts) {
-        if (source[part] === undefined) {
+        if (source[part] === undefined || source[part] === null) {
           return null;
         }
         source = source[part];


taeold:
Feels like this is a good change and not breaking.

IIUC the change would only break if the traversing the full path instead of short-circuiting somehow had a side effect that was worthy.

taeold (approved):

rhodgkins:
@taeold any ideas on the timeline when this might be merged / released?

rhodgkins:
Sorry I missed that there were tests in the project for this! I've added some tests for the issue to make it clearer.
I've also update the fix as adding tests exposed another issue when `this._data` is directly `null` (and not as a result of a child being missing or `null`).

src/common/providers/database.ts
```diff
@@ -128,7 +128,7 @@ export class DataSnapshot implements database.DataSnapshot {
     let source = this._data;
     if (parts.length) {
       for (const part of parts) {
-        if (source[part] === undefined) {
+        if (typeof source === "undefined" || source === null) {
```
rhodgkins, line 131: As well as moving this "up" to the parent object I've also changed the check from `=== undefined` to a `typeof` so it matches the other checks elsewhere:

https://github.com/firebase/firebase-functions/blob/5f7b331abd5a2a38a4e84de1e51411ea5919c0c3/src/common/providers/database.ts#L173

https://github.com/firebase/firebase-functions/blob/5f7b331abd5a2a38a4e84de1e51411ea5919c0c3/src/common/providers/database.ts#L278

rhodgkins:
Anyone any update on this please?

taeold:
Thank you for your contribution again @rhodgkins

PR #1665, opened 2025-01-22
Export CallableResponse type in v2/https.

taeold:
Without the change, reference documentations for CallableResponse isn't being correctly generated.


egilmorez (approved):
Thanks!

PR #1679, opened 2025-03-11
fix(logger): no longer detects duplicate object as circular

### LINKED ISSUE #1510: Log incorrectly prints [Circular] for object property when it shares reference with another property in the same object
### [REQUIRED] Version info

**node:** 18.19.0

**firebase-functions:** 4.6.0

**firebase-tools:** 12.9.1

**firebase-admin:** 11.11.0

### [REQUIRED] Test case

My function is the following:
```ts
import { onCall } from "firebase-functions/v2/https";
import { logger } from "firebase-functions/v2";

export const testFunction = onCall(
  {
    region: "europe-west1",
  },
  (request) => {
    const commonData = { foo: "bar" }
    const obj1 = {
      data: commonData,

    }
    const obj2 = {
      data: commonData,

    }
    logger.debug("TEST 1", { obj1, obj2 })
  }
);

```

- Call the function
- View the logs

When calling the function (e.g. locally) it should log with the json payload:
```json
{
  "obj1": { "data": { "foo": "bar" } },
  "obj2": { "data": { "foo": "bar" } },
  "severity": "DEBUG",
  "message": "TEST 1"
}

```

Instead it prints the following payload:
```json
{
  "obj1": { "data": { "foo": "bar" } },
  "obj2": { "data": "[Circular]" },
  "severity": "DEBUG",
  "message": "TEST 1"
}

```

Yes, I had the same issue when viewing the logs in the log explorer.

CorieW:
Resolves #1510

### DIFF
--- a/src/logger/index.ts
+++ b/src/logger/index.ts
@@ -61,7 +61,7 @@ function removeCircular(obj: any, refs: any[] = []): any {
     return obj.toJSON();
   }
   // Only check for circularity among ancestors in the recursion stack.
-  if (refs.indexOf(obj) !== -1) {
+  if (refs.includes(obj)) {
     return "[Circular]";
   }
   // Add the current object to the recursion stack.
@@ -69,10 +69,9 @@ function removeCircular(obj: any, refs: any[] = []): any {
 
   const returnObj: any = Array.isArray(obj) ? [] : {};
   for (const key in obj) {
-    if (Object.prototype.hasOwnProperty.call(obj, key)) {
-      returnObj[key] = removeCircular(obj[key], refs);
-    }
+    returnObj[key] = removeCircular(obj[key], refs);
   }
+
   // Remove the current object from the stack once its properties are processed.
   refs.pop();
   return returnObj;
--- a/src/logger/index.ts
+++ b/src/logger/index.ts
@@ -56,31 +56,28 @@ function removeCircular(obj: any, refs: any[] = []): any {
   if (typeof obj !== "object" || !obj) {
     return obj;
   }
-  // If the object defines its own toJSON method, use it.
-  if (obj.toJSON && typeof obj.toJSON === "function") {
+  // If the object defines its own toJSON, prefer that.
+  if (obj.toJSON) {
     return obj.toJSON();
   }
-  // Only check for circularity among ancestors in the recursion stack.
   if (refs.includes(obj)) {
     return "[Circular]";
+  } else {
+    refs.push(obj);
   }
-  // Add the current object to the recursion stack.
-  refs.push(obj);
-
-  // If the object is an array, run circular check on each element.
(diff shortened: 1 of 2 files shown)


src/logger/index.ts
```diff
@@ -56,28 +56,25 @@ function removeCircular(obj: any, refs: any[] = []): any {
...
   }
-  // If the object defines its own toJSON, prefer that.
-  if (obj.toJSON) {
+  // If the object defines its own toJSON method, use it.
+  if (obj.toJSON && typeof obj.toJSON === "function") {
     return obj.toJSON();
   }
-  if (refs.includes(obj)) {
+  // Only check for circularity among ancestors in the recursion stack.
+  if (refs.indexOf(obj) !== -1) {
```
cabljac, line 64: why change to indexOf? isn't this functionally the same?

src/logger/index.ts
```diff
@@ -56,28 +56,25 @@ function removeCircular(obj: any, refs: any[] = []): any {
...
-    if (refs.includes(obj[k])) {
-      returnObj[k] = "[Circular]";
-    } else {
-      returnObj[k] = removeCircular(obj[k], refs);
+  // Add the current object to the recursion stack.
+  refs.push(obj);
+
+  const returnObj: any = Array.isArray(obj) ? [] : {};
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
```
cabljac, line 72: We are no longer doing this logic here:

```
    if (refs.includes(obj[k])) {
      returnObj[k] = "[Circular]";
    } else {
```

Makes the code slightly less efficient?

Maybe more readable, but feels like we're changing how the code works a bit?

cabljac (changes requested):
generally lgtm, couple of questions though

src/logger/index.ts
CorieW: oh yeah 😆

CorieW, line 64:
```suggestion
  if (refs.includes(obj)) {
```

src/logger/index.ts
```diff
@@ -56,28 +56,32 @@ function removeCircular(obj: any, refs: any[] = []): any {
...
-  } else {
-    returnObj = {};
+    returnObj = obj.map((item) => removeCircular(item, refs));
   }
-  for (const k in obj) {
-    if (refs.includes(obj[k])) {
-      returnObj[k] = "[Circular]";
-    } else {
-      returnObj[k] = removeCircular(obj[k], refs);
+  // If the object is a Map, run circular check on each key and value.
```
cabljac, line 75: nitpick on the comment but we're not dealing with a js "Map" here, just an object

cabljac (approved):
lgtm!

Issue #1677, opened 2025-03-04
DataSnapshot.val() throws error for nested child when parent doesn't exist


bezysoftware:
### Related issues

https://github.com/firebase/firebase-functions/issues/1426

This is very similar to the linked issue, the difference is when accessing a nested property on an object which is already null. E.g. item is inserted (`before` is null) and accessing a nested property on it via `child(...).val()` will throw. 

Here's a PR which fixed the previous issue: https://github.com/firebase/firebase-functions/pull/1432/files#diff-624712cd6925b4446aeb1f79d621685365ed0cc876000e58e1bf729ff806f34fR123

It seems clear from the code there needs to be a check if `source` is null when accessing it for the first time

![Image](https://github.com/user-attachments/assets/5ccd8051-81ea-4e57-bd40-dda9f5061e18)

### [REQUIRED] Version info

**node:** 18
**firebase-functions:** 6.2.3
**firebase-tools:** 13.32.0
**firebase-admin:** 13.1.0

### [REQUIRED] Test case

```
import * as functions from 'firebase-functions/v1';

export let test = functions
    .database
    .ref('/test/{id}')
    .onWrite((event, context) => console.log(event.before.child('anything').val()));
```

### [REQUIRED] Steps to reproduce

Insert any new value to realtime database at `/test/1/anything`. 

### [REQUIRED] Expected behavior

Call to `before.child(...).val()` should return null

### [REQUIRED] Actual behavior

TypeError: Cannot read properties of null (reading 'anything')

### Were you able to successfully deploy your functions?

Yes, but errors are thrown during runtime.

CorieW:
Hi @bezysoftware,

Thanks for reporting this issue! We’ve received it and are reviewing it. We’ll provide updates as soon as possible.

CorieW:
Reproduced

PR #1682, opened 2025-03-24
fix for DataSnapshot.val() throwing error when source is null

CorieW:
Resolves https://github.com/firebase/firebase-functions/issues/1677


cabljac (approved):
Lgtm

taeold (approved):

Issue #1623, opened 2024-10-02
onTaskDispatched with retry option is not working in firebase-functions v6


daniseijo:
### Related issues

https://github.com/firebase/firebase-functions/issues/1618 - Tried the solution proposed there but it did nothing

### [REQUIRED] Version info

**node:**

v20.12.2

**firebase-functions:**

6.0.1

**firebase-tools:**

13.20.2

**firebase-admin:**

12.6.0

### [REQUIRED] Test case

I am following [this example](https://firebase.google.com/docs/functions/task-functions?gen=2nd) from the docs to create a cloud task. The only thing I am changing in my project is adding an option to avoid retries. If I add this in firebase-functions v6, and launch the emulator or try to deploy, it breaks.

### [REQUIRED] Steps to reproduce

- Create a new project from scratch with `firebase init`. In my case, I added typescript but I think it is irrelevant.
- Update `firebase-functions` in the `package.json` to v6.0.1
- Add a new `onTaskDispatched` function to the `index.ts` file and add the retry option:
```typescript
import { onTaskDispatched } from 'firebase-functions/v2/tasks';

export const cloudTaskTest = onTaskDispatched({ retry: false }, (task) => {
  logger.info('Task dispatched', { task });
  return Promise.resolve();
});
```
- Run `npm run serve` or `npm run deploy` either way an error will show:
```sh
⬢  functions: Failed to load function definition from source: FirebaseError: Failed to parse build specification
```

### [REQUIRED] Expected behavior

The expected behavior should be for the emulator to create the queue. This behavior is achieved in version 5.1.0 for example:

```sh
✔  tasks: Created queue with key: queue:sample-project-us-central1-cloudTaskTest
✔  functions[us-central1-cloudTaskTest]: http function initialized (http://127.0.0.1:5001/sample-project/us-central1/cloudTaskTest).
```

### [REQUIRED] Actual behavior

From version 6.0.1, the emulator returns the error copied above:

```sh
⬢  functions: Failed to load function definition from source: FirebaseError: Failed to parse build specification
```

And if I try to deploy it returns:

```sh
Error: Failed to parse build specification:
- FirebaseError Unexpected key 'endpoints[cloudTaskTest].taskQueueTrigger.retry'. You may need to install a newer version of the Firebase CLI.
```

### Were you able to successfully deploy your functions?

No, I was not able. The error message was:

```sh
Error: Failed to parse build specification:
[...]

PR #1686, opened 2025-04-14
fix(tasks): removes the 'retry' option from tasks

CorieW:
Resolves https://github.com/firebase/firebase-functions/issues/1623

**Note:** this issue is part of a larger issue where `TaskQueueOptions` extends from `EventHandlerOptions`, which provides properties that seemingly have no purpose and aren’t supported in the [API specification](https://github.com/firebase/firebase-tools/blob/master/src/deploy/functions/runtimes/discovery/v1alpha1.ts#L240) for tasks.


cabljac (approved):
lgtm

taeold (approved):

PR #1693, opened 2025-05-20
Refactor: Make `defineString` generic for better type-safety

keeganpotgieter:
### Description

The `defineString` function does not narrow the type when adding `input.select.options`. This means we have to typecast in certain situations, even though we know what the types should be.

### Code sample

```ts
const environmentParam = params.defineString("ENVIRONMENT", {
      default: "production", // This will now error out if it does not match one of the options
      input: {
        select: {
          options: [
            { label: "Production", value: "production" },
            { label: "Staging", value: "staging" },
            { label: "Development", value: "development" },
          ],
        },
      },
});

const environment = environmentParam.value();
// ^? "production" | "staging" | "development"
```

### DIFF
--- a/src/params/index.ts
+++ b/src/params/index.ts
@@ -130,7 +130,7 @@ export function defineSecret(name: string): SecretParam {
  * @param options Configuration options for the parameter.
  * @returns A parameter with a `string` return type for `.value`.
  */
-export function defineString(name: string, options: ParamOptions<string> = {}): StringParam {
+export function defineString<T extends string>(name: string, options: ParamOptions<T> = {}): StringParam<T> {
   const param = new StringParam(name, options);
   registerParam(param);
   return param;
--- a/src/params/types.ts
+++ b/src/params/types.ts
@@ -293,12 +293,18 @@ export interface SelectOptions<T = unknown> {
   value: T;
 }
 
+/**
+ * This can be removed once typescript is upgraded to v5.4+
+ * @internal
+ */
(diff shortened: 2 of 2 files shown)


src/params/types.ts
```diff
@@ -524,10 +530,10 @@ export class JsonSecretParam<T = any> {
  *  A parametrized value of String type that will be read from .env files
  *  if present, or prompted for by the CLI if missing.
  */
-export class StringParam extends Param<string> {
+export class StringParam<T extends string = string> extends Param<T> {
   /** @internal */
-  runtimeValue(): string {
-    return process.env[this.name] || "";
+  runtimeValue(): T {
+    return process.env[this.name] as T;
```
cabljac, line 536: What happens when it's undefined? doesn't this cast `undefined as T`

src/params/index.ts
```diff
@@ -149,7 +149,7 @@ export function defineJsonSecret<T = any>(name: string): JsonSecretParam<T> {
  * @param options Configuration options for the parameter.
```
cabljac, line 149: I think the JSDoc needs updating?

Maybe something like this?
```
 @returns A parameter whose `.value()` return type is narrowed based on the options provided.
 ```

File: src/params/types.ts [hunk 3]
```diff
@@ -293,12 +293,18 @@ export interface SelectOptions<T = unknown> {
   value: T;
 }
 
+/**
```
cabljac, line 296: Can we add a comment here with a brief explanation of why `NoInfer` is needed here for future maintainers?

Issue #1676, opened 2025-02-24
Is it possible to make traceContext of common/trace public?


mzeiher:
### Related issues

-none

### [REQUIRED] Version info

**node**
v21.7.3

**firebase-functions:**
6.3.1

**firebase-tools:**
v13.20.2

**firebase-admin:**

### [REQUIRED] Test case

```
import { onCall } from 'firebase-functions/https';
import { traceContext } from 'firebase-functions/lib/common/trace';

export const callable = onCall((request) => {
  logger.info('Hello from Firebase!');
  // it would be nice to be able to access trace context to use it for adding meta information to firebase documents or forward it if doing specific calls.
  logger.info("traceContext", traceContext);
});
```

### [REQUIRED] Steps to reproduce

firebase deploy <function>

### [REQUIRED] Expected behavior

deployment works, and I get trace information.

### [REQUIRED] Actual behavior

deployment fails with error
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/common/trace.js' is not defined by "exports" in *************/node_modules/firebase-functions/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:303:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:650:9)
    at resolveExports (node:internal/modules/cjs/loader:589:36)
    at Module._findPath (node:internal/modules/cjs/loader:666:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1128:27)
    at Module._load (node:internal/modules/cjs/loader:983:27)
    at Module.require (node:internal/modules/cjs/loader:1230:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (**************)
    at Module._compile (node:internal/modules/cjs/loader:1368:14) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
```

### Were you able to successfully deploy your functions?

no

### Alternative

wrap every call in own AsyncLocalStorage and/or use otel which is a bit to much, I also need to extract traceparent or X-GOOGLE-TRACE header

CorieW:
Hi @mzeiher,

Thanks for reporting this issue! We’ve received it and are reviewing it. We’ll provide updates as soon as possible.

CorieW:
I'll bring up with the team.

mzeiher:
Wow thanks :)

PR #1696, opened 2025-05-28
feat: make traceContext accessible

CorieW:
Resolves https://github.com/firebase/firebase-functions/issues/1676


src/v1/index.ts
```diff
@@ -35,6 +35,8 @@ import * as testLab from "./providers/testLab";
 
 import { setApp as setEmulatedAdminApp } from "../common/app";
 
+export { traceContext } from "../common/trace";
+
```
taeold: let's not add any new "feature" in 1st Gen namespace - not technically necessary but more of in the name of principle.

PR #1705, opened 2025-06-09
feat: disable error logging stacktrace in message

### LINKED ISSUE #1681: logger.error prints the stacktrace of line where it was invoked
### Related issues

Somewhat related to:
- https://github.com/firebase/firebase-functions/issues/1580
- https://github.com/firebase/firebase-functions/issues/1625

### [REQUIRED] Version info

**node:** 20

**firebase-functions:** 6.2.0

**firebase-tools:** 13.30.0

**firebase-admin:** 12.6.0

### [REQUIRED] Test case

```ts
import { logger } from "firebase-functions/v1";
import { onCall } from "firebase-functions/v2/https";

export const functionName =
    onCall(
        {
            region: ["europe-west1"],
            minInstances: minInstancesConfig,
            maxInstances: 200,
            memory: "512MiB",
            concurrency: 120,
            timeoutSeconds: 120,
        }, async (req) => {
// ...
logger.error("User is not allowed to upload data for this userId", { "234", "123" });
// ...
}
```

Error stacktrace should not be printed in plain text if not provided

Code sample above results with following outcome in GCP Logging:

Image

Yes

CorieW:
Resolves https://github.com/firebase/firebase-functions/issues/1681


src/v2/options.ts
```diff
@@ -232,6 +232,15 @@ export interface GlobalOptions {
    * may inadvertently be wiped out.
    */
   preserveExternalChanges?: boolean;
+
+  /**
+   * Controls whether error logging should include the traceback of the error automatically.
+   * Defaults to false.
+   *
+   * @remarks
+   * When true, the error message will include not include the traceback of the error.
```
cabljac: small typo "include not include"

spec/logger.spec.ts
```diff
@@ -218,4 +219,26 @@ describe("logger", () => {
...
+    it("should include stacktrace in error logs", () => {
+      const message = "Test error with stacktrace";
+      logger.error(message);
+      const messageOutput = JSON.parse(lastErr.trim()).message;
+
+      expect(messageOutput).to.include(`Error: ${message}`);
+    });
+
+    it("when disableErrorLoggingTraceback is set to true, should not log stacktrace", () => {
+      const message = "Test error with traceback disabled";
```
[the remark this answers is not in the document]
cabljac: @CorieW do we clean this up, or could it affect other tests like copilot says? might be better to be cautious and clean up

src/v2/options.ts
```diff
@@ -232,6 +232,15 @@ export interface GlobalOptions {
...
+   */
+  disableErrorLoggingTraceback?: boolean;
```
cabljac: I would use "stacktrace" instead of "traceback" as afaik traceback is mostly a python term?

cabljac: disableErrorStackTrace or suppressErrorStack or something?

spec/logger.spec.ts
```diff
@@ -218,4 +219,26 @@ describe("logger", () => {
       });
     });
   });
+
+  describe("error logging stacktrace", () => {
```
cabljac, line 224: I think we should have a test for the default behaviour when the new option is false or not set? to ensure we're not breaking default behaviour

cabljac (changes requested):
Some minor changes/comments

src/v2/options.ts
CorieW: Just went with `disableErrorLoggingStacktrace`, since it refers to logging. I feel like `disableErrorStacktrace` isn't clear enough.

`disableErrorLogStacktrace` could be an alternative?

spec/logger.spec.ts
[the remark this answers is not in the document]
CorieW: Added

src/v2/options.ts
CorieW: Good spot

src/v2/options.ts
cabljac: ya disableErrorLoggingStacktrace sounds good!

cabljac (approved):
lgtm

src/v2/options.ts
```diff
@@ -232,6 +232,15 @@ export interface GlobalOptions {
    * may inadvertently be wiped out.
    */
   preserveExternalChanges?: boolean;
+
+  /**
+   * Controls whether error logging should include the stacktrace of the error automatically.
+   * Defaults to false.
+   *
```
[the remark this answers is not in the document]
CorieW: This would be a breaking change

PR #1727, opened 2025-08-28
Add Firebase Data Connect v2 support

lisajian:
### Description

Adds basic FDC function trigger for onMutationExecuted

Includes unit tests. Also fixes a typo on an existing file

`npm test` passes locally

Internal tracking bug: [b/436623435](http://b/436623435)

### Code sample

Examples of how to use this

```
import { onMutationExecuted } from "firebase-functions/v2/dataconnect";

exports.myfunction = onMutationExecuted(
  {
    service: "service-id",
    connector: "connector-id",
    operation: "mutation-name",
  },
  (event) => { 
    // event.params.service == "service-id" 
    // event.params.connector == "connector-id"
    // event.params.operation == "mutation-name"

    // event.data contains [MutationEventData](https://docs.google.com/document/d/1sCYuBzjhirVTy9Xo2gflGVcKcZyhau9MhkVlbOvC4MI/edit?tab=t.0#bookmark=id.1k7sz3vu87vx)  
  }
);

// Use capture group on operation
exports.myfunctionwithcaptures = onMutationExecuted(
  {
    service: "service-id",
    connector: "connector-id",
    operation: "{operation}",
  },
  (event) => { /** do something */ }
);

// Use wildcard on connector
exports.myfunctionwithwildcards = onMutationExecuted(
  {
    service: "service-id",
    connector: "*",
    operation: "mutation-name",
  },
  (event) => { /** do something */ }
);
```

See [go/fdc-eventarc-functions-sdk](http://go/fdc-eventarc-functions-sdk) for more information

### DIFF
--- a/src/v2/providers/dataconnect.ts
+++ b/src/v2/providers/dataconnect.ts
@@ -259,19 +259,25 @@ function makeEndpoint(
   const eventFilterPathPatterns: Record<string, string> = {};
 
   if (service) {
-    service.hasWildcards()
-      ? (eventFilterPathPatterns.service = service.getValue())
-      : (eventFilters.service = service.getValue());
+    if (service.hasWildcards()) {
+      eventFilterPathPatterns.service = service.getValue();
+    } else {
+      eventFilters.service = service.getValue();
+    }
   }
   if (connector) {
-    connector.hasWildcards()
-      ? (eventFilterPathPatterns.connector = connector.getValue())
-      : (eventFilters.connector = connector.getValue());
+    if (connector.hasWildcards()) {
+      eventFilterPathPatterns.connector = connector.getValue();
+    } else {
+      eventFilters.connector = connector.getValue();
+    }
   }
   if (operation) {
-    operation.hasWildcards()
-      ? (eventFilterPathPatterns.operation = operation.getValue())
-      : (eventFilters.operation = operation.getValue());
+    if (operation.hasWildcards()) {
+      eventFilterPathPatterns.operation = operation.getValue();
+    } else {
+      eventFilters.operation = operation.getValue();
+    }
   }
   return {
     ...initV2Endpoint(getGlobalOptions(), opts),


spec/v2/providers/dataconnect.spec.ts
```diff
@@ -0,0 +1,447 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2023 Firebase
```
taeold:
```suggestion
// Copyright (c) 2025 Firebase
```

src/v2/providers/dataconnect.ts
```diff
@@ -0,0 +1,306 @@
...
+): CloudFunction<DataConnectEvent<MutationEventData<Variables, ResponseData>, DataConnectParams<PathPatternOrOptions>>> {
+  return onOperation<Variables, ResponseData, PathPatternOrOptions>(mutationExecutedEventType, mutationOrOpts, handler);
+}
+
+function getOpts(mutationOrOpts: string | OperationOptions) {
+  const operationRegex = new RegExp("services/([^/]+)/connectors/([^/]+)/operations/([^/]+)");
+
+  let service: string;
+  let connector: string;
+  let operation: string;
```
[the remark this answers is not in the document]
lisajian: Done.

src/v2/providers/dataconnect.ts
```diff
@@ -0,0 +1,306 @@
...
+function onOperation<Variables, ResponseData, PathPatternOrOptions>(
+  eventType: string,
+  mutationOrOpts: PathPatternOrOptions,
+  handler: (event: DataConnectEvent<MutationEventData<Variables, ResponseData>, DataConnectParams<PathPatternOrOptions>>) => any | Promise<any>
+): CloudFunction<DataConnectEvent<MutationEventData<Variables, ResponseData>, DataConnectParams<PathPatternOrOptions>>> {
+  const { service, connector, operation, opts } = getOpts(mutationOrOpts);
+
+  const servicePattern = new PathPattern(service);
+  const connectorPattern = new PathPattern(connector);
+  const operationPattern = new PathPattern(operation);
```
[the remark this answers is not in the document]
lisajian: Don't want empty strings, otherwise it'll try and find filters that exactly equals empty strings

src/v2/providers/dataconnect.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  service.hasWildcards()
+    ? (eventFilterPathPatterns.service = service.getValue())
+    : (eventFilters.service = service.getValue());
+  connector.hasWildcards()
+    ? (eventFilterPathPatterns.connector = connector.getValue())
+    : (eventFilters.connector = connector.getValue());
+  operation.hasWildcards()
+    ? (eventFilterPathPatterns.operation = operation.getValue())
+    : (eventFilters.operation = operation.getValue());
```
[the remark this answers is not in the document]
lisajian: Modified the check in the case that service | connector | operation is undefined

src/v2/providers/dataconnect.ts
```diff
@@ -0,0 +1,306 @@
...
+    const params = makeParams<Variables, ResponseData>(event, servicePattern, connectorPattern, operationPattern);
+
+    const dataConnectEvent: DataConnectEvent<MutationEventData<Variables, ResponseData>, DataConnectParams<PathPatternOrOptions>> = {
+      ...event,
+      authType: event.authtype,
+      authId: event.authid,
+      params: params as DataConnectParams<PathPatternOrOptions>,
+    };
+    delete (dataConnectEvent as any).authtype;
+    delete (dataConnectEvent as any).authid;
```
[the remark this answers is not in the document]
lisajian: Done

src/v2/providers/dataconnect.ts
```diff
@@ -0,0 +1,365 @@
...
+  return onOperation<Variables, ResponseData, PathPatternOrOptions>(
+    mutationExecutedEventType,
+    mutationOrOpts,
+    handler
+  );
+}
+
+function getOpts(mutationOrOpts: string | OperationOptions) {
+  const operationRegex = new RegExp(
+    "locations/([^/]+)/services/([^/]+)/connectors/([^/]*)/operations/([^/]+)"
```
lisajian: @taeold - Could use some feedback on recent commit changes. With the recent changes, a customer can set the region by doing one of:

1. Passing in region to GlobalOptions
2. Passing in region to OperationOptions, e.g.

```
onMutationExecuted({
  region: "us-west4"
}, (event) => {});
```

3. Pass in region to mutation "resource"

```
onMutationExecuted(
  "locations/us-west4/services/*/connectors/*/operations/*",
  (event) => {}
);
```

Not sure about option 3, and it strays a bit from the original API proposal. Other products seem to only expose options 1 and 2 as ways to set the region. WDYT? 

CC: @yuchenshi  - if you have thoughts too from today's bug bash

fredzqm: 1) & 2) make sense.

Question for discussion: What should be the default behavior when region isn't specified explicitly?

A) Default to `us-central1`
     Consistent with RTDB & Firestore, but most likely leads to unmatched trigger because FDC's default region is `us-east4`.

B) Default to `us-east4`.
     Match FDC's default region. Inconsistent with other triggers.

C) Error out and request customers to always specify it. Either globally or per trigger.

yuchenshi: I'm fine with 1 or 2, and maybe even 3.

I'm against B because we may still want to change the default region later based on capacity etc. in the CLI. At that point, we may also want to change the functions SDK to match -- but that'd be a breaking change. So mismatches may be more common.

C means the string overload form will probably rarely be used, no? If that's the case, we may not even need that overload, let alone a "locations/..." form

lisajian: Discussed offline w/ @taeold capturing some of the points here:

No contention for 1) and 2) since that's what exists in SDK today. As for 3) similar concerns around inconsistency with other triggers.

Let's refrain from introducing the "locations/..." for now. I'll include a callout in our docs that the default is us-central1 and can adjust according to customer feedback. I'll close this thread once I've made the appropriate updates

src/v2/providers/dataconnect.ts
```diff
@@ -0,0 +1,363 @@
...
+  const baseOpts = optionsToEndpoint(getGlobalOptions());
+  const specificOpts = optionsToEndpoint(opts);
+
+  const eventFilters: Record<string, string> = {};
+  const eventFilterPathPatterns: Record<string, string> = {};
+
+  if (service) {
+    service.hasWildcards()
+      ? (eventFilterPathPatterns.service = service.getValue())
+      : (eventFilters.service = service.getValue());
```
taeold: nit* find this ternary operator a bit odd. I usually expect ternarary operator to return a value. For assignments, I might jus stick to classic if-statements

src/v2/providers/dataconnect.ts
```diff
@@ -0,0 +1,363 @@
...
+export type AuthType = "app_user" | "admin" | "unknown";
+
+/** OperationOptions extend EventHandlerOptions with a provided service, connector, and operation. */
+export interface OperationOptions<
+  Service extends string = string,
+  Connector extends string = string,
+  Operation extends string = string
+> extends EventHandlerOptions {
+  /** Firebase Data Connect service ID */
+  service?: Service;
```
taeold, line 104: i'm a bit surprised that all these options are optional 🤔  should the all be required? what does not providing one mean? is it interpreted as wildcard or something else

lisajian: Good question - leaving it empty is equivalent wildcard, effectively no filter for those fields. This matches what's possible on GCP Console. 

Side bar: Interestingly, you can also add a filter for an empty string, which is different than no filter. Empty string is possible for the Connector filter

src/v2/providers/dataconnect.ts
lisajian: Done

spec/v2/providers/dataconnect.spec.ts
lisajian: Looks like this comment is outdated

PR #1745, opened 2025-10-16
Add defineJsonSecret API for structured secret configuration

taeold:
Implements `defineJsonSecret()` to retrieve JSON objects in Cloud Secret Manager.

Useful for consolidating related secrets (e.g., API keys, webhooks, client IDs) into a single secret, reducing costs and improving organization.

`defineJsonSecret` is really just a string that the SDK will parse as JSON at runtime. If the value is `undefined` or not a valid JSON, the call to `.value()` will immediately throw.

This change also introduces a backward compatible wire protocol changes (backward compatible). A new annotation `format` will be set to `json` for json secrets, and CLI can use it to validate the user input as valid JSON at deploy time.

### DIFF
--- a/src/params/index.ts
+++ b/src/params/index.ts
@@ -125,7 +125,7 @@ export function defineSecret(name: string): SecretParam {
 }
 
 /**
- * Declares a secret parameter that stores a structured JSON object in Cloud Secret Manager.
+ * Declares a secret parameter that retrieves a structured JSON object in Cloud Secret Manager.
  * This is useful for managing groups of related configuration values, such as all settings
  * for a third-party API, as a single unit.
  *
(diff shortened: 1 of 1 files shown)


spec/params/params.spec.ts
```diff
@@ -223,6 +239,96 @@ describe("Params value extraction", () => {
...
+    expect(() => jsonSecret.value()).to.throw(
+      'No value found for secret parameter "NON_EXISTENT". A function can only access a secret if you include the secret in the function\'s dependency array.'
+    );
+  });
+
+  it("handles empty object JSON", () => {
+    const jsonSecret = params.defineJsonSecret("EMPTY_OBJECT");
+    const value = jsonSecret.value();
+    expect(value).to.deep.equal({});
+  });
```
jhuleatt, line 283: is this the behavior we want? maybe we should error on empty JSON

src/params/index.ts
```diff
@@ -123,6 +124,47 @@ export function defineSecret(name: string): SecretParam {
...
+ * Declares a secret parameter that stores a structured JSON object in Cloud Secret Manager.
+ * This is useful for managing groups of related configuration values, such as all settings
+ * for a third-party API, as a single unit.
+ *
+ * The secret value must be a valid JSON string. At runtime, the value will be automatically parsed
+ * and returned as a JavaScript object. If the value is not set or is not valid JSON, an error will be thrown.
+ *
+ * @param name The name of the environment variable to use to load the parameter.
+ * @returns A parameter whose `.value()` method returns the parsed JSON object.
+ *
```
jhuleatt: fyi @egilmore reference doc copy for `defineJsonSecret`

jhuleatt (approved):

spec/params/params.spec.ts
taeold: Given that user would've have to store `{}` in SCM, I feel kinda okay not throwing.

PR #1747, opened 2025-10-21
BREAKING: Drop Node.js 14/16 support, require Node.js >=18

taeold:
- Update package.json engines to require Node.js >=18.0.0
- Upgrade @types/node from v14 to v18
- Update TypeScript target from ES2019 to ES2022
- Add `useDefineForClassFields: false` for backward compatibility
- Update CI/CD workflows to test Node.js 18, 20, 22, and 24
- Set default CI/CD Node.js version to 24
- Update integration test documentation

### DIFF
--- a/integration_test/README.md
+++ b/integration_test/README.md
@@ -8,7 +8,7 @@ Run the integration test as follows:
 ./run_tests.sh <project_id> [<project_id2>]
 ```
 
-Test runs cycles of testing for supported Node.js versions (18, 20, 22, 24).
+Test runs cycles of testing, once for Node.js 14 and another for Node.js 16.
 
 Test uses locally installed firebase to invoke commands for deploying function. The test also requires that you have
 gcloud CLI installed and authenticated (`gcloud auth login`).


tsconfig.release.json
```diff
@@ -1,13 +1,14 @@
...
-    "lib": ["es2019"],
+    "lib": ["es2022"],
     "module": "commonjs",
     "noImplicitAny": false,
     "noUnusedLocals": true,
     "outDir": "lib",
     "stripInternal": true,
-    "target": "es2019",
+    "target": "es2022",
+    "useDefineForClassFields": false,
```
jhuleatt, line 11: neat! TIL [`useDefineForClassFields`](https://www.typescriptlang.org/tsconfig/#useDefineForClassFields)

package.json
```diff
@@ -324,6 +324,6 @@
     "firebase-admin": "^11.10.0 || ^12.0.0 || ^13.0.0"
   },
   "engines": {
-    "node": ">=14.10.0"
+    "node": ">=18.0.0"
```
jhuleatt, line 327: I'm a little surprised we don't cap the Node version. Why is that? Because the CLI guarantees we can't deploy an unsupported version?

jhuleatt (approved):

package.json
taeold: I don't think we've done it since

1) we generally expect new node versions to work
2) we don't need to cut a release to support a new major version release

tsconfig.release.json
taeold: me too! best learning experiences come when you break stuff.

PR #1748, opened 2025-10-21
BREAKING: remove functions.config() implementation

taeold:
Replace functions.config() implementation with an error that provides clear migration guidance to the params module. This removes the deprecated API as planned for the next major release while helping developers migrate.

The error message directs users to: https://firebase.google.com/docs/functions/config-env#migrate-config

BREAKING CHANGE: functions.config() now throws an error instead of returning configuration values. Migrate to environment parameters using the params module.

### DIFF
--- a/src/v1/config.ts
+++ b/src/v1/config.ts
@@ -23,14 +23,14 @@
 export { firebaseConfig } from "../common/config";
 
 /**
- * @deprecated functions.config() has been removed in firebase-functions v7.
- * Please migrate to environment parameters using the params module.
+ * @deprecated `functions.config()` has been removed in firebase-functions v7.
+ * Migrate to environment parameters using the `params` module immediately.
  * Migration guide: https://firebase.google.com/docs/functions/config-env#migrate-config
  */
 export function config(): Record<string, any> {
   throw new Error(
     "functions.config() has been removed in firebase-functions v7. " +
-      "Please migrate to environment parameters using the params module. " +
+      "Migrate to environment parameters using the params module. " +
       "Migration guide: https://firebase.google.com/docs/functions/config-env#migrate-config"
   );
 }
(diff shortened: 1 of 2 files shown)


src/v1/config.ts
```diff
@@ -20,63 +20,17 @@
...
-  } catch (e) {
-    // Do nothing
-  }
-
-  singleton = {};
+  throw new Error(
+    "functions.config() has been removed in firebase-functions v7. " +
+      "Please migrate to environment parameters using the params module. " +
+      "Migration guide: https://firebase.google.com/docs/functions/config-env#migrate-config"
+  );
```
jhuleatt, line 35: @egilmorez FYI, this is an important error message

src/v1/config.ts
```diff
@@ -20,63 +20,17 @@
...
-}
-
 /**
- * Store and retrieve project configuration data such as third-party API
- * keys or other settings. You can set configuration values using the
- * Firebase CLI as described in
- * https://firebase.google.com/docs/functions/config-env.
- *
- * @deprecated Using functions.config() is discouraged. See https://firebase.google.com/docs/functions/config-env.
+ * @deprecated functions.config() has been removed in firebase-functions v7.
```
egilmorez: Can we do formatting in this comment?  I think the answer is no, but if it's yes, we should backtick the literals :)

src/v1/config.ts
```diff
@@ -20,63 +20,17 @@
...
+ * Please migrate to environment parameters using the params module.
```
egilmorez: The style guide does not like "please."  I'd even add urgency by doing something like "[Migrate from runtime configuration](https://firebase.google.com/docs/functions/config-env#migrate-config) using the `params` module as soon as possible."  (Again assuming we could format a link, which may not be the case).

PR #1749, opened 2025-10-21
Firebase Functions v8

taeold:
* https://github.com/firebase/firebase-functions/pull/1746
* https://github.com/firebase/firebase-functions/pull/1747


package.json
jhuleatt, line 3: time to update version here too? (and in package-lock.json?)

PR #1750, opened 2025-10-25
feat: enable ESM support via dual-publishing

taeold:
- Configure tsdown to build both CJS and ESM outputs.
- Update package.json exports to support both 'require' and 'import'.
- Fix type hygiene issues by using explicit 'export type'.
- Update protos/update.sh to generate ESM version of compiledFirestore and fix its imports for Node.js compatibility.
- Configure build aliases to handle relative paths to protos correctly in both CJS and ESM builds.
- Add packaging integration test to CI.

~Note: I enabled `esModuleInterop` in tsconfig.release.json to support default imports from CommonJS modules (like cors), which is required for proper ESM interop. This caused some chain-effect that required updating several unit tests and source files to use default imports for other CJS dependencies (e.g., fs, nock, jsonwebtoken) to align with the new compiler configuration.~ Fixed in https://github.com/firebase/firebase-functions/pull/1751.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 22.x
+          - 20.x
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 22.x
+          node-version: 20.x
           cache: npm
       - run: npm ci
       - run: npm run build
--- a/scripts/verify-exports.mjs
+++ b/scripts/verify-exports.mjs
@@ -22,9 +22,10 @@ console.log(`Found ${entryPoints.length} entry points to verify.`);
 let hasError = false;
 
 async function verify() {
-  console.log('\n--- Verifying CommonJS (require) ---');
+  console.log('\n--- Verifying Entry Points (CJS & ESM) ---');
   for (const exp of entryPoints) {
     const importPath = exp === '.' ? 'firebase-functions' : `firebase-functions/${exp.replace('./', '')}`;
+
     try {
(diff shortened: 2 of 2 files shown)


scripts/verify-exports.mjs
```diff
@@ -0,0 +1,57 @@
...
+  for (const exp of entryPoints) {
+    const importPath = exp === '.' ? 'firebase-functions' : `firebase-functions/${exp.replace('./', '')}`;
+    try {
+      await import(importPath);
+      console.log(`✅ ESM: ${importPath}`);
+    } catch (e) {
+      console.error(`❌ ESM Failed: ${importPath}`, e.message);
+      hasError = true;
+    }
+  }
```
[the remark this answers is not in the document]
taeold: nah. i don't see this being an issue yet to worry about.

jhuleatt, line 44: nit: since both loops do the same thing, just with different import statements, would it be a little easier to follow if it were a single loop that had a try/catch for ESM and a try/catch for CJS? 

when I first read it I was trying to find the differences in `importPath` but then realized it was the same

.github/workflows/test.yaml
```diff
@@ -14,17 +14,36 @@ jobs:
...
       - run: npm ci
       - run: npm run lint
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22.x
```
jhuleatt: just confirming that it's ok to only build on one Node version

jhuleatt (approved):

.github/workflows/test.yaml
taeold: yet i think this is fine. it matches the builder version we have here:

https://github.com/firebase/firebase-functions/blob/master/scripts/publish-container/Dockerfile

... which is actually node20! I'm going to make this node20 instead 😓

PR #1751, opened 2025-10-26
chore: enable esModuleInterop

taeold:
- Enable esModuleInterop in tsconfig.release.json for better CJS/ESM interop.
- Update unit tests and source code to use default imports for CJS modules (fs, nock, jsonwebtoken, etc.) to align with new compiler settings.

Changes are in preparation for ESM publishing for Functions SDK. Enablign `esModuleInterop` should align the codebase's import style to better match how nodejs supports default imports from "legacy" commonjs only modules like express.

Note: I also updated all `fs` import to use default import style. Default import enables fs module to be mocked.

### DIFF
--- a/src/bin/firebase-functions.ts
+++ b/src/bin/firebase-functions.ts
@@ -24,7 +24,7 @@
 
 import * as http from "http";
 import express from "express";
-import * as fs from "fs/promises";
+import fs from "fs/promises";
 import * as path from "path";
 import { loadStack } from "../runtime/loader";
 import { stackToWire } from "../runtime/manifest";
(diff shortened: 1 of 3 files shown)


src/bin/firebase-functions.ts
```diff
@@ -23,7 +23,7 @@
 // SOFTWARE.
 
 import * as http from "http";
-import * as express from "express";
+import express from "express";
 import * as fs from "fs/promises";
```
jhuleatt: fs was changed in `config.ts` and `config.spec.ts`, but not here

taeold: good comment. I was wondering why I needed to change this in my tests to make mocking work. Here's what Gemini says:

> When you enable esModuleInterop: true in your tsconfig.json:
> 1. `import * as fs from "fs"`: TypeScript treats fs as an ES Module Namespace object. According to the ECMAScript specification, namespace objects are immutable.
> 2. `import fs from "fs"`: TypeScript synthesizes a default export from the CommonJS module.exports of the fs module. This results in a standard, mutable JavaScript object.


I'm going to follow it's advice and update all fs imports to use default import style.

PR #1753, opened 2025-10-27
chore: migrate to ESLint v9 and Flat Config

taeold:
Upgraded eslint to v9 and replaced the legacy `.eslintrc.js` with the new `eslint.config.js` format.

Applied minor fixes to 4 files (renaming unused catch variables to _e and one small refactor) to satisfy slightly stricter baseline rules from updated plugins.

Mostly doing this due to our upgrade in TS to latest v5 version which the old versions of eslint ts plugins no longer support.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -43,8 +43,7 @@ module.exports = [
                 project: "tsconfig.json",
                 tsconfigRootDir: __dirname,
             },
-            ecmaVersion: 2018,
-            sourceType: "module",
+            ecmaVersion: 2022
         },
         plugins: {
             "prettier": require("eslint-plugin-prettier"),


eslint.config.js
```diff
@@ -0,0 +1,111 @@
...
+        "prettier"
+    ),
+    {
+        languageOptions: {
+            parser: require("@typescript-eslint/parser"),
+            parserOptions: {
+                project: "tsconfig.json",
+                tsconfigRootDir: __dirname,
+            },
+            ecmaVersion: 2018,
```
jhuleatt: tsconfig is on 2022:

https://github.com/firebase/firebase-functions/blob/7027996440286a29d8e961ff44e599eeec1f79a6/tsconfig.release.json#L4

eslint.config.js
```diff
@@ -0,0 +1,111 @@
...
+            sourceType: "module",
```
jhuleatt: lowpri nit: [`module` is the default](https://eslint.org/docs/latest/use/configure/language-options#specifying-javascript-options) so this can be left out

eslint.config.js
taeold: I'd love to do this, but this change DESTROYED our lint set up haha. I'll leave a comment on why I kept this for now.

taeold: oh wait this is not the problem! Great catch!

eslint.config.js
taeold: great catch!

PR #1755, opened 2025-10-28
feat: add async error handling to onRequest

taeold:
We'll now  wraps an HTTP handler with a safety net for unhandled errors.

This wrapper catches both synchronous errors and rejected Promises from `async` handlers. Without this, an unhandled error in an `async` handler would cause the request to hang until the platform timeout, as Express (v4) does not await handlers.

This is a BREAKING CHANGE in the most technical sense - unhandled errors in async onRequest handlers now immediately return a 500 Internal Server Error instead of hanging until timeout. This may affect clients relying on timeouts (for whatever reason).

### DIFF
--- a/spec/helper.ts
+++ b/spec/helper.ts
@@ -40,7 +40,11 @@ export interface RunHandlerResult {
  * data populated into the response.
  */
 export function runHandler(
-  handler: express.Handler,
+  handler: (
+    req: https.Request,
+    res: express.Response,
+    next?: express.NextFunction
+  ) => void | Promise<void>,
   request: https.Request
 ): Promise<RunHandlerResult> {
   return new Promise((resolve) => {
@@ -119,7 +123,7 @@ export function runHandler(
       }
     }
     const response = new MockResponse();
-    handler(request, response as any, () => undefined);
+    return void handler(request, response as any, () => undefined);
   });
 }
 
--- a/spec/v1/providers/httpsAsync.spec.ts
+++ b/spec/v1/providers/httpsAsync.spec.ts
@@ -17,21 +17,25 @@ describe("CloudHttpsBuilder async onRequest", () => {
   });
 
   it("should catch and log unhandled rejections in async onRequest handlers", async () => {
-    const error = new Error("Async error");
+    const err = new Error("boom");
     const fn = https.onRequest(async (_req, _res) => {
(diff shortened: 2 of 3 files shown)


spec/helper.ts
```diff
@@ -40,7 +40,11 @@ export interface RunHandlerResult {
  * data populated into the response.
  */
 export function runHandler(
-  handler: express.Handler,
+  handler: (
+    req: https.Request,
+    res: express.Response,
+    next?: express.NextFunction
+  ) => void | Promise<void>,
```
jhuleatt, line 47: Is this changed because of this PR, or a correction to how it has always worked? Just curious. I don't see `rawBody` referenced anywhere in the diff

jhuleatt: ooooh, it's a test helper.

spec/helper.ts
```diff
@@ -119,7 +123,7 @@ export function runHandler(
       }
     }
     const response = new MockResponse();
-    handler(request, response as any, () => undefined);
+    return void handler(request, response as any, () => undefined);
```
jhuleatt, line 126: what is this syntax?

src/v2/providers/https.ts
```diff
@@ -319,6 +320,8 @@ export function onRequest(
     opts = optsOrHandler as HttpsOptions;
   }
 
+  handler = withErrorHandler(handler);
```
jhuleatt, line 323: does this add the error handler to Callables, too?

spec/helper.ts
taeold: yep had to do this to make the types happier 😢

spec/helper.ts
taeold: I think this is unncessary change. undoing.

(void operator to basically return undefined, but this should be same as not doing return call at all)

taeold: ah shoot i did this because this was a way to suppress TS 😢

src/v2/providers/https.ts
taeold: ah ha we already do this!

https://github.com/firebase/firebase-functions/blob/1bd75b0423e0c18876b8432ca4c5586408f8ae0e/src/common/providers/https.ts#L941

PR #1757, opened 2025-10-30
BREAKING: Break TS build that uses functions.config()

taeold:
Type out `functions.config()` API. Use of `functions.config()` API should now cause a BUILD error at compile time.

```ts
import {onRequest} from "firebase-functions/https";
import * as logger from "firebase-functions/logger";

const cfg = functions.config()

export const helloWorld = onRequest((request, response) => {
  logger.info("Hello logs!", cfg);
  response.send("Hello from Firebase!");
});
```
```bash
$ npm run build

> build
> tsc

src/index.ts:5:23 - error TS2349: This expression is not callable.
  Type 'never' has no call signatures.

5 const cfg = functions.config()
                        ~~~~~~


Found 1 error in src/index.ts:5
```

### DIFF
--- a/src/v1/config.ts
+++ b/src/v1/config.ts
@@ -33,4 +33,4 @@ export const config: never = (() => {
       "Migrate to environment parameters using the params module. " +
       "Migration guide: https://firebase.google.com/docs/functions/config-env#migrate-config"
   );
-}) as unknown as never;
+}) as never;


src/v1/config.ts
```diff
@@ -27,10 +27,10 @@ export { firebaseConfig } from "../common/config";
...
  */
-export function config(): Record<string, any> {
+export const config: never = (() => {
   throw new Error(
     "functions.config() has been removed in firebase-functions v7. " +
       "Migrate to environment parameters using the params module. " +
       "Migration guide: https://firebase.google.com/docs/functions/config-env#migrate-config"
   );
-}
+}) as unknown as never;
```
jhuleatt: What does adding in the `as unknown` do? I tried removing it in the TS playground and it doesn't seem to change anything

taeold: iiuc `as unknown` basically erases types before and I think I needed to do it to ultimately cast this type as never to prevent tsc from complaining. let me double check to confirm!

taeold: Confirmed that this isn't necessary. Great catch!

PR #1758, opened 2025-11-03
Firebase Functions 7.0.0

taeold:
* Breaking: Minimum Node.js version is now 18. As of v7.0.0, the SDK requires a Node.js runtime of version 18 or newer. Support for Node.js 14 and 16 has been removed.
  * Impact: Functions using this SDK version will fail to deploy or run on Node.js runtimes older than v18.
  * Action Required: Before upgrading the SDK, you must [set the node.js version](https://firebase.google.com/docs/functions/manage-functions#set-node.js) to update your function's runtime environment to Node.js 18 or higher. We recommend using the latest Node.js LTS version supported by Cloud Functions. 

* Breaking: The functions.config() API has been removed. Following the deprecation schedule announced in v6.0.0, functions.config() has been removed entirely.
  * Impact: Calling this API will now throw a runtime error and cause build time errors if you are using TypeScript. Any function relying on functions.config() will fail at runtime.
  * Action Required: You must [migrate from runtime configuration](https://firebase.google.com/docs/functions/config-env#migrate-config) to environment variables using the params module. 

* Breaking: TypeScript v5 is now required. The SDK has been upgraded from TypeScript v4.3 to v5.9 and now targets ES2022.
  * Impact: This may introduce new type-checking errors in your project if you are using an older version of TypeScript or if your code relies on older type definitions.
  * Action Required: Update your project's TypeScript version to v5.x and resolve any resulting type errors.

* Breaking: Improved error handling for async handlers in the Emulator. We have improved how unhandled errors are processed in the Local Emulator Suite. Previously, an unhandled error or rejected Promise in an async onRequest handler would cause the request to hang until a platform timeout.
  * Impact: As of v7.0.0, these unhandled errors will now correctly and immediately return a 500 Internal Server Error.
  * Action Required: This is a technical breaking change, as it changes error-handling behavior. This provides faster, more accurate error feedback during local development. Ensure your error handling logic is correct and not reliant on the old timeout behavior.

* Feature: The SDK is now published as both CommonJS (CJS) and ECMAScript Modules (ESM). This provides first-class support for modern bundlers (like Vite and esbuild) and native import syntax, which can lead to optimized function sources and smaller bundle sizes. The package.json exports map now properly resolves for both require and import.

* Feature: New FDC Trigger This release adds a new function trigger for Firebase Data Connect (FDC), onMutationExecuted(). This allows you to execute functions in response to data mutations.


jhuleatt (approved):
🚢 🚢 🚢

PR #1764, opened 2025-11-05
Fix prerelease publishing

taeold:
Allows 'prerelease' as a version argument to increment RC versions. Ensures that the version bump commit and tag are pushed to GitHub even for pre-releases, preventing subsequent runs from attempting to publish the same version.

The current release "fails" to create a second prerelease 😢


brittanycho (approved):
wdyt of the gemini-code-assist comments (re: git push)?

PR #1770, opened 2025-11-13
add authcontxt params to the DatabaseEvent object

kushalpalesha:
### Description

Add authctxt fields so users can access them in their v2 functions.

### Code sample

```
// The Cloud Functions for Firebase SDK to setup triggers and logging.
const {onValueWritten} = require("firebase-functions/v2/database");
const {logger} = require("firebase-functions");

exports.dbtrigger = onValueWritten("/messages/{pushId}/original", (event) => {
  // Log the AuthContext
  logger.log(`Auth type: ${event.authType}`);
  logger.log(`Auth ID: ${event.authId}`);
``
  // Grab the current value of what was written to the Realtime Database.
  const original = event.data.after.val();
  logger.log("Uppercasing", event.params.pushId, original);
  const uppercase = original.toUpperCase();

  // You must return a Promise when performing asynchronous tasks inside a function.
  return event.data.after.ref.parent.child("uppercase").set(uppercase);
});
```

### DIFF
--- a/src/v2/providers/database.ts
+++ b/src/v2/providers/database.ts
@@ -403,7 +403,7 @@ function makeDatabaseEvent<Params>(
     firebaseDatabaseHost: event.firebasedatabasehost,
     data: snapshot,
     params,
-    authType: event.authtype,
+    authType: event.authtype || "unknown",
     authId: event.authid,
   } as any;
   delete (databaseEvent as any).firebasedatabasehost;
@@ -433,7 +433,7 @@ function makeChangedDatabaseEvent<Params>(
       after,
     },
     params,
-    authType: event.authtype,
+    authType: event.authtype || "unknown",
     authId: event.authid,
   } as any;
   delete (databaseEvent as any).firebasedatabasehost;


src/v2/providers/database.ts
```diff
@@ -384,8 +403,12 @@ function makeDatabaseEvent<Params>(
     firebaseDatabaseHost: event.firebasedatabasehost,
     data: snapshot,
     params,
-  };
+    authType: event.authtype,
```
shettyvarun268: Since authtype might be undefined on legacy events (or older versions), but DatabaseEvent.authType is typed as a required field, should we provide a default fallback here to ensure runtime safety? Something like authType: event.authtype ?? "unknown" .

src/v2/providers/database.ts
```diff
@@ -410,8 +433,12 @@ function makeChangedDatabaseEvent<Params>(
       after,
     },
     params,
-  };
+    authType: event.authtype,
```
shettyvarun268: Same here. Please correct me if I'm missing something.

spec/v2/providers/database.spec.ts
```diff
@@ -461,6 +461,21 @@ describe("database", () => {
...
+        authType: "unauthenticated",
+      };
+
+      const func = database.onValueWritten("foo/bar", (event) => {
+        expect(event.authId).to.equal("uid");
+        expect(event.authType).to.equal("unauthenticated");
+      });
+
+      await func(raw as any);
+    });
```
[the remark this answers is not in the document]
kushalpalesha: done

spec/v2/providers/database.spec.ts
```diff
@@ -461,6 +463,19 @@ describe("database", () => {
...
+        ...RAW_RTDB_EVENT,
+      };
+
+      const func = database.onValueWritten("foo/bar", (event) => {
+        expect(event.authId).to.equal("uid");
+        expect(event.authType).to.equal("unauthenticated");
+      });
+
+      await func(raw);
+    });
```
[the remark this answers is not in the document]
kushalpalesha: done

src/v2/providers/database.ts
kushalpalesha: done!

src/v2/providers/database.ts
kushalpalesha: done!

Issue #1775, opened 2025-11-24
Using parameters for `minInstances` runtime option fails during deployment with ESM (v7.0.0)


weevilgenius:
Using a parameterized config value in an ESM project for a function's `minInstances` runtime option fails to deploy with a `CEL expression 'params.MY_PARAMETER' is of an unsupported form` error. It works in v6.6.0 and below and also works if you change your project to CommonJS.

### Version info

**node:** 24.11.1

**firebase-functions:** 7.0.0

**firebase-tools:** 14.26.0

**firebase-admin:** N/A

### Test case

With `"type": "module"` in your package.json:

```ts
import { onRequest } from "firebase-functions/https";
import { defineInt } from "firebase-functions/params";

const CRITICAL_HOT_INSTANCES = defineInt('CRITICAL_HOT_INSTANCES', {
  description: 'How many instances to keep hot for critical functions?',
  default: 0,
});

export const exampleFunc = onRequest({ minInstances: CRITICAL_HOT_INSTANCES },
  (req, res) => {
    res.send("Hello, world!");
  }
);
```

### Steps to reproduce

Deploy the function using Firebase CLI: `firebase deploy --only functions:exampleFunc`

### Expected behavior

Function deploys with `minInstances` set by the parameter.

### Actual behavior

```
$ firebase deploy --only functions:exampleFunc

=== Deploying to 'dev'...

i  deploying functions
Running command: pnpm run build

> firebase-functions-bug@1.0.0 build /Users/user/projects/functions-bug
> tsc

✔  functions: Finished running predeploy script.
i  functions: preparing codebase default for deployment
i  functions: ensuring required API cloudfunctions.googleapis.com is enabled...
i  functions: ensuring required API cloudbuild.googleapis.com is enabled...
i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
i  functions: Loading and analyzing source code for codebase default to determine what to deploy
Serving at port 8857

i  extensions: ensuring required API firebaseextensions.googleapis.com is enabled...
i  functions: Loaded environment variables from .env, .env.dev.

Error: CEL expression 'params.CRITICAL_HOT_INSTANCES' is of an unsupported form
```

Note that deployment halts immediately and no `firebase-debug.log` is produced.

CorieW:
Hi @weevilgenius. Thanks for creating this issue! We'll review it as soon as possible.

CorieW:
[...]

Issue #1779, opened 2025-12-03
Invalid bucket name, while deploying function (v7.0.0)


mandalornl:
### [REQUIRED] Version info

**node:** 22.19.0

**firebase-functions:** 7.0.0

**firebase-tools:** 14.26.0

**firebase-admin:** 13.6.0

### [REQUIRED] Test case

```javascript
import { defineString } from 'firebase-functions/params';
import { onObjectFinalized } from 'firebase-functions/storage';

const myStorageBucket = defineString('MY_STORAGE_BUCKET');

export const test = onObjectFinalized({
  bucket: myStorageBucket,
}, () => {
  // Do something
});
```

Using e.g. `import { storageBucket } from 'firebase-functions/params'` results in the same issue.

### [REQUIRED] Steps to reproduce

1. Upgrade from firebase-functions 6.x to 7.0.0.
2. Deploy function.

### [REQUIRED] Expected behavior

I expect the function to deploy without issues. 
Works fine in <= 6.x.

### [REQUIRED] Actual behavior

```log
[...]
[2025-12-03T09:03:24.344Z] Looking up bucket region for the storage event trigger
[2025-12-03T09:03:24.345Z] Looking up bucket region for the storage event trigger on bucket params.MY_STORAGE_BUCKET
[2025-12-03T09:03:24.345Z] Checked if tokens are valid: true, expires at: 1764754302033
[2025-12-03T09:03:24.346Z] Checked if tokens are valid: true, expires at: 1764754302033
[2025-12-03T09:03:24.346Z] >>> [apiv2][query] GET https://storage.googleapis.com/storage/v1/b/params.MY_STORAGE_BUCKET [none]
[2025-12-03T09:03:24.699Z] <<< [apiv2][status] GET https://storage.googleapis.com/storage/v1/b/params.MY_STORAGE_BUCKET 400
[2025-12-03T09:03:24.700Z] <<< [apiv2][body] GET https://storage.googleapis.com/storage/v1/b/params.MY_STORAGE_BUCKET {"error":{"code":400,"message":"Invalid bucket name: 'params.MY_STORAGE_BUCKET'","errors":[{"message":"Invalid bucket name: 'params.MY_STORAGE_BUCKET'","domain":"global","reason":"invalid"}]}}
Request to https://storage.googleapis.com/storage/v1/b/params.MY_STORAGE_BUCKET had HTTP Error: 400, Invalid bucket name: 'params.MY_STORAGE_BUCKET'
[...]

Issue #1777, opened 2025-11-26
SecretParam and other param types not exported from firebase-functions/params in v7.0.0


unxavi:
### [REQUIRED] Version info

**node:** v22.20.0

**firebase-functions:** 7.0.0

**firebase-tools:** 14.1.0

**firebase-admin:** 13.1.0

### [REQUIRED] Test case

```typescript
// authMiddleware.ts
import { https } from "firebase-functions";
import type { SecretParam } from "firebase-functions/params";

export const authenticatedFunction = (
  secrets: (string | SecretParam)[],
  handler: (request: https.CallableRequest, userId: string) => Promise<any>,
) => {
  return https.onCall({ secrets: secrets }, async (request) => {
    if (\!request.auth) {
      throw new https.HttpsError("failed-precondition", "Must be authenticated.");
    }
    return handler(request, request.auth.uid);
  });
};
```

### [REQUIRED] Steps to reproduce

1. Upgrade from firebase-functions 6.x to 7.0.0
2. Attempt to import `SecretParam` type: `import type { SecretParam } from "firebase-functions/params"`
3. Run TypeScript compiler

### [REQUIRED] Expected behavior

The `SecretParam` type should be importable from `firebase-functions/params` since:
- It's defined and exported in `lib/params/types.d.ts`
- It's used as the return type of the exported `defineSecret()` function
- Similar param types (`StringParam`, `BooleanParam`, etc.) should also be available for type annotations

### [REQUIRED] Actual behavior

TypeScript compilation fails with:
```
error TS2459: Module '"firebase-functions/params"' declares 'SecretParam' locally, but it is not exported.
```

**Root cause:**
Looking at `lib/params/index.d.ts`, the type is imported internally:
```typescript
import { BooleanParam, Expression, IntParam, Param, ParamOptions, SecretParam, JsonSecretParam, StringParam, ListParam } from "./types";
```

But it's never re-exported. The file only exports:
```typescript
export { BUCKET_PICKER, select, multiSelect } from "./types";
export type { TextInput, SelectInput, SelectOptions, MultiSelectInput } from "./types";
export { Expression };
export type { ParamOptions };
```

**Workaround:**
Use TypeScript's `ReturnType` utility:
```typescript
import type { params } from "firebase-functions";
type SecretParam = ReturnType<typeof params.defineSecret>;
```

### Were you able to successfully deploy your functions?

[...]

PR #1780, opened 2025-12-04
fix: Resolve dual-package hazard for params in ESM

taeold:
This PR resolves two critical issues caused by the "Dual-Package Hazard" when using `firebase-functions` in ESM projects:

1. Separate instances of the `declaredParams` in `firebase-functions/params` for CJS and ESM led to params not being detected when used in ESM modules.

To fix, we implement a global singleton pattern for `declaredParams` using `globalThis` and a version-scoped `Symbol.for` key. This ensures both builds share the same global storage for collecting param API uses.

2. `instanceof` checks failed because the CJS `Expression`/`ResetValue` instances differed from the ESM instances.

To fix, we implement a custom `[Symbol.hasInstance]` on `Expression` and `ResetValue` classes using branded tags to identify instances across package boundaries.

Fixes https://github.com/firebase/firebase-functions/issues/1775, https://github.com/firebase/firebase-functions/issues/1777, https://github.com/firebase/firebase-functions/issues/1779

### DIFF
--- a/src/common/options.ts
+++ b/src/common/options.ts
@@ -44,7 +44,7 @@ export class ResetValue {
     return null;
   }
   // eslint-disable-next-line @typescript-eslint/no-empty-function
-  private constructor() { }
+  private constructor() {}
   public static getInstance() {
     return new ResetValue();
   }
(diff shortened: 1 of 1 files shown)


jhuleatt (approved):
gnarly!

Once Node 20 is EOL in the spring, we should be able to stop publishing the cjs package entirely: https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require. Hopefully that will simplify things!

PR #1782, opened 2025-12-04
Implement V1 compat API for PubSub

shettyvarun268:
This PR adds the v1-compatible argument getters (context and message) directly to the v2 Pub/Sub CloudEvent<MessagePublishedData> class.
This allows developers to access v1-style arguments via property access, e.g., const { message, context } = event;, for a smoother migration of existing business logic to the v2 function signature.Implementation Details

The v1-compatible context getter is implemented as a lazily-computed property on the CloudEvent object. To correctly construct the v1 EventContext from the v2 event's properties, the following two internal helper functions are introduced:

1. `attachPubSubContext<T>(event: CloudEvent<MessagePublishedData<T>>, topic: string):`
Adds the v1-style context property to the v2 event object using Object.defineProperty.
The getter computes the EventContext by mapping v2 properties (event.id, event.time) and by calling getResourceName.
This ensures the context is only computed once upon first access.

2. `getResourceName(event: CloudEvent<MessagePublishedData<any>>, topic: string):`
Extracts the full resource name for the v1 EventContext resource object.
It attempts to parse the project ID and topic name from event.source or falls back to environment variables and the function's topic name parameter if necessary.

This is the first of a series of PRs aimed at simplifying the migration from v1 to v2. By introducing this new 
onMessagePublished function, we are providing a more robust and flexible way to handle Pub/Sub events in Cloud Functions for Firebase v2.

### DIFF
--- a/spec/v2/providers/pubsub.spec.ts
+++ b/spec/v2/providers/pubsub.spec.ts
@@ -1,7 +1,6 @@
 import { expect } from "chai";
 import { EventContext } from "../../../src/v1/cloud-functions";
 
-import { CloudEvent } from "../../../src/v2/core";
 import * as options from "../../../src/v2/options";
 import * as pubsub from "../../../src/v2/providers/pubsub";
 import { FULL_ENDPOINT, MINIMAL_V2_ENDPOINT, FULL_OPTIONS, FULL_TRIGGER } from "./fixtures";
@@ -130,9 +129,23 @@ describe("onMessagePublished", () => {
   it("should have a .run method", () => {
     const func = pubsub.onMessagePublished("topic", (event) => event);
 
-    const res = func.run("input" as any);
+    const event: pubsub.PubSubCloudEvent<any> = {
+      specversion: "1.0",
+      source: "//pubsub.googleapis.com/projects/aProject/topics/topic",
+      id: "uuid",
+      type: EVENT_TRIGGER.eventType,
+      time: new Date().toISOString(),
+      data: {
+        message: { data: Buffer.from("input").toString("base64") } as any,
+        subscription: "projects/aProject/subscriptions/aSubscription",
+      },
+      message: { data: Buffer.from("input").toString("base64") } as any,
+      context: {} as EventContext,
+    };
 
-    expect(res).to.equal("input");
+    const res = func.run(event);
+
+    expect(res).to.deep.equal(event);
   });
 
   it("should parse pubsub messages", async () => {
@@ -148,7 +161,9 @@ describe("onMessagePublished", () => {
       message: messageJSON as any,
       subscription: "projects/aProject/subscriptions/aSubscription",
     };
-    const event: CloudEvent<pubsub.MessagePublishedData<any>> = {
+    const event: pubsub.PubSubCloudEvent<any> = {
+      context: {} as EventContext,
+      message: publishData.message as any,
       specversion: "1.0",
       source: "//pubsub.googleapis.com/projects/aProject/topics/topic",
       id: "uuid",
@@ -158,7 +173,7 @@ describe("onMessagePublished", () => {
     };
 
     const func = pubsub.onMessagePublished("topic", (event) => {
-      json = event.data.message.json;
+      json = (event.data as any).message.json;
       return event;
     });
 
@@ -182,7 +197,9 @@ describe("onMessagePublished", () => {
       message: messagePayload as any,
       subscription: "projects/aProject/subscriptions/aSubscription",
     };
-    const event: CloudEvent<pubsub.MessagePublishedData> = {
+    const event: pubsub.PubSubCloudEvent<pubsub.MessagePublishedData> = {
+      context: {} as EventContext,
+      message: data.message as any,
       specversion: "1.0",
       id: "uuid",
       time: publishTime,
@@ -191,16 +208,11 @@ describe("onMessagePublished", () => {
       data,
     };
 
-    type PubSubCloudEvent = CloudEvent<pubsub.MessagePublishedData> & {
-      message: pubsub.Message<{ hello: "world" }>;
-      context: EventContext;
-    };
-
-    let destructuredMessage: pubsub.Message<{ hello: "world" }>;
+    let destructuredMessage: pubsub.Message<any>;
     let context: EventContext;
-    const func = pubsub.onMessagePublished("topic", (e) => {
-      const pubsubEvent = e as PubSubCloudEvent;
-      ({ message: destructuredMessage, context } = pubsubEvent);
+    const func = pubsub.onMessagePublished("topic", (e: pubsub.PubSubCloudEvent<any>) => {
+      destructuredMessage = e.message;
+      context = e.context;
     });
 
     await func(event);
@@ -221,58 +233,22 @@ describe("onMessagePublished", () => {
     pubsub.onMessagePublished<string>(
       "topic",
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      (event: CloudEvent<pubsub.MessagePublishedData<string>>) => undefined
+      (event: pubsub.PubSubCloudEvent<string>) => undefined
     );
     pubsub.onMessagePublished<string>(
       "topic",
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      (event: CloudEvent<pubsub.MessagePublishedData>) => undefined
+      (event: pubsub.PubSubCloudEvent<any>) => undefined
     );
     pubsub.onMessagePublished(
       "topic",
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      (event: CloudEvent<pubsub.MessagePublishedData<string>>) => undefined
+      (event: pubsub.PubSubCloudEvent<string>) => undefined
     );
     pubsub.onMessagePublished(
       "topic",
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      (event: CloudEvent<pubsub.MessagePublishedData>) => undefined
+      (event: pubsub.PubSubCloudEvent<any>) => undefined
     );
-  });
-
-
-
-
-
-
-  it("should use 'unknown-project' as fallback for resource name", async () => {
-    delete process.env.GCLOUD_PROJECT;
-    const publishTime = new Date().toISOString();
-    const message = {
-      messageId: "uuid",
-      data: Buffer.from(JSON.stringify({ hello: "world" })).toString("base64"),
-      publishTime,
-    };
-    const data: pubsub.MessagePublishedData = {
-      message: message as any,
-      subscription: "projects/aProject/subscriptions/aSubscription",
-    };
-    const event: CloudEvent<pubsub.MessagePublishedData> = {
-      specversion: "1.0",
-      id: "uuid",
-      time: publishTime,
-      type: "google.cloud.pubsub.topic.v1.messagePublished",
-      source: "//pubsub.googleapis.com/topics/topic", // Malformed source
-      data,
-    };
-
-    let receivedEvent: CloudEvent<pubsub.MessagePublishedData<any>>;
-    const func = pubsub.onMessagePublished("topic", (e) => {
-      receivedEvent = e;
-    });
-
-    await func(event);
-
-    expect(receivedEvent.context.resource.name).to.equal("projects/unknown-project/topics/topic");
   });
 });
--- a/src/v2/core.ts
+++ b/src/v2/core.ts
@@ -25,7 +25,6 @@
  * @packageDocumentation
  */
 
-import type { EventContext } from "../v1/cloud-functions";
 import { Change } from "../common/change";
 import { ManifestEndpoint } from "../runtime/manifest";
 
@@ -92,13 +91,6 @@ export interface CloudEvent<T> {
 
   /** Information about this specific event. */
   data: T;
-
-  /** V1- compatible context of this event.
-   * 
-   * This getter is added at runtime for V1 compatibility.
-   * May be undefined if not set by a provider
-   */
-  readonly context?: EventContext;
 }
 
 /**
--- a/src/v2/providers/pubsub.ts
+++ b/src/v2/providers/pubsub.ts
@@ -152,6 +152,15 @@ export interface MessagePublishedData<T = any> {
   readonly subscription: string;
 }
 
+/** A Pub/Sub CloudEvent with a `message` property. */
+export interface PubSubCloudEvent<T> extends CloudEvent<MessagePublishedData<T>> {
+  /** Google Cloud Pub/Sub message. */
+  message: Message<T>;
+
+  /** V1- compatible context of this event. */
+  readonly context: EventContext;
+}
+
 /** PubSubOptions extend EventHandlerOptions but must include a topic. */
 export interface PubSubOptions extends options.EventHandlerOptions {
   /** The Pub/Sub topic to watch for message events */
@@ -264,8 +273,12 @@ export interface PubSubOptions extends options.EventHandlerOptions {
  */
 export function onMessagePublished<T = any>(
   topic: string,
-  handler: (event: CloudEvent<MessagePublishedData<T>>) => any | Promise<any>
-): CloudFunction<CloudEvent<MessagePublishedData<T>>>;
+  handler: (message: Message<T>, context: EventContext) => any | Promise<any>
+): CloudFunction<PubSubCloudEvent<T>>;
+export function onMessagePublished<T = any>(
+  topic: string,
+  handler: (event: PubSubCloudEvent<T>) => any | Promise<any>
+): CloudFunction<PubSubCloudEvent<T>>;
 
 /**
  * Handle a message being published to a Pub/Sub topic.
@@ -275,8 +288,12 @@ export function onMessagePublished<T = any>(
  */
 export function onMessagePublished<T = any>(
   options: PubSubOptions,
-  handler: (event: CloudEvent<MessagePublishedData<T>>) => any | Promise<any>
-): CloudFunction<CloudEvent<MessagePublishedData<T>>>;
+  handler: (message: Message<T>, context: EventContext) => any | Promise<any>
+): CloudFunction<PubSubCloudEvent<T>>;
+export function onMessagePublished<T = any>(
+  options: PubSubOptions,
+  handler: (event: PubSubCloudEvent<T>) => any | Promise<any>
+): CloudFunction<PubSubCloudEvent<T>>;
 
 /**
  * Handle a message being published to a Pub/Sub topic.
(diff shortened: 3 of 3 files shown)


src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,66 @@ export function onMessagePublished<T = any>(
...
+
+/**
+ * @internal
+ *
+ * Adds a v1-style context to the event.
+ *
+ * @param event - The event to add the context to.
+ * @param topic - The topic the event is for.
+ */
+function attachPubSubContext<T>(event: CloudEvent<MessagePublishedData<T>>, topic: string) {
```
taeold: i don't think we strictly need `topic`. all information should be derivable from event obj.

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,66 @@ export function onMessagePublished<T = any>(
...
+ *
+ * Extracts the resource name from the event source.
+ *
+ * @param event - The event to extract the resource name from.
+ * @param topic - The topic the event is for.
+ * @returns The resource name.
+ */
+function getResourceName(event: CloudEvent<MessagePublishedData<any>>, topic: string) {
+  const match = event.source?.match(/projects\/([^/]+)\/topics\/([^/]+)/);
+  const project =
```
taeold: i think this code is overly defensive.

cloudevent is a standard format, and it's unlikely you won't be able to retrieve project/topic information from the event payload. defaulitng to "unknown" or "" is probably fine in the unlikely case we can't derive the info.

src/v2/providers/pubsub.ts
```diff
@@ -304,7 +305,9 @@ export function onMessagePublished<T = any>(
       subscription: string;
     };
     messagePublishedData.message = new Message(messagePublishedData.message);
-    return wrapTraceContext(withInit(handler))(raw as CloudEvent<MessagePublishedData<T>>);
+    const event = raw as CloudEvent<MessagePublishedData<T>>;
+    attachPubSubContext(event, topic);
```
taeold: ptal look at v1 pubsub signature:

```ts
 (message: Message, context: EventContext) => PromiseLike<any> 
```

We also need a getter for `message` as well as `context`

spec/v2/providers/pubsub.spec.ts
```diff
@@ -193,4 +235,106 @@ describe("onMessagePublished", () => {
       (event: CloudEvent<pubsub.MessagePublishedData>) => undefined
     );
   });
+
+  //Test case to ensure Idempotency. makes things dont break if there is already context present
```
taeold: this comment seems redudant

spec/v2/providers/pubsub.spec.ts
```diff
@@ -170,6 +170,48 @@ describe("onMessagePublished", () => {
...
+      id: "uuid",
+      time: publishTime,
+      type: "google.cloud.pubsub.topic.v1.messagePublished",
+      source: "//pubsub.googleapis.com/projects/aProject/topics/topic",
+      data,
+    };
+
+    let receivedEvent: CloudEvent<pubsub.MessagePublishedData<any>>;
+    const func = pubsub.onMessagePublished("topic", (e) => {
+      receivedEvent = e;
```
taeold: i'd prefer we try to use the context here directly:


```suggestion
      { message, context } = e;
```

spec/v2/providers/pubsub.spec.ts
```diff
@@ -193,4 +235,106 @@ describe("onMessagePublished", () => {
       (event: CloudEvent<pubsub.MessagePublishedData>) => undefined
     );
   });
+
+  //Test case to ensure Idempotency. makes things dont break if there is already context present
+  it("should not modify a CloudEvent that already has a context", async () => {
```
taeold: not sure why this exist? why would there be existing context?

taeold (changes requested):

src/v2/core.ts
```diff
@@ -91,6 +92,13 @@ export interface CloudEvent<T> {
 
   /** Information about this specific event. */
   data: T;
+
+  /** V1- compatible context of this event.
+   * 
+   * This getter is added at runtime for V1 compatibility.
+   * May be undefined it not set by a provider
+   */
```
jhuleatt: @egilmorez heads up, copy here will go into the reference docs on this page: https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.cloudevent

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,66 @@ export function onMessagePublished<T = any>(
...
+  const context: EventContext = {
+    eventId: event.id,
+    timestamp: event.time,
+    resource,
+    eventType: "google.cloud.pubsub.topic.v1.messagePublished",
+    params: {}
+
+  };
+
+  Object.defineProperty(event, "context", {
```
jhuleatt, line 416: for my own learning, is `defineProperty` used here so that `event.context` isn't settable?

src/v2/core.ts
```diff
@@ -25,6 +25,7 @@
  * @packageDocumentation
  */
 
+import type { EventContext } from "../v1/cloud-functions";
```
jhuleatt: Would it be worth redefining this as `1stGenEventContext` or `LegacyEventContext` so that it's clear it is not really meant for people using 2nd gen?

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,66 @@ export function onMessagePublished<T = any>(
...
+/**
+ * @internal
+ *
+ * Adds a v1-style context to the event.
+ *
+ * @param event - The event to add the context to.
+ * @param topic - The topic the event is for.
+ */
+function attachPubSubContext<T>(event: CloudEvent<MessagePublishedData<T>>, topic: string) {
+  if ("context" in event && event.context) {
```
jhuleatt: when would context already be in the event? should this throw if that happens?

spec/v2/providers/pubsub.spec.ts
shettyvarun268: Made the changes. Used deconstructedMessage and context instead

spec/v2/providers/pubsub.spec.ts
shettyvarun268: Noted. Addressed this by removing the test case below.

spec/v2/providers/pubsub.spec.ts
shettyvarun268: Understood. The real cases won't have an existing case. Removed this test case.

src/v2/providers/pubsub.ts
shettyvarun268: Got it. Made the needed changes. Added the message getter. This would allow us to access event.messag directly now.

src/v2/providers/pubsub.ts
shettyvarun268: Made the change and got rid of the topic.

src/v2/providers/pubsub.ts
shettyvarun268: Good point. Changed the code to throw an error in this case. The context existing would be unlikely. Would you prefer I get rid of the check altogether?

src/v2/providers/pubsub.ts
shettyvarun268: That's correct. This is to ensure event.context a read-only property so as to not accidentally end up changing it inside the function.

src/v2/providers/pubsub.ts
shettyvarun268: Good point. Removed the condition. Also removed the corresponding test case because it won't be needed.

src/v2/core.ts
shettyvarun268: Does v1EventContext sound more clear and concise?

src/v2/core.ts
shettyvarun268: Would you like me to make the comments more informative and developer friendly. Considering it might be put into the reference docs directly from here?

spec/v2/providers/pubsub.spec.ts
```diff
@@ -170,6 +170,45 @@ describe("onMessagePublished", () => {
...
+      time: publishTime,
+      type: "google.cloud.pubsub.topic.v1.messagePublished",
+      source: "//pubsub.googleapis.com/projects/aProject/topics/topic",
+      data,
+    };
+
+    let destructuredMessage: pubsub.Message<any>;
+    let context: any;
+    const func = pubsub.onMessagePublished("topic", (e) => {
+      ({ message: destructuredMessage, context } = e as any);
```
taeold: curious - can we do better than using `any` types here?

spec/v2/providers/pubsub.spec.ts
```diff
@@ -193,4 +232,40 @@ describe("onMessagePublished", () => {
...
     );
   });
+
+
+
+
+
+
+  it("should use 'unknown-project' as fallback for resource name", async () => {
+    delete process.env.GCLOUD_PROJECT;
```
taeold: is this necessary?

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,70 @@ export function onMessagePublished<T = any>(
 
   return func;
 }
+
+/**
+ * @internal
+ *
+ * Adds a v1-style context to the event.
```
taeold: nit* this is adding both context and data (`message`) to it. Maybe we need a slightly better name + update the docstring to reflect that?

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,70 @@ export function onMessagePublished<T = any>(
...
+    eventType: "google.cloud.pubsub.topic.v1.messagePublished",
+    params: {}
+
+  };
+
+  Object.defineProperty(event, "context", {
+
+    get: () => context,
+    enumerable: false,
+    configurable: false,
```
taeold: q: why set these value to false?

and also - they are by default set to false.

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,70 @@ export function onMessagePublished<T = any>(
...
+
+  Object.defineProperty(event, "message", {
+    get: () => (event.data as MessagePublishedData<T>).message,
+    enumerable: false,
+    configurable: false,
+  });
+}
+
+/**
+ * @internal
```
taeold: this seems to be unexported function, so @internal marking is unncessary imo

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,70 @@ export function onMessagePublished<T = any>(
...
+function attachPubSubContext<T>(event: CloudEvent<MessagePublishedData<T>>, topic: string) {
+  if ("context" in event && event.context) {
+    throw new Error("Unexpected context in event.");
+  }
+
+  const resourceName = getResourceName(event, topic);
+  const resource: Resource = {
+
+    service: "pubsub.googleapis.com",
+    name: resourceName ?? "",
```
taeold: this seems overly defensive. iiuc resourceName would never be undefined according to getResourceName impl

src/v2/core.ts
taeold: i;d want to right. 


right now, the fact that we reveal this as a "getter" is a bit odd. Imagine user looking at this and trying to understand WHAT this is. also, will it actually ever be undefined?

spec/v2/providers/pubsub.spec.ts
shettyvarun268: Yeah I think we should keep this. In case we receive a malformed event where the EventID is not parsed, this would ensure that the function does not crash. Also considering GCLOUD_PROJECT is an environmental variable, there could be a case when its not populated (?). What do you think?

src/v2/core.ts
```diff
@@ -91,6 +92,13 @@ export interface CloudEvent<T> {
 
   /** Information about this specific event. */
   data: T;
+
+  /** V1- compatible context of this event.
+   * 
+   * This getter is added at runtime for V1 compatibility.
+   * May be undefined it not set by a provider
```
brittanycho: nit: May be undefined if not set by a provider*

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,62 @@ export function onMessagePublished<T = any>(
...
+  if ("context" in event && event.context) {
+    throw new Error("Unexpected context in event.");
+  }
+
+  const resourceName = getResourceName(event, topic);
+  const resource: Resource = {
+
+    service: "pubsub.googleapis.com",
+    name: resourceName,
+
```
brittanycho: nit: looks like some empty lines might be causing lint failures

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,62 @@ export function onMessagePublished<T = any>(
...
+/**
+ * Adds v1-style `context` and `message` properties to the event.
+ *
+ * @param event - The event to add the context to.
+ * @param topic - The topic the event is for.
+ */
+function addV1Compatibility<T>(event: CloudEvent<MessagePublishedData<T>>, topic: string) {
+  if ("context" in event && event.context) {
+    throw new Error("Unexpected context in event.");
```
brittanycho: nit: should we add a test for this unexpected context error?

src/v2/core.ts
shettyvarun268: Good catch! Thank you!

src/v2/providers/pubsub.ts
shettyvarun268: Tried removing the lines. The lint errors seem to be coming from the use of "any" in other files.

src/v2/providers/pubsub.ts
shettyvarun268: As pointed out by Daniel previously, the 'context' will be received in a standard format and it is highly unlikely that it will have an existing context. I had a test case earlier but removed it as per Daniel's recommendation since it seemed redundant. Felt like this test case still made sense to be thorough so kept it there.

spec/v2/providers/pubsub.spec.ts
```diff
@@ -170,6 +171,51 @@ describe("onMessagePublished", () => {
...
+    const event: CloudEvent<pubsub.MessagePublishedData> = {
+      specversion: "1.0",
+      id: "uuid",
+      time: publishTime,
+      type: "google.cloud.pubsub.topic.v1.messagePublished",
+      source: "//pubsub.googleapis.com/projects/aProject/topics/topic",
+      data,
+    };
+
+    type PubSubCloudEvent = CloudEvent<pubsub.MessagePublishedData> & {
```
taeold: it is pretty uncommon to define a type outside of top-level. I'd suggest just inlining this.

(adding on, there's an argument on whether this specific typing is helpful or not! For testing code, i'd not concern too much with narrowly typing things out)

e.g. `message: pubsub.Message<{ hello: "world" }>`; could just be `message: pubsub.Message`; without a lot of loss of type info.

spec/v2/providers/pubsub.spec.ts
taeold: anything could happen, but it would be extremely rare. My general thinking here is that if we aren't aware of any condition/scenarios where this could happen, then we may choose to fail explicilty and early instead of letting the unknown condition seem like it's working as intended.

Also - i'm not sure if `GCLOUD_PROJECT` env var is actually used in our logic? Isn't project id extracted from the data included in the cloud event payload?

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,60 @@ export function onMessagePublished<T = any>(
...
   return func;
 }
+
+/**
+ * Adds v1-style `context` and `message` properties to the event.
+ *
+ * @param event - The event to add the context to.
+ * @param topic - The topic the event is for.
+ */
+function addV1Compatibility<T>(event: CloudEvent<MessagePublishedData<T>>, topic: string) {
```
taeold: nit* I think I'd prefer name like `addV1CompatProperties` to be more explicit

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,60 @@ export function onMessagePublished<T = any>(
...
+  Object.defineProperty(event, "context", {
+
+    get: () => context,
+
+  });
+
+  Object.defineProperty(event, "message", {
+    get: () => (event.data as MessagePublishedData<T>).message,
+  });
+}
```
taeold, line 423: can you make sure that the type of `event.message` is properly typed when using it?

src/v2/providers/pubsub.ts
```diff
@@ -353,3 +356,60 @@ export function onMessagePublished<T = any>(
...
+ * Extracts the resource name from the event source.
+ *
+ * @param event - The event to extract the resource name from.
+ * @param topic - The topic the event is for.
+ * @returns The resource name.
+ */
+function getResourceName(event: CloudEvent<MessagePublishedData<any>>, topic: string) {
+  const match = event.source?.match(/projects\/([^/]+)\/topics\/([^/]+)/);
+  const project = match?.[1];
+  const topicName = match?.[2] ?? topic;
```
taeold: nit* this is usally called ID not name

(name refers to fulll resource name)

src/v2/core.ts
```diff
@@ -91,6 +92,13 @@ export interface CloudEvent<T> {
 
   /** Information about this specific event. */
   data: T;
+
+  /** V1- compatible context of this event.
+   * 
+   * This getter is added at runtime for V1 compatibility.
+   * May be undefined if not set by a provider
+   */
+  readonly context?: EventContext;
```
taeold: I'd rather not make this change

Instead in v2/providers/pubsub.ts, consider defining a new type like this:

```
type PubSubEvent = CloudEvent<MessagePublishedData<T>> & 
{
  context: ...
  message: ...
}
```

That way, we don't make `context` property optional as well as be able to incrementally add support for compat properties to each provider

spec/v2/providers/pubsub.spec.ts
shettyvarun268: Makes sense. Removed the test case since it wasnt necessary.

spec/v2/providers/pubsub.spec.ts
shettyvarun268: Made the type definition inline.

src/v2/providers/pubsub.ts
shettyvarun268: Gotcha. Changed the name to addV1CompatProperties. And made the necessary changes to reflect this.

src/v2/providers/pubsub.ts
shettyvarun268: Thanks! Refactored the code to use topicID instead of name.

src/v2/core.ts
shettyvarun268: Noted. reverted the changes in core.ts. Instead added a new type interface in pubsub.ts itself. Going ahead all the changes will be made in the provider file itself so that we can incrementally add support.

src/v2/providers/pubsub.ts
shettyvarun268: Got it. Fixed the typing.

PR #1784, opened 2025-12-06
Initial FDC resolver support in functions SDK.

rosalyntan:
### Description

Implement the `onGraphRequest` helper function for building a Firebase Data Connect resolver detailed in go/fdc-webhooks-functions-api.

### Code sample

```
import { onGraphRequest } from 'firebase-functions/dataconnect';

const opts = {
    schemaFilePath: "dataconnect/schema_resolver/schema.gql",
    // or schema: "type Query { hello(name: String): String }",
    resolvers: {
        query: {
            hello(parent, args, contextValue, info) {
                return `Hello ${args.name}!`;
            },
        },
    },
}
exports.resolver = onGraphRequest(opts);
```

### DIFF
--- a/src/v2/providers/dataconnect.ts
+++ b/src/v2/providers/dataconnect.ts
@@ -475,13 +475,14 @@ export interface GraphqlServerOptions extends Omit<HttpsOptions, "cors"> {
    */
   schema?: string;
   /**
-   * A file path to a valid GraphQL schema.
+   * A relative file path from the Firebase project directory to a valid GraphQL schema.
    * Either `schema` or `schemaFilePath` is required.
    */
   schemaFilePath?: string;
   /**
    * The path where the GraphQL server will be served on the Cloud Run function.
-   * If no path is provided, the server will be served at `/graphql`.
+   * e.g. https://...run.app/{path}
+   * If no path is provided, "graphql" is used as the default.
    */
   path?: string;
   /** A map of functions that populate data for individual GraphQL schema fields. */


src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +377,167 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+  if ((!opts.schema && !opts.schemaFilePath) || (opts.schema && opts.schemaFilePath)) {
+    throw new Error("Exactly one of 'schema' or 'schemaFilePath' must be provided.");
+  }
+  if (opts.schemaFilePath) {
+    opts.schema = fs.readFileSync(opts.schemaFilePath, "utf-8");
+  }
+  if (!opts.resolvers.query && !opts.resolvers.mutation) {
+    throw new Error("At least one query or mutation resolver must be provided.");
+  }
+  const apolloResolvers: { [key: string]: any } = {};
```
taeold: nit* curious if this type can be tightened to match expecations of ApolloServer instead of it being a generic object.

also nit, I usually prefer using Record, e.g. `Record<string, any>`

package.json
```diff
@@ -505,9 +505,12 @@
     "@types/express": "^4.17.21",
     "cors": "^2.8.5",
     "express": "^4.21.0",
+    "graphql": "^16.12.0",
```
taeold: I only see use of graphql to import a type - can this be a dev dep?

src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +377,167 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+      // Respect `cors: false` to turn off cors even if debug feature is enabled.
+      origin = opts.cors === false ? false : true;
+    }
+    // Arrays cause the access-control-allow-origin header to be dynamic based
+    // on the origin header of the request. If there is only one element in the
+    // array, this is unnecessary.
+    if (Array.isArray(origin) && origin.length === 1) {
+      origin = origin[0];
+    }
+    const middleware = cors({ origin });
```
taeold: i wonder if cors settings are necessary. we don't expect graphql requests to be public and accessed by your web apps right?

src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +377,167 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+    const userProvidedHandler = handler;
+    handler = (req: Request, res: express.Response): void | Promise<void> => {
+      return new Promise((resolve) => {
+        res.on("finish", resolve);
+        middleware(req, res, () => {
+          resolve(userProvidedHandler(req, res));
+        });
+      });
+    };
+  }
```
taeold: if we remove cors, this might also be unncessary?

src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +377,167 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+    const { ApolloServer } = await import("@apollo/server");
+    const { expressMiddleware } = await import("@as-integrations/express4");
+    const serverPromise = (async () => {
+      const app = express();
+      const server = new ApolloServer({
+        typeDefs: opts.schema,
+        resolvers: apolloResolvers,
+      });
+      await server.start();
+      app.use(`/${opts.path ?? "graphql"}`, cors(), express.json(), expressMiddleware(server));
```
taeold: it's a bit odd that we are adding cors middleware here and again below. but i suggest removing cors concern all together if we are going to strongly suggest that custom resolvers should not be publicly exposed.

src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +377,167 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+   */
+  schemaFilePath?: string;
+  /**
+   * The path where the GraphQL server will be served on the Cloud Run function.
+   * If no path is provided, the server will be served at `/graphql`.
+   */
+  path?: string;
+  /** A map of functions that populate data for individual GraphQL schema fields. */
+  resolvers: GraphqlResolvers;
+  // TODO: Add a field for a context function.
```
taeold, line 490: do you want to leave this todo?

src/v2/providers/dataconnect.ts
taeold: hmm it looks like this should be `GraphqlResolvers` type we defined below, with the change that `query` -> `Query` and `mutation` -> `Mutation`?

src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +377,167 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+    convertInvoker
+  );
+  convertIfPresent(endpoint.dataConnectHttpsTrigger, opts, "invoker", "invoker", convertInvoker);
+  (handler as HttpsFunction).__endpoint = endpoint;
+
+  return handler as HttpsFunction;
+}
+
+/** Options for configuring the GraphQL server. */
+export interface GraphqlServerOptions extends HttpsOptions {
```
taeold: you may have noticed that most of trigger options ends up repeating the common options... this is because we learned that our reference doc generation doesn't show the full range of options that's available in the parent type (in this case `HttpOptions`).

For better reference docs, we might want to repeat all the available options for the graphql server options here.

src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +377,167 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+    ...baseOpts,
+    ...specificOpts,
+    labels: {
+      ...baseOpts?.labels,
+      ...specificOpts?.labels,
+    },
+    dataConnectHttpsTrigger: {},
+  };
+  convertIfPresent(
+    endpoint.dataConnectHttpsTrigger,
```
taeold: just trying to have fun with naming stuff (not even a public thing)

i think it might make sense for this trigger type to include "graphql" no? e.g. `dataConnectGraphQLTrigger` or something like that?

src/v2/providers/dataconnect.ts
```diff
@@ -20,6 +20,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+import cors from "cors";
+import express from "express";
+import fs from "fs";
+import { GraphQLResolveInfo } from "graphql";
```
taeold: would prefer `import type` to make sure TS treat this as type import only.

package.json
rosalyntan: Done.

src/v2/providers/dataconnect.ts
rosalyntan: Done, removed references to cors

src/v2/providers/dataconnect.ts
rosalyntan: Done.

src/v2/providers/dataconnect.ts
rosalyntan: Done.

src/v2/providers/dataconnect.ts
rosalyntan: Done.

src/v2/providers/dataconnect.ts
rosalyntan: This wasn't included in the original API proposal, so wanted to run it by you and the council (maybe in an email follow-up) before adding this field.

src/v2/providers/dataconnect.ts
rosalyntan: Done.

src/v2/providers/dataconnect.ts
rosalyntan: So looking at https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.https.httpsoptions.md, it looks like it does link to the parent type which I think seems sufficient (and I think this also makes the GraphQL-specific options more visible to developers).

On a related note, if we're ignoring CORS configuration with these options, should I do something like `Omit<GlobalOptions, "cors">` here?

src/v2/providers/dataconnect.ts
rosalyntan: The reason we're not directly using ApolloServer's types here is for future extensibility, for example the developer could pass in an option to use the `graphql-http` library instead.

src/v2/providers/dataconnect.ts
taeold: Looping in @egilmorez for some thoughts on what we want the reference docs to look like. The reference doc decision was something we made some time ago, and want to make sure this special casing make sense.

I don't think cors is part of GlobalOptions.

src/v2/providers/dataconnect.ts
```diff
@@ -368,3 +375,145 @@ function onOperation<Variables, ResponseData, PathPatternOrOptions>(
...
+export interface FirebaseContext {
+  auth?: {
+    /** The UID of the Firebase user that made the request, if present. */
+    uid?: string;
+    /** The token attached to the `X-Firebase-Auth-Token` in the request, if present. */
+    token?: string;
+  };
+}
+
+export interface GraphqlResolvers {
```
taeold, line 504: can we add documentation string here?

spec/v2/providers/dataconnect.spec.ts
```diff
@@ -527,4 +529,134 @@ describe("dataconnect", () => {
...
+        schema: "type Query { hello: String }",
+        schemaFilePath: "schema.gql",
+        resolvers: {
+          query: {
+            hello: () => "Hello, world!",
+          },
+        },
+      };
+      try {
+        await dataconnect.initGraphqlServer(opts);
```
taeold: a little more idiomatic to use `.to.throw` helper, e.g.

```
      expect(() => identity.parseMultiFactor(factors)).to.throw(
        "INTERNAL ASSERT FAILED: Invalid multi-factor info response"
      );
```

spec/v2/providers/dataconnect.spec.ts
```diff
@@ -527,4 +529,134 @@ describe("dataconnect", () => {
...
+          },
+        },
+      };
+      const func = dataconnect.onGraphRequest(opts);
+      expect(func.__endpoint).to.deep.equal({
+        ...expectedEndpointBase,
+        dataConnectGraphqlTrigger: {},
+      });
+    });
+    it("returns callable function with schemaFilePath Data Connect trigger", () => {
```
taeold: here and below - does `callable function` make sense in the test description?

spec/v2/providers/dataconnect.spec.ts
```diff
@@ -527,4 +529,134 @@ describe("dataconnect", () => {
...
+        },
+      };
+      try {
+        await dataconnect.initGraphqlServer(opts);
+      } catch (err: any) {
+        expect(err.message).to.equal(
+          "Exactly one of 'schema' or 'schemaFilePath' must be provided."
+        );
+      }
+    });
```
taeold, line 557: here and below

insert space between each test case

src/v2/providers/dataconnect.ts
rosalyntan: Done.

spec/v2/providers/dataconnect.spec.ts
rosalyntan: Done.

spec/v2/providers/dataconnect.spec.ts
rosalyntan: Done.

spec/v2/providers/dataconnect.spec.ts
rosalyntan: Done.

src/v2/providers/dataconnect.ts
rosalyntan: I'm going to go ahead and merge this PR, but I can update later if we want to change the reference docs.

PR #1795, opened 2026-01-14
Move `onGraphRequest` to different namespace so `graphql` dependency doesn't affect other dataconnect methods.

### LINKED ISSUE #1793: Missing dependency on grahql
I have a Renovate PR to bump from 7.0.2 to 7.0.3, but it seems like a peer dependency on GraphQL was introduced even though I'm not using any new features. https://github.com/k0swe/w1aw-schedule/pull/437

### Related issues

None that I found

### [REQUIRED] Version info

  **node**: v24.12.0
  npm: 11.6.2
  yarn: 1.22.22

**firebase-functions:** 7.0.3

**firebase-tools:** latest(?)

**firebase-admin:** ^13.6.0

### [REQUIRED] Test case

Attempt to upgrade from firebase-functions 7.0.2 to 7.0.3, as in https://github.com/k0swe/w1aw-schedule/pull/437.

```
firebase emulators:exec --only firestore 'npm run test:unit'
```

Bumping the patch version does not require new peer dependencies.

My [CI pipeline](https://github.com/k0swe/w1aw-schedule/actions/runs/20899543720/job/60043291789?pr=437) fails with `Cannot find module 'graphql'`:
```
⚠  emulators: You are not currently authenticated so some features may not work correctly. Please run firebase login to authenticate the CLI.
i  emulators: Starting emulators: firestore
⚠  It appears you are running in a CI environment. You can avoid downloading the Firestore Emulator repeatedly by caching the /home/runner/.cache/firebase/emulators directory.
i  firestore: downloading cloud-firestore-emulator-v1.20.2.jar...

i  firestore: Firestore Emulator logging to firestore-debug.log
✔  firestore: Firestore Emulator UI websocket is running on 9150.
i  Running script: npm run test:unit

> test:unit
> npm run build && mocha --reporter spec ./build/ --recursive


> build
> tsc

Error: node_modules/firebase-functions/lib/v2/providers/dataconnect.d.ts(2,41): error TS2307: Cannot find module 'graphql' or its corresponding type declarations.
⚠  Script exited unsuccessfully (code 2)
i  emulators: Shutting down emulators.
i  firestore: Stopping Firestore Emulator
i  hub: Stopping emulator hub
i  logging: Stopping Logging Emulator

Error: Script "npm run test:unit" exited with code 2
Error: Process completed with exit code 2.
```

Haven't tried, TS compilation failed.

rosalyntan:
### Description

1. Move the `onGraphRequest` to a separate `firebase-functions/dataconnect/graphql` to avoid its optional peer dependencies from affecting other `firebase-functions` methods. Note: we are not exporting `graphql` from `firebase-functions` for the same reason. This means an import statement like `import {graphql} from firebase-functions` will not work; developers will need to use `import {onGraphRequest} from firebase-functions/dataconnect/graphql` instead.

2. Import `ApolloServer` and `expressMiddleware` statically instead of dynamically, now that the `firebase-functions/dataconnect/graphql` module is isolated. This allows missing dependencies to be a deploy time rather than a runtime error if missing, while still not affecting usage of other `firebase-functions` namespaces.

Tested: https://paste.googleplex.com/6219756904513536

Fixes #1793

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+- Fixed issue with missing dependency on `graphql`. (#1795)


package.json
```diff
@@ -559,6 +575,7 @@
   "peerDependencies": {
     "@apollo/server": "^5.2.0",
     "@as-integrations/express4": "^1.1.2",
+    "graphql": "^16.12.0",
```
taeold, line 578: huh do you remember why we didn't add this last time?

taeold (approved):
Please add some changelog entry - the release automation fails if CHANGELOG.md is empty 😬

package.json
rosalyntan: I think it was originally a direct dependency, but then since it was a type-only import, we tried moving it to `devDependencies`.

PR #1796, opened 2026-01-16
Add Firebase Auth token to resolver context

rosalyntan:
### Description

Populates the `FirebaseContext` object with the result of the `X-Firebase-Auth-Token` request header, making it available inside resolver functions.

Removing the `uid` field for now from `FirebaseContext`.

### Code sample

```
import { FirebaseContext, onGraphRequest } from 'firebase-functions/dataconnect/graphql';
import { log } from 'firebase-functions/logger';

const opts = {
    schemaFilePath: "dataconnect/schema_test_webhook/schema.gql",
    resolvers: {
        query: {
            hello(parent, args: Record<string, unknown>, contextValue: FirebaseContext, info) {
                log(`Saying hello to the user with Firebase Auth token ${contextValue.auth.token}`);
                return `Hello ${args.name}!`;
            },
        },
    },
}
exports.resolver = onGraphRequest(opts);
```

### DIFF
--- a/src/v2/providers/dataconnect/graphql.ts
+++ b/src/v2/providers/dataconnect/graphql.ts
@@ -43,7 +43,7 @@ export async function initGraphqlServer(opts: GraphqlServerOptions): Promise<exp
         `/${opts.path ?? "graphql"}`,
         express.json(),
         expressMiddleware(server, {
-          context: async ({ req }) =>
+          context: ({ req }) =>
             Promise.resolve({
               auth: {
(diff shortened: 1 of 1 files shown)


taeold (approved):
Please add changelog entry (required for release)

rosalyntan:
> Please add changelog entry (required for release)

Chatted offline, since this is an internal/unreleased change and changelog is already populated, we don't need to add anything in this PR.

PR #1801, opened 2026-01-22
fix(v1): Call onInit for schedule.onRun functions

### LINKED ISSUE #1680: V1 schedule.onRun doesn't call onInit
firebase-functions: 6.3.2

If creating a schedule.onRun function, onInit is not called because of this lines:

Will set label deployment-scheduled to true
https://github.com/firebase/firebase-functions/blob/cce55b562f325474a2ea13d7db06392309667990/src/v1/providers/pubsub.ts#L155

makeCloudFunction is wrapping withInit on the first line
https://github.com/firebase/firebase-functions/blob/cce55b562f325474a2ea13d7db06392309667990/src/v1/cloud-functions.ts#L370

However, because of the deployment-scheduled label it uses the non wrapped handler:
https://github.com/firebase/firebase-functions/blob/cce55b562f325474a2ea13d7db06392309667990/src/v1/cloud-functions.ts#L411

Expected behavior

onInit callback is called

Actual behavior

onInit is not called and causes issues with the function execution

HassanBahati:
## Description

Fixes #1680 

When creating a V1 scheduled function using `functions.pubsub.schedule().onRun()`, the `onInit` callback was never executed. This was caused by scheduled functions (identified by the `deployment-scheduled` label) using the unwrapped `contextOnlyHandler` instead of the `withInit`-wrapped version.

## Solution

Added a separately wrapped version of `contextOnlyHandler` so that scheduled functions also execute any registered `onInit` callbacks:

- Created `wrappedContextOnlyHandler` by wrapping `contextOnlyHandler` with `withInit`
- Changed scheduled function execution to use `wrappedContextOnlyHandler` instead of the unwrapped `contextOnlyHandler`


samuel-pp:
👀, Is there anyone who can handle the merging of this PR? I see that a review is currently missing.

samuel-pp:
@cabljac Hello, I would like to know whether this fix can continue to proceed. I see that the pipeline is stuck at review required. This fix is very important for our team, as Firebase's migration rules require us to migrate runtime config to Secret Manager.

samuel-pp:
> @cabljac Hello, I would like to know whether this fix can continue to proceed. I see that the pipeline is stuck at review required. This fix is very important for our team, as Firebase's migration rules require us to migrate runtime config to Secret Manager.

@inlined Hello, I would like to know if this PR is still being processed normally.

cabljac (approved):
lgtm

PR #1802, opened 2026-01-22
fix: Remove false warning when using Expression in cors option

### LINKED ISSUE #1715: Warning about using `.value()` whenever an expression is used in the `cors` option
### [REQUIRED] Version info

**node:** 22.17.1

**firebase-functions:** 6.4.0

**firebase-tools:** 14.11.1

**firebase-admin:** 13.4.0

### [REQUIRED] Test case

```
import { onCall } from 'firebase-functions/v2/https';
import { projectID } from 'firebase-functions/params';

export const helloworld = onCall<unknown, string>(
  { cors: projectID.equals('something').thenElse(['http://localhost:5173'], []) },
  () => {
    return 'Hello from Firebase!';
  },
);
```

1. Just create a firebase-functions project and deploy the given function.

Deploys cleanly without warnings.

Gets a warning:
```
{"severity":"WARNING","message":"params.PROJECT_ID == \"something\" ? [\"http://localhost:5173\"] : [].value() invoked during function deployment, instead of during runtime."}
{"severity":"WARNING","message":"This is usually a mistake. In configs, use Params directly without calling .value()."}

{"severity":"WARNING","message":"example: { memory: memoryParam } not { memory: memoryParam.value() }"}
```

Despite not using `.value()` in my code. The `.value()` is inside firebase-functions own code.

Yes.

HassanBahati:
### Description

Fixes https://github.com/firebase/firebase-functions/issues/1715

Changed the internal CORS processing to use `.runtimeValue()` instead of `.value()` when extracting values from Expression objects. The `.runtimeValue()` method appears to be the correct internal API for this use case and doesn't emit warnings.


IzaakGough (commented):
lgtm

cabljac (approved):
lgtm

shettyvarun268 (approved):
LGTM!

`enforceAppCheck` and `consumeAppCheckToken` in `src/v2/providers/https.ts` (lines 442 & 447) also call `.value()` and suffer from the same false warning. Maybe we can address those in a separate follow-up PR.

IzaakGough:
@shettyvarun268 Thanks for reviewing this! I will have a go at the follow-up PR :)

PR #1804, opened 2026-01-23
fix(v2/pubsub): support StringParam for topic in onMessagePublished

### LINKED ISSUE #1791: Cannot use StringParam as topic value in onMessagePublished
In the types definition [here](https://github.com/firebase/firebase-functions/blob/master/src/v2/providers/pubsub.ts#L157), the topic can only be a string. Also, [here](https://github.com/firebase/firebase-functions/blob/master/src/v2/providers/pubsub.ts#L290-L299), there is no check if to see if the topic is a parameter and retrieve the value of it unlike in other function types.

### Related issues

None that I could find.

### [REQUIRED] Version info

**node:** v20.10.0

**firebase-functions:** v7.0.1

**firebase-tools:** v15.1.0

**firebase-admin:**

### [REQUIRED] Test case

```
import {defineString} from "firebase-functions/params";
import {onMessagePublished} from "firebase-functions/v2/pubsub";

const topicName = defineString("some-topic");

export const myMessageConsumer = onMessagePublished(
  {
    topic: topicName
  },
  async (message) => {
    console.log(message);
  }
);
```

```
firebase deploy --only functions:myMessageConsumer
```

See test case above

During deployment, the function should deploy successfully and the parameter used and configured at deploy-time should parse and configure the topic to receive messages from.

Function fails to deploy

No

```
src/functions/test.ts:6:34 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(topic: string, handler: (event: CloudEvent<MessagePublishedData<any>>) => any): CloudFunction<CloudEvent<MessagePublishedData<any>>>', gave the following error.
    Argument of type '{ topic: StringParam; }' is not assignable to parameter of type 'string'.
  Overload 2 of 2, '(options: PubSubOptions, handler: (event: CloudEvent<MessagePublishedData<any>>) => any): CloudFunction<CloudEvent<MessagePublishedData<any>>>', gave the following error.
    Type 'StringParam' is not assignable to type 'string'.

  6 export const myMessageConsumer = onMessagePublished(
                                     ~~~~~~~~~~~~~~~~~~~
  7     {
    ~~~~~
... 
 12     }
    ~~~~~
 13 );
    ~

  node_modules/firebase-functions/lib/v2/providers/pubsub.d.ts:86:5
    86     topic: string;
           ~~~~~
    The expected type comes from property 'topic' which is declared here on type 'PubSubOptions'
```

HassanBahati:
### Description

Fixes #1791

Update PubSubOptions.topic to accept Expression<string> in addition to string, allowing StringParam to be used for the topic value. This makes onMessagePublished consistent with other function types and options that already support Expression<string>.

### Manual Testing
Before
image
After
image


HassanBahati (commented):
/gemini review

IzaakGough (commented):
lgtm

src/v2/providers/pubsub.ts
```diff
@@ -325,7 +325,7 @@ export function onMessagePublished<T = any>(
     event: CloudEvent<MessagePublishedData<T>> & V1Compat<"message", V1PubSubMessage<T>>
   ) => any | Promise<any>
 ): CloudFunction<CloudEvent<MessagePublishedData<T>>> {
-  let topic: string;
+  let topic: string | Expression<string>;
   let opts: options.EventHandlerOptions;
   if (typeof topicOrOptions === "string") {
```
ajperel: Doesn't this if branch now also need to handle "Expression<String>"

storage.ts seems to do this with a check more like
`if (typeof bucketOrOpts === "string" || "value" in bucketOrOpts) {`

src/v2/providers/pubsub.ts
ajperel: Should this topic also be updated to string | Expression<string>? And again below in the next declaration.

ajperel (commented):
Left two comments, and if accurately things that need to be fixed, we should probably make sure the test cases _also_ cover those issues.

ajperel (approved):
Looks like you addressed all my previous feedback?

IzaakGough:
@ajperel yep I believe so!

cabljac (approved):
lgtm

PR #1813, opened 2026-02-04
Fix params.defineList cors parsed as undefined at deploy

### LINKED ISSUE #1773: params.defineList is always parsed as undefined in v6
### Related issues

I searched but did not find any

### [REQUIRED] Version info

**node:** v22.21.0 (via nvm)

**firebase-functions:** 6.6.0

**firebase-tools:** 14.25.0

**firebase-admin:** 13.6.0

### [REQUIRED] Test case

// runtimeConfig.ts
export const allowedOrigins = params.defineList('ALLOWED_ORIGINS');

// used in a function file
import * as config from '../runtimeConfig';
// I know this isn't "supposed" to be here, but needed a call outside of the function signature to get access to the cors content. I also tried to resolve it first in onInit but that was also unsuccessful.

const allowedOrigins = config.allowedOrigins;

export const loginToPortal = functions.https.onRequest(
	{ cors: allowedOrigins as unknown as string[], secrets: config.noteableSecrets },
	async (request, response) => {...}

// .env.projectId
// I tried many variations of an array syntax here including just a string, comma separated, surround with single quotes, etc. Nothing worked
ALLOWED_ORIGINS=["https://mywebsite.com", "https://app.mywebsite.com"]

Run:
`firebase deploy --only functions`

A successful deploy (which does work when I return to functions v5 and admin v12)

This did not produce a firebase-debug.log file.

➜ firebase deploy --only functions

=== Deploying to '<project-id>'...

i  deploying functions
Running command: npm --prefix "$RESOURCE_DIR" run build

> build
> npm run lint && tsc

> lint
> eslint

+  functions: Finished running predeploy script.
i  functions: preparing codebase default for deployment
i  functions: ensuring required API cloudfunctions.googleapis.com is enabled...
i  functions: ensuring required API cloudbuild.googleapis.com is enabled...
i  artifactregistry: ensuring required API artifactregistry.googleapis.com is enabled...
!  functions: package.json indicates an outdated version of firebase-functions. Please upgrade using npm install --save firebase-functions@latest in your functions directory.
!  functions: Please note that there will be breaking changes when you upgrade.
i  functions: Loading and analyzing source code for codebase default to determine what to deploy
Serving at port 8910

{"severity":"WARNING","message":"params.ALLOWED_ORIGINS.value() invoked during function deployment, instead of during runtime."}

{"severity":"WARNING","message":"This is usually a mistake. In configs, use Params directly without calling .value()."}
{"severity":"WARNING","message":"example: { memory: memoryParam } not { memory: memoryParam.value() }"}

SyntaxError: "undefined" is not valid JSON
    at JSON.parse (<anonymous>)
    at ListParam.runtimeValue (C:\Users\me\RiderProjects\Portal\functions\node_modules\firebase-functions\lib\params\types.js:420:26)
    at ListParam.value (C:\Users\me\RiderProjects\Portal\functions\node_modules\firebase-functions\lib\params\types.js:39:21)
    at Object.onRequest (C:\Users\me\RiderProjects\Portal\functions\node_modules\firebase-functions\lib\v2\providers\https.js:72:75)
    at Object.<anonymous> (C:\Users\me\RiderProjects\Portal\functions\lib\portal\function-name.js:14:41)
    at Module._compile (node:internal/modules/cjs/loader:1706:14)
    at Object..js (node:internal/modules/cjs/loader:1839:10)
    at Module.load (node:internal/modules/cjs/loader:1441:32)
    at Function._load (node:internal/modules/cjs/loader:1263:12)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)

Error: Functions codebase could not be analyzed successfully. It may have a syntax or runtime error

no

HassanBahati:
## Description

Fixes deploy failure when params.defineList() is used as the cors option in https.onRequest or https.onCall. Previously, the SDK resolved the param at definition time (when the deploy analyzer loads the code), so process.env.ALLOWED_ORIGINS was unset and ListParam.runtimeValue() did JSON.parse(undefined), causing SyntaxError: "undefined" is not valid JSON and blocking deploy. This change defers cors Expression resolution to request time, so the param is only read when the function runs and env is set.

### Related Issues
Resolves #1773

### Changes Made
- onRequest & onCall: When cors is an Expression (e.g. defineList(...)), stop calling .value() at definition time. Use CORS origin as a function that calls corsExpression.runtimeValue() only when the cors middleware runs (per request), so the param is resolved at request time.
- ListParam.runtimeValue(): If the param’s env var is missing or empty, throw a clear “parameter not set” error instead of letting JSON.parse(undefined) throw, so deploy-time mistakes give an actionable message.

### Testing
The fix is covered by these existing tests in spec/v2/providers/https.spec.ts:
- onRequest – cors with defineList
Test: it("should allow cors params", ...) at line 231 (test body: lines 231–266), inside describe("onRequest", ...).
Uses defineList("ORIGINS") and cors: origins in https.onRequest, sets process.env.ORIGINS before calling onRequest, then sends an OPTIONS request and asserts CORS headers. With the fix, the origin callback runs at request time and the test passes.
- onCall – cors with defineList
Test: it("should allow cors params", ...) at line 488 (test body: lines 488–508), inside describe("onCall", ...).
Uses origins = defineList("ORIGINS") (with process.env.ORIGINS set in beforeEach) and https.onCall({ cors: origins }, () => 42), then sends a request and checks CORS. With the fix, the param is resolved at request time and the test passes.
No new tests were added; both tests above were already in the repo and continue to pass with the deferred resolution.

### Additional Notes
The cors package supports origin as a function (reqOrigin, callback) => void, which is used for deferred resolution.
ListParam extends Param extends Expression, so opts.cors instanceof Expression is true for defineList(...) and the deferred path is used.
The enableCors debug flag is still honored in both onRequest and onCall when cors is an Expression.


src/v2/providers/https.ts
```diff
@@ -434,30 +506,38 @@ export function onCall<T = any, Return = any | Promise<any>, Stream = unknown>(
...
-    cors = true;
-  }
-
-  let origin = isDebugFeatureEnabled("enableCors") ? true : cors;
-  // Arrays cause the access-control-allow-origin header to be dynamic based
-  // on the origin header of the request. If there is only one element in the
-  // array, this is unnecessary.
-  if (Array.isArray(origin) && origin.length === 1) {
-    origin = origin[0];
+    let origin = isDebugFeatureEnabled("enableCors") ? true : cors;
```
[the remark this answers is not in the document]
cabljac: @HassanBahati

src/params/types.ts
```diff
@@ -634,9 +634,24 @@ export class ListParam extends Param<string[]> {
 
   /** @internal */
   runtimeValue(): string[] {
-    const val = JSON.parse(process.env[this.name]);
-    if (!Array.isArray(val) || !(val as string[]).every((v) => typeof v === "string")) {
-      return [];
+    const raw = process.env[this.name];
+    if (!raw) {
```
inlined: Would it be better to return an empty list if the environment variable is missing? I might be steering into the idea of params with default values soon

src/params/types.ts
```diff
@@ -634,9 +634,24 @@ export class ListParam extends Param<string[]> {
...
+      throw new Error(
+        `Parameter "${this.name}" is not set. Set it in .env or .env.local, or ensure the Functions runtime has provided it.`
```
inlined: You can also have .env.project or .env.alias. Maybe we should just say "a dotenv file" and point to documentation?

src/v2/providers/https.ts
```diff
@@ -286,6 +286,70 @@ export interface CallableFunction<T, Return, Stream = unknown> extends HttpsFunc
...
+  options: { respectCorsFalse?: boolean; corsOpt?: unknown }
+): NonNullable<cors.CorsOptions["origin"]> {
+  return (reqOrigin: string | undefined, callback: (err: Error | null, allow?: boolean | string) => void) => {
+    if (isDebugFeatureEnabled("enableCors") && (!options.respectCorsFalse || options.corsOpt !== false)) {
+      callback(null, true);
+      return;
+    }
+    let resolved: string | string[];
+    try {
+      resolved = corsExpression.runtimeValue();
```
inlined, line 340: Since you wrote this I added a helper valOf that you can use if you want

IzaakGough (commented):
lgtm

IzaakGough:
Closing in favour of #1887

PR #1869, opened 2026-04-09
feat: accept Expression for v2 RTDB instance option

### LINKED ISSUE #1613: Realtime instance can not be configured using params for onValueWritten functions
### Version info

**node:** `v20.10.0`

**firebase-functions:** `^5.1.1`

**firebase-tools:** `13.16.0`

**firebase-admin:** `^12.4.0`

### Test case

Setup a Firebase Cloud Function with the following code:

```
exports.foo = onValueWritten(
  {
    ref: '/foo/bar,
    instance: defineString('REALTIME_INSTANCE'),
  },
  async (change) => {
    console.log('hey!);
  },
);
``` 

###  Steps to reproduce

Try to compile following the example code, `npm run build`.

### Expected behavior

We should be able to deploy functions with Realtime instance as a parameter. This is very needed in projects with different Realtime instance per environment. For example, we have to different Realtime instances, one for production and another for staging, and we want to have our functions deployed twice, once per instance.

### Actual behavior

As the `ReferenceOptions<Ref>` object only accepts `string` as type for the `instance` value, it's impossible to configure it based on the environment.

No, as it's impossible to build the codebase.

IzaakGough:
Fixes https://github.com/firebase/firebase-functions/issues/1613

## Summary

Allow the v2 Realtime Database trigger `instance` option to accept a deploy-time param (`Expression<string>`), not just a plain string.
This lets functions target different RTDB instances per environment without triggering the deployment-time `.value()` warning.

## Problem

`ReferenceOptions.instance` was typed and handled as a plain `string`. 

Passing `defineString('REALTIME_INSTANCE')` for `instance` either failed type-checking or caused the SDK to eagerly resolve the param during deploy, which emitted the standard `.value()` deployment warning.

## Solution

- Widen `ReferenceOptions.instance` to `string | Expression<string>`.
- Preserve expression-backed instances through endpoint generation instead of resolving them in `getOpts()`.
- Serialize expression-backed instances into the manifest as CEL expressions.
- Resolve the expression only at runtime when the trigger needs to construct its path pattern and event params.

## Testing

- Added unit coverage for `getOpts()` preserving an `Expression<string>`.
- Added endpoint/manifest coverage for expression-backed `instance` values.
- Added a test that verifies no deployment-time warning is emitted when defining the trigger with a param-backed instance.
- Added a runtime test confirming instance resolution and param extraction still work.


src/v2/providers/database.ts
```diff
@@ -488,7 +488,10 @@ export function getOpts(referenceOrOpts: string | ReferenceOptions) {
     opts = {};
   } else {
     path = normalizePath(referenceOrOpts.ref);
-    instance = referenceOrOpts.instance || "*";
+    instance =
+      referenceOrOpts.instance instanceof Expression
+        ? referenceOrOpts.instance.value()
```
CorieW: Does this not cause a warning log during deployment?

IzaakGough: Yes you're right, thanks for spotting this. I've changed the RTDB trigger path to preserve Expression<string> through endpoint generation and only resolve it at runtime, so the warning no longer appears during deploy. 

All tests pass and I did a manual deployment test with a dummy function, seems to work fine.

CorieW (approved):
lgtm

src/v2/providers/database.ts
```diff
@@ -641,13 +647,18 @@ export function onChangedOperation<Ref extends string>(
...
   const pathPattern = new PathPattern(path);
-  const instancePattern = new PathPattern(instance);
+  const instancePattern = instance instanceof Expression ? instance : new PathPattern(instance);
 
   // wrap the handler
   const func = (raw: CloudEvent<unknown>) => {
     const event = raw as RawRTDBCloudEvent;
     const instanceUrl = getInstance(event);
-    const params = makeParams(event, pathPattern, instancePattern) as unknown as ParamsOf<Ref>;
+    const resolvedInstancePattern = resolveInstancePattern(instancePattern);
```
cabljac: some duplication here, can we clean up

cabljac (dismissed):

src/v2/providers/database.ts
```diff
@@ -676,13 +687,18 @@ export function onOperation<Ref extends string>(
...
+    const params = makeParams(
+      event,
+      pathPattern,
+      resolvedInstancePattern
+    ) as unknown as ParamsOf<Ref>;
```
cabljac: is this casting necessary - can we double check

IzaakGough: this casting doesn't appear to be necessary; builds fine with just `as ParamsOf<Ref>`

ajperel: I appreciate this getting cleaned up!

ajperel (approved):

PR #1871, opened 2026-04-15
fix(logger): preserve all fields when logging object with message property

### LINKED ISSUE #1707: logging object with 'message' property
### [REQUIRED] Version info

**node:** v20.9.0
**firebase-functions:** 6.3.1
**firebase-tools:** 14.5.1
**firebase-admin:** 13.4.0

### [REQUIRED] Test case

```ts
logger.info({message: "Hello from Firebase v2!", test: "hello"});
// Output: "Hello from Firebase v2!"
// Expected Output: {"message": "Hello from Firebase v2!", "test": "hello"}
```

1. Use the test case code
2. Deploy firebase function

Log the object:
```json
{"message": "Hello from Firebase v2!", "test": "hello"}
```

Image

Yep

kyungseopk1m:
## Problem

When a plain object containing a `message` property is passed as the sole argument to logger methods, Cloud Logging extracts the `message` field as the primary display text and hides other fields from the default log view.

```ts
logger.info({ message: "Hello from Firebase!", test: "hello" });
// Cloud Logging shows: "Hello from Firebase!"
// "test" field not visible in default view
```

Fixes #1707

## Solution

Skip structured data extraction when the only argument is a plain object containing a `message` property. This lets `util.format` render the full object, making all fields visible in the Cloud Logging summary line.

```ts
logger.info({ message: "Hello from Firebase!", test: "hello" });
// Output: { severity: "INFO", message: "{ message: 'Hello from Firebase!', test: 'hello' }" }
// Cloud Logging shows the full object — all fields visible
```

Objects without a `message` property continue to work as structured data:

```ts
logger.info({ foo: "bar" });
// Output: { severity: "INFO", foo: "bar" }  ← unchanged
```

Users who need structured logging with an explicit `message` field can use `logger.write()` directly.

## Testing

- Updated existing test to reflect new behavior
- Added reproduction test for issue #1707
- All 878 tests passing

### DIFF
--- a/src/logger/index.ts
+++ b/src/logger/index.ts
@@ -161,12 +161,17 @@ function entryFromArgs(severity: LogSeverity, args: any[]): LogEntry {
   let entry = {};
   const lastArg = args[args.length - 1];
   if (lastArg && typeof lastArg === "object" && lastArg.constructor === Object) {
-    // If the only argument is a plain object containing a 'message' property,
-    // don't extract it as structured data. The 'message' field is treated as a
-    // special field by Cloud Logging, which suppresses other fields from the
-    // default log view. Let util.format render the full object instead so all
-    // fields are visible in the log summary.
-    if (!(args.length === 1 && "message" in lastArg)) {
+    if (args.length === 1 && "message" in lastArg) {
+      // The only argument is a plain object with a 'message' property.
+      // Cloud Logging treats 'message' as a special field that becomes the
+      // primary display text, which can make other fields invisible in the
+      // default log view. To ensure all fields remain visible and queryable,
+      // preserve non-message fields as structured data and format the full
+      // object as the message string.
+      const { message: _, ...rest } = args[0];
+      entry = rest;
+      // Leave args intact so format() renders the full object (all fields visible).
+    } else {
       entry = args.pop();
     }
   }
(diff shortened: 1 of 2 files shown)


src/logger/index.ts
```diff
@@ -161,7 +161,19 @@ function entryFromArgs(severity: LogSeverity, args: any[]): LogEntry {
   let entry = {};
   const lastArg = args[args.length - 1];
   if (lastArg && typeof lastArg === "object" && lastArg.constructor === Object) {
-    entry = args.pop();
+    if (args.length === 1 && "message" in lastArg) {
+      // The only argument is a plain object with a 'message' property.
```
cabljac, line 165: This changes the logging output from structured `jsonPayload` fields to a stringified object representation.

Users relying on Cloud Logging queries against individual fields (e.g. filtering on foo) would lose that capability. The current behaviour is intentional: Cloud Logging uses message as the display text, but all other fields remain queryable in jsonPayload.

This would be a breaking change for users depending on that.

before this change:

```typescript
logger.info({ message: "Hello", foo: "bar" });
```
The logger takes the object as-is and writes it as the JSON log entry. Cloud Logging receives:
```json
{ "severity": "INFO", "message": "Hello", "foo": "bar" }
```

After this PR:
```json
{ "severity": "INFO", "message": "{ message: 'Hello', foo: 'bar' }", "foo": "bar" }
```

they're not invisible, they're just not in the summary line. They should be fully present in jsonPayload. I'm going to close the original issue as it's working as intended i think

kyungseopk1m: @cabljac Thanks for the detailed explanation! I now understand that the fields are fully present in `jsonPayload` and queryable - the summary line display is a Cloud Logging UI concern rather than an SDK issue. Closing this PR. Appreciate you taking the time to review.

PR #1874, opened 2026-04-20
fix(v2): validate timeoutSeconds per trigger type

### LINKED ISSUE #1737: Cannot Deploy Firestore Trigger Function with timeoutSeconds: 3600 (Error: Max 540s)
### Related issues

none

### [REQUIRED] Version info

**node:** v22.14.0

**firebase-functions:** 6.4.0

**firebase-tools:** 14.17.0

**firebase-admin:** 12.6.0

When I try to deploy a Firestore trigger function with `timeoutSeconds` set to `3600`, I get a deployment error.

```typescript
module.exports.onwrite = onDocumentWritten(
  {
    document: 'segments/{id}',
    timeoutSeconds: 3600,
  },
  async event => {
    console.log('test')
  }
)
```

According to the official documentation, it should be possible to deploy successfully with a value of 3600.

Documentation: https://firebase.google.com/docs/functions/quotas#time_limits:~:text=60%20minutes%20for%20HTTP%20functions.%0A60%20minutes%20for%20event%2Ddriven%20functions.
https://cloud.google.com/functions/quotas#time_limits

It occurs the error below.

```
⚠  functions: Request to https://cloudfunctions.googleapis.com/v2/projects/[project id]/locations/[region]/functions?functionId=firestore-segment-onwrite had HTTP Error: 400, The timeout for functions with an event trigger cannot exceed 540 seconds.
```

No

kyungseopk1m:
Fixes #1737

### Description

The v2 SDK currently accepts any `timeoutSeconds` value and only the Cloud Functions control plane rejects invalid ones at deploy time with errors like `cannot exceed 540 seconds`. v1 has had `assertRuntimeOptionsValid` for a long time but v2 was missing the equivalent, so users only find out they configured the wrong limit after a multi-minute deploy round-trip.

This PR adds per-trigger-type validation that throws at function-definition time (or at manifest-extraction time for providers whose `__endpoint` is a lazy getter, such as `storage`).

Limits enforced:

- **event-handling** — 540s (firestore, database, pubsub, storage, scheduler, eventarc, testLab, remoteConfig, alerts, dataconnect)
- **HTTPS and callable** — 3600s (`onRequest`, `onCall`, `onCallGenkit`, identity, dataconnect/graphql, ai)
- **task queue** — 1800s (`onTaskDispatched`)

Implementation:

- New internal helper `assertTimeoutSecondsValid(opts, kind)` and `MAX_{EVENT,HTTPS,TASK}_TIMEOUT_SECONDS` constants in `src/v2/options.ts`.
- `optionsToEndpoint` and `optionsToTriggerAnnotations` gained an optional `kind: TimeoutKind` argument. When omitted they behave exactly as before — validation is opt-in at the call site, so existing consumers of these helpers are unaffected.
- Each v2 provider passes its own `kind` when converting the function-specific options. Global-option calls stay unannotated so they keep working exactly as before (no regression for a project that sets `timeoutSeconds` globally on a function type that can tolerate it).
- `Expression<number>`, `RESET_VALUE`, and `undefined` are deliberately skipped. The SDK cannot know the concrete value for parametrized expressions, and `RESET_VALUE` is the user explicitly asking us to clear the global.
- `onCallGenkit` delegates to `onCall(opts, …)` and is therefore covered transitively.

Error messages follow the v1 shape but include the per-kind label:

```
timeoutSeconds must be between 0 and 540 for event-handling functions. Got 3600.
timeoutSeconds must be between 0 and 3600 for HTTPS and callable functions. Got 3601.
timeoutSeconds must be between 0 and 1800 for task queue functions. Got 1801.
```

### Tests

- `spec/v2/options.spec.ts` — 13 unit tests for `assertTimeoutSecondsValid` covering each kind's limit, negative values, `Expression`, `RESET_VALUE`, global fallback, function-level override, and function-level `RESET_VALUE` clearing an out-of-range global. Plus 6 integration tests for `optionsToEndpoint` / `optionsToTriggerAnnotations` (omitted kind = no-op for backwards compatibility, provided kind = enforced).
- Per-provider regressions:
  - `spec/v2/providers/https.spec.ts` — `onRequest` and `onCall` both reject 3601s.
  - `spec/v2/providers/pubsub.spec.ts` — function-level 3600s and a global 3600s (via `setGlobalOptions`) are both rejected.
  - `spec/v2/providers/tasks.spec.ts` — 1801s rejected.
  - `spec/v2/providers/storage.spec.ts` — validation fires when `__endpoint` is accessed (lazy-getter path).

Full suite: **900 / 900 passing**, `npm run build` clean, `npm run lint` 0 errors on the touched files, `prettier --check` clean.

### Backwards compatibility

- The new `kind` argument is optional. Calls to `optionsToEndpoint` / `optionsToTriggerAnnotations` that don't pass it keep their current behavior, so downstream code that reaches into these helpers is unaffected.
- In-range timeouts (0–540 for events, 0–3600 for HTTPS, 0–1800 for tasks) continue to work identically.
- `0` is still accepted, matching v1's `assertRuntimeOptionsValid`. Enforcing the documented 1s minimum can be a separate follow-up if desired.
- Users who were silently shipping invalid timeouts will now see the error while running locally instead of at deploy time — this is the intended fix.

### Code sample

Before: this deploys, hits `gcloud beta functions deploy`, and fails after ~30 seconds:

```ts
import { onDocumentWritten } from "firebase-functions/v2/firestore";

export const fn = onDocumentWritten(
  { document: "users/{id}", timeoutSeconds: 3600 },
  (event) => { /* ... */ },
);
```

After: fails immediately when the function module is loaded with:

```
Error: timeoutSeconds must be between 0 and 540 for event-handling functions. Got 3600.
```

### DIFF
--- a/spec/v2/providers/https.spec.ts
+++ b/spec/v2/providers/https.spec.ts
@@ -336,6 +336,14 @@ describe("onRequest", () => {
     await runHandler(func, req);
     expect(hello).to.equal("world");
   });
+
+  it("rejects timeoutSeconds above the 3600s HTTPS limit", () => {
+    expect(() =>
+      https.onRequest({ timeoutSeconds: 3601 }, (_req, res) => {
+        res.end();
+      })
+    ).to.throw(/between 0 and 3600 for HTTPS and callable functions/);
+  });
 });
 
 describe("onCall", () => {
@@ -603,6 +611,12 @@ describe("onCall", () => {
     expect(hello).to.be.undefined;
     await runHandler(func, req);
     expect(hello).to.equal("world");
+  });
+
+  it("rejects timeoutSeconds above the 3600s HTTPS limit", () => {
+    expect(() => https.onCall({ timeoutSeconds: 3601 }, () => 42)).to.throw(
+      /between 0 and 3600 for HTTPS and callable functions/
+    );
   });
 
   describe("authPolicy", () => {
--- a/spec/v2/providers/pubsub.spec.ts
+++ b/spec/v2/providers/pubsub.spec.ts
@@ -134,6 +134,19 @@ describe("onMessagePublished", () => {
     expect(res).to.equal("input");
   });
 
+  it("rejects timeoutSeconds above the 540s event-handler limit", () => {
+    expect(() =>
+      pubsub.onMessagePublished({ topic: "topic", timeoutSeconds: 3600 }, () => 42)
+    ).to.throw(/between 0 and 540 for event-handling functions/);
+  });
+
+  it("rejects a global timeoutSeconds above the 540s event-handler limit", () => {
+    options.setGlobalOptions({ timeoutSeconds: 3600 });
+    expect(() => pubsub.onMessagePublished("topic", () => 42)).to.throw(
+      /between 0 and 540 for event-handling functions/
+    );
+  });
+
   it("should parse pubsub messages", async () => {
     let json: unknown;
     const messageJSON = {
--- a/src/v2/options.ts
+++ b/src/v2/options.ts
@@ -42,6 +42,31 @@ import * as logger from "../logger";
 export { RESET_VALUE } from "../common/options";
 
 /**
+ * Maximum timeout in seconds for event-handling functions (e.g. Firestore,
+ * Realtime Database, Pub/Sub, Storage, Eventarc, alerts, scheduler).
+ * @internal
+ */
+export const MAX_EVENT_TIMEOUT_SECONDS = 540;
+
+/**
+ * Maximum timeout in seconds for HTTPS and callable functions.
+ * @internal
+ */
+export const MAX_HTTPS_TIMEOUT_SECONDS = 3600;
+
+/**
+ * Maximum timeout in seconds for Task Queue functions.
+ * @internal
+ */
+export const MAX_TASK_TIMEOUT_SECONDS = 1800;
+
+/**
+ * Function category used to pick a `timeoutSeconds` upper bound.
+ * @internal
+ */
+export type TimeoutKind = "event" | "https" | "task";
+
+/**
  * List of all regions supported by Cloud Functions (2nd gen).
  */
 export type SupportedRegion =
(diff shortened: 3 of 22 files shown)


IzaakGough (commented):
Hi @kyungseopk1m, thank you for contributing! This looks good overall. The only issue I found is the failing lint check from CHANGELOG.md formatting. Could you please fix that?

kyungseopk1m:
@IzaakGough Thanks for the review! Fixed in 750b2e91 — formatted CHANGELOG.md with prettier.

spec/v2/providers/pubsub.spec.ts
```diff
@@ -134,6 +134,19 @@ describe("onMessagePublished", () => {
...
   });
 
+  it("rejects timeoutSeconds above the 540s event-handler limit", () => {
+    expect(() =>
+      pubsub.onMessagePublished({ topic: "topic", timeoutSeconds: 3600 }, () => 42)
+    ).to.throw(/between 0 and 540 for event-handling functions/);
+  });
+
+  it("rejects a global timeoutSeconds above the 540s event-handler limit", () => {
+    options.setGlobalOptions({ timeoutSeconds: 3600 });
```
cabljac, line 144: I think we might need to reset globals in an afterEach otherwise this will pollute other tests

src/v2/options.ts
```diff
@@ -334,13 +359,57 @@ export interface EventHandlerOptions extends Omit<GlobalOptions, "enforceAppChec
...
+ * `Expression`, `RESET_VALUE`, and `undefined` are skipped — only literal
+ * numbers are checked.
+ * @internal
+ */
+export function assertTimeoutSecondsValid(
+  opts: GlobalOptions | EventHandlerOptions | HttpsOptions | undefined,
+  kind: TimeoutKind
+): void {
+  const timeoutSeconds = opts?.timeoutSeconds ?? getGlobalOptions().timeoutSeconds;
+  if (typeof timeoutSeconds !== "number") {
```
cabljac, line 376: if `timeoutSeconds` is `NaN` here, i think we might have an issue, i think it might slip through?

spec/v2/providers/https.spec.ts
```diff
@@ -336,6 +336,14 @@ describe("onRequest", () => {
     await runHandler(func, req);
     expect(hello).to.equal("world");
   });
+
+  it("rejects timeoutSeconds above the 3600s HTTPS limit", () => {
```
cabljac, line 340: I like that we're testing some of the providers, but I think we need more direct test of the wiring across providers

a parameterized suite like "each v2 provider rejects out-of-range timeout" would be a cheap way to get coverage

src/v2/options.ts
```diff
@@ -334,13 +359,57 @@ export interface EventHandlerOptions extends Omit<GlobalOptions, "enforceAppChec
...
+ * Validate that `timeoutSeconds` (resolved with global option fallback) does
+ * not exceed the maximum allowed value for the given function kind. Throws a
+ * plain `Error` matching the v1 `assertRuntimeOptionsValid` shape so the
+ * problem surfaces at function-definition time instead of at deploy time.
+ * `Expression`, `RESET_VALUE`, and `undefined` are skipped — only literal
+ * numbers are checked.
+ * @internal
+ */
+export function assertTimeoutSecondsValid(
+  opts: GlobalOptions | EventHandlerOptions | HttpsOptions | undefined,
```
cabljac, line 372: nit: I don't know if we need the `| undefined`

cabljac (changes requested):
Thanks for this PR! I've added some comments. @IzaakGough we may be able to pick this up

IzaakGough:
Closing since https://github.com/firebase/firebase-functions/pull/1877 has been merged.

PR #1877, opened 2026-04-23
fix(v2): validate timeoutSeconds per trigger type (addition) 

### LINKED ISSUE #1737: Cannot Deploy Firestore Trigger Function with timeoutSeconds: 3600 (Error: Max 540s)

IzaakGough:
Supercedes https://github.com/firebase/firebase-functions/pull/1877
Fixes https://github.com/firebase/firebase-functions/issues/1737
## Summary

- Adds v2 `timeoutSeconds` validation at function-definition or manifest-extraction time using trigger-specific limits: 0-540s for event handlers, 0-3600s for HTTPS/callable functions, and 0-1800s for task queues.
- Rejects invalid literal timeout values, including negative numbers, non-finite numbers, `null`, and invalid non-number types, while preserving `Expression` and `RESET_VALUE` support.
- Wires the validation through v2 providers so out-of-range timeout values fail before deploy.

## Testing

- `npm test -- --grep "assertTimeoutSecondsValid|optionsToEndpoint timeout validation|optionsToTriggerAnnotations timeout validation|v2 provider timeout validation|timeoutSeconds above"`
- `npx prettier --check CHANGELOG.md src/v2/options.ts spec/v2/options.spec.ts spec/v2/providers/timeout.spec.ts spec/v2/providers/https.spec.ts spec/v2/providers/storage.spec.ts spec/v2/providers/pubsub.spec.ts spec/v2/providers/tasks.spec.ts`
- `npm run build`

## Notes

- `timeoutSeconds: 0` remains accepted to match existing runtime option validation behavior.
- Blocking auth timeouts may deserve a follow-up limit that is stricter than the general HTTPS/callable limit. Possible follow-ups: add a dedicated `blocking` timeout kind with the platform-specific cap, validate blocking providers before calling `optionsToEndpoint`, or keep the current HTTPS kind until the backend/CLI constraint is confirmed.


CorieW (commented):
Missing tests for `timeoutSeconds` being `null`

CorieW (commented):
Add PR description

src/v2/options.ts
```diff
@@ -334,13 +359,71 @@ export interface EventHandlerOptions extends Omit<GlobalOptions, "enforceAppChec
...
+    case "task":
+      max = MAX_TASK_TIMEOUT_SECONDS;
+      label = "task queue";
+      break;
+    default:
+      max = MAX_EVENT_TIMEOUT_SECONDS;
+      label = "event-handling";
+      break;
+  }
+  if (timeoutSeconds <= 0 || timeoutSeconds > max) {
```
cabljac: can we double check the behaviour before with timeoutSeconds = 0

https://github.com/firebase/firebase-functions/blob/master/src/v1/function-builder.ts#L63

seems to use `<0` here

cabljac: dont think this is wrong (as this is v2) but we should mention in CHANGELOG i think

src/v2/providers/identity.ts
```diff
@@ -325,7 +325,7 @@ export function beforeOperation(
 
   /** Endpoint */
   const baseOptsEndpoint = options.optionsToEndpoint(options.getGlobalOptions());
-  const specificOptsEndpoint = options.optionsToEndpoint(opts);
+  const specificOptsEndpoint = options.optionsToEndpoint(opts, "https");
```
cabljac: I have a feeling blocking timeout needs to be stricter, maybe we need a different kind added

src/v2/options.ts
cabljac: worth covering this edge case with a test asserting the behaviour too

IzaakGough: yes you are right, 0 is allowed. I will change this back (and the tests accordingly). I'll also add the new test for 0.

src/v2/providers/identity.ts
cabljac: We can catch up on this tomorrow @IzaakGough  - i think the timeouts are shorter for identity perhaps. I need to dig into this more

IzaakGough: @cabljac I have been working on this. You were right, the max timeout for identity blocking functions is 7 seconds per the [docs](https://firebase.google.com/docs/auth/extend-with-blocking-functions#:~:text=Your%20function%20must%20respond%20within%207%20seconds%2E%20After%207%20seconds%2C%20Firebase%20Authentication%20returns%20an%20error%2C%20and%20the%20client%20operation%20fails)

I have added a new identity kind and tests.

src/v2/options.ts
ajperel: I used this PR to test some code review prompts I'm working on. AI on that prompt also called out (same as gemini review bot) that the minimum timeout is 1 not 0. 

I noticed you resolved the Gemini review comment. Are you confident the minimum timeout is 0? Why? (I honestly don't know) If it is 0 it'd be good to update this part of the documentation so this file stays consistent with itself.

IzaakGough: I wasn't certain. I worked under the assumption that the minimum timeout for v2 functions was the same as v1 (i.e. 0 seconds).

I double checked and the docs definitely say that the minimum timeout **for v1 functions** is 0 [here](https://firebase.google.com/docs/reference/functions/firebase-functions#:~:text=timeout%20for%20the%20function%20in%20seconds%2C%20possible%20values%20are%200%20to%20540%2E%203%2E%20failurePolicy)
It seems like the underlying cloud run request-timeout docs say that the minimum should be 1 [here](https://docs.cloud.google.com/run/docs/configuring/request-timeout#:~:text=ranging%20from%201%20to%203600%20seconds).

Perhaps we do want a minimum timeout of 1 for v2 then?

ajperel: Interesting. Your v1 reference was in Firebase docs so I went digging into that for v2 and found Global Options allows setting timeouts between 0 and 540 seconds: https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.globaloptions.md#globaloptions_interface

But I don't see how that works if the cloud run documentation is accurate that it needs to be at least 1s.

Maybe the safest thing to do is try 0 and use that if it works and 1 others. TBH, I'm not sure what behavior I'd expect for 0?

cabljac: Hi all, I went digging through the CLI to understand this:

I think 0 means "use the platform default" not a literal zero-second timeout:

[1. The CLI treats `0` as "unset" and skips during validation with an `if (!timeout) continue`](https://github.com/firebase/firebase-tools/blob/68e5671310faf5fb6f5ff755b5c439eb4d7dee48/src/deploy/functions/validate.ts#L217-L223)

[2. The value is sent to the API ](https://github.com/firebase/firebase-tools/blob/68e5671310faf5fb6f5ff755b5c439eb4d7dee48/src/gcp/cloudfunctionsv2.ts#L470-L478) as a plain protobuf integer - so when the server parses, it's identical to being omitted, so it will be falling back to the default value i think

The doc comment is an inconsistency though, i'd drop the "minimum is 1s"

IzaakGough: I have added a note about the behavior/meaning of the 0s minimum timeout, but as discussed, we still need to make a documentation change.

src/v2/options.ts
```diff
@@ -41,6 +41,37 @@ import * as logger from "../logger";
 
 export { RESET_VALUE } from "../common/options";
 
+/**
+ * Maximum timeout in seconds for event-handling functions (e.g. Firestore,
+ * Realtime Database, Pub/Sub, Storage, Eventarc, alerts, scheduler).
```
ajperel, line 47: Can we include links to the source documentation on where these max's come from. That way it's both easy to verify and slightly more discoverable if they change in the future.

src/v2/options.ts
```diff
@@ -334,13 +367,76 @@ export interface EventHandlerOptions extends Omit<GlobalOptions, "enforceAppChec
...
+    );
+  }
+  // Handle the case where timeoutSeconds is NaN
+  if (!Number.isFinite(timeoutSeconds)) {
+    throw new Error(`timeoutSeconds must be a finite number. Got ${timeoutSeconds}.`);
+  }
+
+  let max: number;
+  let label: string;
+  switch (kind) {
```
ajperel, line 406: Can we do this instead in a way that we can at the end `assertExhaustive` and make sure that if we add cases in the future we don't forget to add new cases we add in the future.

CHANGELOG.md
```diff
@@ -1 +1,2 @@
+- Validate literal `timeoutSeconds` values per v2 trigger type (0-540s for events, 0-3600s for HTTPS/callable, 0-1800s for task queues) so misconfigured values fail at function-definition or manifest-extraction time instead of at deploy time. (#1877)
```
ajperel: should we also document the separate timeout for identity functions?

ajperel (commented):
A little minor feedback but this is very close. Thanks for doing this.

IzaakGough:
@ajperel Thanks for the feedback! Made those suggested changes, let me know if you spot anything else.

ajperel (approved):

PR #1895, opened 2026-05-28
fix: preserve quoted hyphenated export names in discovery

IzaakGough:
Fixes `firebase/firebase-tools#8469`

This change fixes function discovery for quoted export names that contain hyphens, for example:

```js
const func = onRequest((req, res) => {
  res.send("ok");
});

export { func as "dummystore-bot" };
```

## Problem

The current discovery logic derives `entryPoint` from the manifest endpoint id by replacing `-` with `.`.

That works for grouped exports, where a flattened manifest id like `grouped-fn` should map back to an entry point like `grouped.fn`.

However, it is incorrect for quoted export names that literally contain hyphens. In that case, the discovered function id is something like `dummystore-bot`, but converting that to `dummystore.bot` produces the wrong runtime entry point.

As a result, downstream tooling can fail to deploy or invoke these functions correctly.

## Change

This PR updates runtime discovery to track two separate paths during stack extraction:

- the manifest endpoint id, which remains hyphen-flattened for nested/grouped exports
- the runtime `entryPoint`, which preserves the actual export access path

With this change:

- grouped exports still behave as before
- quoted export names with hyphens are preserved correctly
- discovery emits the correct `entryPoint` for downstream consumers

## Tests

This PR adds coverage for the fix in two places:

- unit/spec coverage for `extractStack`
- end-to-end bin/discovery coverage using an ESM fixture with a quoted hyphenated export name

## Companion Change In `firebase-tools`

This is the primary fix.

There is a companion PR in `firebase-tools` that updates the CLI to trust the SDK-provided `entryPoint` instead of re-deriving it from function ids, and also updates `functions:shell` to use the discovered `entryPoint`.

Companion PR: 

## Validation

Validated with:

- local `firebase-functions` discovery tests
- local `firebase-tools` follow-up changes
- a packaged end-to-end deploy flow using:
  - `npm pack` on local `firebase-functions`
  - `npm pack` on local `firebase-tools`
  - a throwaway test project deploying a quoted export named `dummystore-bot`

That deploy succeeded and the function was recognized and deployed under the correct name.

### DIFF
--- a/scripts/bin-test/sources/esm-quoted-hyphen/index.js
+++ b/scripts/bin-test/sources/esm-quoted-hyphen/index.js
@@ -1,0 +1,7 @@
+import * as functions from "firebase-functions/v1";
+
+const func = functions.https.onRequest((req, resp) => {
+    resp.status(200).send("PASS");
+});
+
+export { func as "dummystore-bot" };
--- a/scripts/bin-test/sources/esm-quoted-hyphen/package.json
+++ b/scripts/bin-test/sources/esm-quoted-hyphen/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "esm-quoted-hyphen",
+  "type": "module"
+}
--- a/scripts/bin-test/test.ts
+++ b/scripts/bin-test/test.ts
@@ -448,6 +448,23 @@ describe("functions.yaml", function () {
           extensions: {},
         },
       },
+      {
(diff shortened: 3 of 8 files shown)


IzaakGough:
Closing in favor of https://github.com/firebase/firebase-functions/pull/1896

PR #1897, opened 2026-05-28
feature/CF3-Auth-Event-Triggers

shettyvarun268:
This pull request introduces support for Firebase Authentication event triggers in Cloud Functions v2 by adding the onUserCreated and onUserDeleted functions. These new triggers leverage Eventarc to listen for user lifecycle events at both the project and tenant levels, unblocking a key migration path for developers moving from V1 to V2. To ensure a maintainable implementation, a shared helper function called makeAuthTrigger was created to handle the common logic for manifest generation and event execution wrapping, effectively reducing code duplication and keeping the implementation DRY. The PR also ensures correct deployment behavior by setting the trigger region to global in the generated manifest, aligning with Eventarc's requirements for these global events and preventing location mismatch errors during deployment. Comprehensive unit tests have been added to cover standard behavior as well as edge cases involving missing or parameterized tenant IDs, ensuring robust operation in diverse environments.


src/v2/providers/identity.ts
```diff
@@ -88,7 +92,6 @@ export interface BlockingOptions {
   /**
    * Timeout for the function in seconds, possible values are 0 to 540.
    * HTTPS functions can specify a higher timeout.
-   *
```
wandamora, line 91: nit: unnecessary line removals in the documentation?

src/v2/providers/identity.ts
```diff
@@ -371,3 +371,181 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+  /** The ID of the Identity Platform Tenant Associated with the event. If Applicable */
+  tenantId?: string;
+}
+
+/** Options for configuring a Firebase Authentication Trigger */
+export interface AuthOptions extends options.EventHandlerOptions {
+  /**
+   * The Id of the Identity Platform tenant to scope the function to.
+   * If not set, the function triggers on users across all tenants
+   * SEt to `IS_NOT_TENANT` to only trigger on users in the default
```
wandamora: nit: s/SEt/set

src/v2/providers/identity.ts
```diff
@@ -371,3 +371,181 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+  }
+  const rawAny = raw as any;
+  // Support both lowercase (CloudEvents standard) and camelCase (local testing)
+  const tenantId = rawAny.tenantid || rawAny.tenantId;
+  if (tenantId) {
+    event.tenantId = tenantId;
+  }
+  if (raw.source) {
+    // Matches both absolute paths (/projects/...) and relative paths (projects/...)
+    const match = raw.source.match(/(?:^|\/)projects\/([^\/]+)/);
```
wandamora: Consider moving this regex to a module-level constant to make the code cleaner and avoid any runtime redeclaration.

src/v2/providers/identity.ts
```diff
@@ -371,3 +371,181 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+    ...specificOptsEndpoint,
+    labels: {
+      ...baseOptsEndpoint?.labels,
+      ...specificOptsEndpoint?.labels,
+    },
+    eventTrigger: {
+      eventType,
+      eventFilters: {},
+      retry: false, // Default retry policy
+      region: "global",
```
wandamora, line 483: Can you add a comment describing why we set this to global? This could signal to developers why this is hardcoded and discourage them from trying to override this value.

src/v2/providers/identity.ts
shettyvarun268: Fixed

src/v2/providers/identity.ts
shettyvarun268: Thanks!!
Fixed

src/v2/providers/identity.ts
shettyvarun268: Good point. Moved the regex to a module-level constant (PROJECT_ID_REGEX).

src/v2/providers/identity.ts
shettyvarun268: Done. Added a comment explaining Eventarc's requirement for a global trigger region on Auth events.

wandamora (approved):
Thanks for making the changes!

spec/v2/providers/identity.spec.ts
```diff
@@ -421,6 +421,171 @@ describe("identity", () => {
...
+    });
+
+    it("should handle tenantId options", () => {
+      const func = identity.onUserCreated({ tenantId: "my-tenant" }, () => null);
+      expect(func.__endpoint.eventTrigger?.eventFilters?.tenantid).to.equal("my-tenant");
+    });
+
+    it("should handle IS_NOT_TENANT option", () => {
+      const func = identity.onUserCreated({ tenantId: identity.IS_NOT_TENANT }, () => null);
+      expect(func.__endpoint.eventTrigger?.eventFilters?.tenantid).to.equal("");
```
inlined, line 448: Just to be clear, does this mean there's a semantic difference between any of null, undefined, and ""?

Is it that unspecified listens to all tenants and "" listens to no tenant?

spec/v2/providers/identity.spec.ts
```diff
@@ -421,6 +421,171 @@ describe("identity", () => {
...
+      let called = false;
+      const func = identity.onUserCreated((event) => {
+        called = true;
+        expect(event.project).to.be.undefined;
+        return null;
+      });
+
+      const mockEvent = {
+        specversion: "1.0" as const,
+        source: "//identitytoolkit.googleapis.com/something-else",
```
inlined, line 492: This is peculiar. What are the cases for this. I clearly don't understand something about GCIP.

spec/v2/providers/identity.spec.ts
```diff
@@ -421,6 +421,171 @@ describe("identity", () => {
...
+
+      func(mockEvent);
+      expect(called).to.be.true;
+    });
+
+    it("should handle missing tenantid", () => {
+      let called = false;
+      const func = identity.onUserCreated((event) => {
+        called = true;
+        expect(event.tenantId).to.be.undefined;
```
inlined: Is it to be present but undefined or not present?

spec/v2/providers/identity.spec.ts
```diff
@@ -421,6 +421,171 @@ describe("identity", () => {
...
+        time: new Date().toISOString(),
+        data: {} as any,
+      };
+
+      func(mockEvent);
+      expect(called).to.be.true;
+    });
+
+    it("should handle Expression for tenantId", () => {
+      const param = {
```
inlined: You can also just call defineString and delete it with the private API later

src/v2/providers/identity.ts
```diff
@@ -371,3 +375,185 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+ */
+export type User = AdminUserRecord;
+
+/**
+ * The event object passed to the handler funcation for Firebase Authentication
+ * events.
+ */
+export interface AuthEvent<T> extends CloudEvent<T> {
+  /** The project identifier*/
+  project?: string;
```
inlined, line 390: nit: spaces between fields with comments.

src/v2/providers/identity.ts
```diff
@@ -371,3 +375,185 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+  /** The ID of the Identity Platform Tenant Associated with the event. If Applicable */
+  tenantId?: string;
+}
+
+/** Options for configuring a Firebase Authentication Trigger */
+export interface AuthOptions extends options.EventHandlerOptions {
+  /**
+   * The Id of the Identity Platform tenant to scope the function to.
+   * If not set, the function triggers on users across all tenants
+   * Set to `IS_NOT_TENANT` to only trigger on users in the default
```
inlined, line 401: I wonder if IS_NOT_TENANT needs to be well defined since a parameter will have to have that value type. The very very committed-to-DX option would be to create a new resource selector but I think we just need to have a default that is documented/makes sense (even if it is "")

src/v2/providers/identity.ts
```diff
@@ -371,3 +375,185 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+    ...baseOptsEndpoint,
+    ...specificOptsEndpoint,
+    labels: {
+      ...baseOptsEndpoint?.labels,
+      ...specificOptsEndpoint?.labels,
+    },
+    eventTrigger: {
+      eventType,
+      eventFilters: {},
+      retry: false, // Default retry policy
```
inlined: I'm curious: if it's the default, why does it need to be specified? Do we need to fix something elsewhere?

src/v2/providers/identity.ts
```diff
@@ -371,3 +375,185 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+      // Firebase Auth event triggers are global by nature, and Eventarc requires the trigger
+      // region to be set to "global" even if the Cloud Function is deployed regionally.
+      region: "global",
```
inlined, line 483: I wonder if the CLI should also set this? That way we're defensive against other SDKs in other languages getting it wrong? We now have automatic region setting (I think that was added in my absence?) so that should be compatible with the coding style there.

src/v2/providers/identity.ts
```diff
@@ -371,3 +375,185 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+    },
+  };
+  if (opts.tenantId !== undefined) {
+    if (opts.tenantId === IS_NOT_TENANT) {
+      endpoint.eventTrigger.eventFilters["tenantid"] = "";
```
inlined, line 488: I thought IS_NOT_TENANT already is "", making this a safe assignment in all cases?

src/v2/providers/identity.ts
```diff
@@ -371,3 +375,185 @@ export function getOpts(blockingOptions: BlockingOptions): InternalOptions {
...
+    } else {
+      endpoint.eventTrigger.eventFilters["tenantid"] = opts.tenantId as string | Expression<string>;
+    }
+  }
+  copyIfPresent(endpoint.eventTrigger, opts, "retry");
+  func.__endpoint = endpoint;
+  return func;
+}
+
+const USER_CREATED_EVENT = "google.firebase.auth.user.v2.created";
```
inlined, line 497: Nit: Space before block comments

CHANGELOG.md
```diff
@@ -1 +1,2 @@
-- fix(v1): Call onInit for schedule.onRun functions (#1801)
+fix(v1): Call onInit for schedule.onRun functions (#1801)
+Added support for Firebase Authentication event triggers in Cloud Functions v2 (`onUserCreated` and `onUserDeleted`). (#1897)
```
inlined: Leave it out of changelogs while we're testing.

inlined (commented):
Since you're creating new files you're going to need to edit package.json to map the import path to the actual file for both JS and .d.ts files. You also need to create a dummy file in /v2 (not /src/v2) due to a longstanding bug in eslint that follows actual paths instead of declared import paths. IDR if it's been fixed; last I checked and filed the bug report, the package maintainer tried to extort me to fix it 🫠

spec/v2/providers/identity.spec.ts
shettyvarun268: Thats right. Unspecified (undefined) matches the default Eventarc behavior which listens across all tenants (as well as default project users). Setting it to IS_NOT_TENANT sets the manifest filter to "", which instructs Eventarc to match only users created in the default project (no tenant). Providing a specific string scopes the trigger to that specific tenant ID.

spec/v2/providers/identity.spec.ts
shettyvarun268: This test case is primarily here for defensive programming. While standard production events will include /projects/YOUR_PROJECT_ID, local emulator events, custom tests or future Eventarc routing changes might omit the /projects/ segment in source. This test ensures our regex parsing doesn't throw a runtime error and gracefully leaves event.project as undefined if raw.source has an unexpected shape. Might not be entirely necessary to keep this since we don't really expect event arc sources to change.

spec/v2/providers/identity.spec.ts
shettyvarun268: Setting it explicitly to undefined matches the optional property definition (tenantId?: string) in the interface, while ensuring a consistent object shape for consumers who might be destructuring the payload

spec/v2/providers/identity.spec.ts
shettyvarun268: Good point. I've updated the test to use defineString('MY_TENANT') and clean up afterward with clearParams()

src/v2/providers/identity.ts
shettyvarun268: Done! Added blank lines between fields in AuthEvent

src/v2/providers/identity.ts
shettyvarun268: While false is the default behavior on the GCP side, all other v2 event provider endpoints in the SDK explicitly initialize retry: opts.retry ?? false to ensure a fully populated manifest definition before applying overrides. Leaving it explicit here keeps this consistent with the rest of the SDK

src/v2/providers/identity.ts
shettyvarun268: Okay noted. I've implemented a dedicated autheventarc service definition in firebase-tools. The CLI now defensively intercepts v2 Auth Eventarc triggers during deploy validation and automatically ensures endpoint.eventTrigger.region = 'global', matching the pattern used for Test Lab and Firebase Alerts. I'll send in another PR for that.

src/v2/providers/identity.ts
shettyvarun268: IS_NOT_TENANT is actually assigned to RESET_VALUE (which is the special { __type: 'reset' }), rather than ''. Because of this, we have to check for IS_NOT_TENANT explicitly so we can map it to '' in the manifest filters, preventing the manifest from serializing the sentinel object.

src/v2/providers/identity.ts
shettyvarun268: Done! Added blank lines before the block comments.

CHANGELOG.md
shettyvarun268: Noted. Removed it from the change logs

spec/v2/providers/identity.spec.ts
inlined: Oooof. I understand how Eventarc got there, but it's not something that's obvious and loveable. It's not your API though so I won't ask you to make their API friendlier.

spec/v2/providers/identity.spec.ts
inlined: I would strongly assume that it would require a v2 event type change, but I won't ask you to remove the code. Just maybe explain in comments that this is for defense against pathological changes we should not expect. I do have a sliiiight worry that this allows the emulator to drift from production in subtle ways rather than catastrophic failure.

spec/v2/providers/identity.spec.ts
inlined: As a rule, we don't like present-but-undefined in our codebases. It's a different kind of nullish that confuses everything. I'd prefer the test to verify that tenantId is not a key in event if that's possible. If you don't have experience with this, read up on the delete and in keywords in JS/TS

src/v2/providers/identity.ts
inlined: Shouldn't we then also be using opts.retry ?? false? Auth endpoints are one of the most important endpoints for people to use retry on because they might be getting used for compliance with privacy laws.

src/v2/providers/identity.ts
```diff
@@ -499,11 +500,16 @@ function makeAuthTrigger(
 }
 
 const USER_CREATED_EVENT = "google.firebase.auth.user.v2.created";
+
 /**
  * Handles user creation events in Firebase Authentication.
  *
  * To filter for users not associated with a tenant, use the `IS_NOT_TENANT` constant in options.
  *
+ * @beta
```
inlined, line 504: I'd honestly prefer that we just not export the functions in index.ts and not make our lives more complicated. This has been the precedence for many years.

inlined (commented):
Again, you need to
1. Create the dummy file (NO 'src')/v2/identity.js to overcome a bug in eslint that will say the import doesn't exist
2. Modify package.json to make the files nameable. As is, you won't actually be able to import the file in your pacakge.

shettyvarun268:
Regarding the dummy file, iidentity.ts is actually an already-existing provider file (created when beforeUserCreated blocking functions were added). Because of that, v2/identity.js and the package.json export mappings are already fully set up in main

src/v2/providers/identity.ts
shettyvarun268: Similar to the package.json note, export * as identity from './providers/identity' was actually established in src/v2/index.ts in a previously when blocking functions were launched, so we are just riding on the existing top-level export here.

shettyvarun268:
I've also made changes towards your other comments. For some reason I dont see an option to reply to some of the comments individually.

inlined (approved):
Sorry that I didn't catch the fact that we're editing an existing file meant my guidance on package.json exports, dummy import files for eslint, etc. were unnecessary.

PR #1911, opened 2026-06-17
fix: support encoding of special characters in path params for gen2 f…

### LINKED ISSUE #1459: Gen2: Encoding issue with special characters in path parameters
### Version info

**node:** v18.16.0
**firebase-functions:** v4.4.1
**firebase-tools:** v12.5.2
**firebase-admin:** v11.8.0

### Steps to reproduce
**This issue was not encountered in gen1, and special characters in path parameters were correctly encoded.**

- Deploy a Firebase Cloud Function Gen2 using a Firestore trigger as follows:
  ```javascript
  import { log } from "firebase-functions/logger";
  import { onDocumentUpdated } from "firebase-functions/v2/firestore";
  
  export const testPlanet = onDocumentUpdated("planetes/{planeteId}", async (event) => {
    log(event.params.planeteId);
  });
  ```
- Ensure that the `planeteId` document contains special characters such as "é", "è", "à", etc. Example: "Helvétios", "Électre".
- Observe the `planeteId` value in the deployed function

### Expected behavior

The `planeteId` value should be correctly encoded and displayed in UTF-8, preserving special characters such as "é", "è", "à", etc., in the deployed Firebase Cloud Function Gen2.

### Actual behavior

After deploying the Firebase Cloud Function Gen2, the `planeteId` value appears to be incorrectly encoded, resulting in a value like `Helvétios` becoming `HelvÃ©tios` or `Électre` becoming `Ãlectre`.

Yes, the deployment was successful without any error messages.

IzaakGough:
## Summary

Resolves #1459 

Fix Gen 2 path parameter decoding when Eventarc delivers UTF-8 mojibake instead of already-correct Unicode, so wildcard params like `Helvétios`, `中文`, and emoji are exposed correctly in `event.params`.

## Problem/Root Cause

Firestore Gen 2 triggers could receive path wildcard values with special characters mangled into mojibake such as `HelvÃ©tios` instead of `Helvétios`. The existing path param decoding only handled percent-encoded values via `decodeURIComponent`, so non-percent-encoded mojibake was passed through unchanged.

## Solution/Changes

Update the shared path pattern decoder to:

- keep the existing `decodeURIComponent` behavior for properly encoded values
- preserve malformed percent-encoded input
- detect byte sequences that look like UTF-8 mojibake and re-decode them as UTF-8 before returning path params

Added shared regression coverage for accented Latin text, CJK, emoji, Hindi, Arabic, already-correct Unicode input, and multi-segment captures.

## Testing

- `npm test`
- Manual verification in a sample Gen 2 Firestore trigger with accented Latin text, emoji, CJK, Hindi, and Arabic path params


src/common/utilities/path-pattern.ts
```diff
@@ -36,13 +36,36 @@ export function trimParam(param: string) {
...
+
+  try {
+    return new TextDecoder("utf-8", { fatal: true }).decode(Uint8Array.from(bytes));
+  } catch (_e) {
+    return value;
+  }
+}
+
 /** @internal */
 export function tryDecodeURIComponent(uri: string): string {
```
CorieW, line 61: Would the following work?
```ts
export function tryDecodeURIComponent(uri: string): string {
  try {
    return decodeURIComponent(uri);
  } catch (_e) {
    return tryDecodeUtf8Mojibake(decoded);
  }
}
```

IzaakGough: yeah that looks cleaner

IzaakGough:
#### AI just made a good point:

It treats some raw strings as “obviously broken UTF-8 mojibake” when they are not obviously broken.

Current code in `src/common/utilities/path-pattern.ts`:
- runs `decodeURIComponent`
- then always runs `tryDecodeUtf8Mojibake`
- if string matches byte-looking pattern, it reinterprets chars as UTF-8 bytes

Bug:
same observed string can mean 2 different things.

Example:
- `Ã©` might be:
  - legit literal path segment `Ã©`
  - broken form of intended `é`

Code cannot know which one user meant.
Current implementation always picks “broken UTF-8” and converts:
- `Ã©` -> `é`
- `Â£` -> `£`
- `Â©` -> `©`
- `fooÃ©bar` -> `fooébar`

So ambiguity is:
- fix real mojibake from issue `#1459`
- but corrupt legit IDs/path segments that happen to look like mojibake bytes

Why impossible in current layer:
- `PathPattern` only sees final string
- no extra metadata says “this came in broken”
- Firestore/RTDB/Data Connect current param path gives no trusted second source

So current implementation has no safe rule for:
- repair all mojibake-like strings
vs
- preserve all raw strings

Tests prove repair cases, but miss legit ambiguous cases.

IzaakGough:
Closing in favor of https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/780. This SDK is not the correct place for fixing #1459.

PR #1915, opened 2026-06-25
feature: functions can now declare lifecycle hooks

wandamora:
### Description

This PR adds support for declaring codebase lifecycle hooks in the Firebase Functions SDK. Developers can now register actions (`afterInstall` and `afterUpdate`) that are automatically executed post-deployment.

### 1. Hook Registrations (`src/lifecycle/index.ts`)
- Added new top-level entrypoints and types under `firebase-functions/lifecycle`:
  - `afterInstall(action: LifecycleAction)`: Registers a hook to run when resources in the codebase are deployed for the first time.
  - `afterUpdate(action: LifecycleAction)`: Registers a hook to run when resources in the codebase are updated.
- Supported action types:
  - `task`: Triggers a task queue function with an optional payload `body`.
  - `call`: Calls a function (callable trigger) with optional `params`.
  - `http`: Triggers a HTTP endpoint via request `url`/`function`, `method`, and `body`.
- Ensures only one instance of `afterInstall` and `afterUpdate` can be registered per codebase at runtime.

### 2. Stack Manifest Parsing (`src/runtime/loader.ts`, `src/runtime/manifest.ts`)
- Added `ManifestLifecycleAction` and updated `ManifestStack` types to include the `lifecycleHooks` field.
- Updated `loadStack` inside `loader.ts` to capture any declared lifecycle hooks from global symbols during function discovery, mapping them into the compiled JSON/wire format returned to `firebase-tools`.
- Updated `stackToWire` serialization helper to recursively resolve CEL expressions and reset values inside `lifecycleHooks` properties.

### 3. Tests
- Created `spec/lifecycle/lifecycle.spec.ts` to test hook registration and duplicate hook prevention.
- Added tests in `spec/runtime/loader.spec.ts` verifying that `afterInstall` and `afterUpdate` declarations are correctly discovered and serialized into the `ManifestStack`.
### Code sample

```
export const runInitialSetup = onTaskDispatched(async (request) => {
  await performDatabaseSeeding(request.data);
});

// The CLI automatically discovers this call and queues the task after first-time deploy
afterInstall({
  task: {
    function: "runInitialSetup",
    body: { data: "default_catalog" }
  }
});
````

### DIFF
--- a/src/lifecycle/index.ts
+++ b/src/lifecycle/index.ts
@@ -34,18 +34,24 @@ export interface CallAction {
   params?: Record<string, unknown>;
 }
 
-export interface HttpAction {
-  function?: TargetFunction;
-  url?: string | Expression<string>;
-  method?: "GET" | "POST" | "PUT" | "DELETE";
-  body?: unknown;
-}
+export type HttpAction =
+  | {
+      function: TargetFunction;
+      url?: never;
+      method?: "GET" | "POST" | "PUT" | "DELETE";
+      body?: unknown;
+    }
+  | {
+      url: string | Expression<string>;
+      function?: never;
+      method?: "GET" | "POST" | "PUT" | "DELETE";
+      body?: unknown;
+    };
 
-export interface LifecycleAction {
-  task?: TaskAction;
-  call?: CallAction;
-  http?: HttpAction;
-}
+export type LifecycleAction =
+  | { task: TaskAction; call?: never; http?: never }
+  | { call: CallAction; task?: never; http?: never }
+  | { http: HttpAction; task?: never; call?: never };
 
 const majorVersion =
   // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
--- a/src/lifecycle/index.ts
+++ b/src/lifecycle/index.ts
@@ -1,0 +1,104 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2026 Firebase
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import { Expression } from "../params";
+
+export type TargetFunction = string | Expression<string>;
+
+export interface TaskAction {
+  function: TargetFunction;
+  body?: Record<string, unknown>;
+}
+
+export interface CallAction {
+  function: TargetFunction;
+  params?: Record<string, unknown>;
+}
+
+export interface HttpAction {
+  function?: TargetFunction;
+  url?: string | Expression<string>;
+  method?: "GET" | "POST" | "PUT" | "DELETE";
+  body?: unknown;
+}
+
+export interface LifecycleAction {
+  task?: TaskAction;
+  call?: CallAction;
+  http?: HttpAction;
+}
+
+const majorVersion =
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/lifecycle/index.ts [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import { Expression } from "../params";
+
+export type TargetFunction = string | Expression<string>;
```
inlined: I don't think a function needs to be an expression. I'm also curious whether anything needs to be parameterized except for authorization which is automatic.

src/lifecycle/index.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+export interface CallAction {
+  function: TargetFunction;
+  params?: Record<string, unknown>;
+}
+
+export type HttpAction =
+  | {
+      function: TargetFunction;
```
inlined: What does it mean to pass a function and a URL? I would expect this type to be:

```typescript
export type HttpAction = ({ function: TargetFunction} | { url: string }) & {
  method, headers, body ...
}
```

src/runtime/loader.ts
```diff
@@ -209,5 +211,36 @@ export async function loadStack(functionsDir: string): Promise<ManifestStack> {
   if (declaredRoles.size > 0) {
     stack.requiredRoles = Array.from(declaredRoles);
   }
+
+  const hooks = Object.entries(declaredLifecycleHooks);
+  if (hooks.length > 0) {
+    stack.lifecycleHooks = {};
```
inlined: Why can't this all be replaced with stack.lifecycleHooks = hooks? Aren't they the wire protocol already?

inlined (commented):
Don't forget that when we publish this it should also go in index.ts
You're going to need to add the fake empty file to avoid eslint problems (v2 not src/v2)

inlined (approved):
Sorry, forgot to set approve. I trust that this feedback is straightforward enough to fix if you also agree about the expression thing

src/lifecycle/index.ts
wandamora: I think your comments called out Jetski overcomplicating things. You're right that Expression isn't needed for any of these fields. Fixed.

src/lifecycle/index.ts
wandamora: Updated

src/runtime/loader.ts
wandamora: Updated.

src/lifecycle/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  headers?: Record<string, string>;
+  body?: unknown;
+};
+
+export type LifecycleAction =
+  | { task: TaskAction; call?: never; http?: never }
+  | { call: CallAction; task?: never; http?: never }
+  | { http: HttpAction; task?: never; call?: never };
+
+const majorVersion =
```
ajperel, line 44: I was definitely like "Why is this here and why do we need it?" I'd suggest having a nice comment like params/index.ts does to explain what's going on here.

src/lifecycle/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  function: string;
+  body?: Record<string, unknown>;
+}
+
+export interface CallAction {
+  function: string;
+  params?: Record<string, unknown>;
+}
+
+export type HttpAction = ({ function: string } | { url: string }) & {
```
ajperel, line 33: Does the URL here need to accept  `string | Expression<string>` similar to ManifestLifecycleAction below to fully support CEL expressions?

src/runtime/loader.ts
```diff
@@ -209,5 +210,10 @@ export async function loadStack(functionsDir: string): Promise<ManifestStack> {
   if (declaredRoles.size > 0) {
     stack.requiredRoles = Array.from(declaredRoles);
   }
+
+  if (Object.keys(declaredLifecycleHooks).length > 0) {
+    stack.lifecycleHooks = { ...declaredLifecycleHooks };
+  }
+  clearDeclaredLifecycleHooks();
```
ajperel, line 217: clearDeclaredLifecycleHooks() documentation said it was primarily for testing. Why do you need to call it here?

Seems like you do need it here similar to  `clearGlobalRequiredAPIs();` and maybe we should have a slightly different comment on the method itself.  IF AI is correct because loadStack() will be called multiple times when reloading things for the emulator?

ajperel (commented):
Some mostly nitty comments after the fact as I review this PR to learn more than find issues. But still could be nice to address eventually. I could potentially send a PR myself if you prefer?

src/lifecycle/index.ts
wandamora: I missed updating ManifestLifecycleAction to only accept a string. None of these fields should be parameterized. I'll update this in a follow-up PR along with the documentation comment above.

src/runtime/loader.ts
wandamora: Hm, I can see how the "primarily for testing" bit can cause confusion since it's also used to clear the cache to avoid throwing errors about the hooks being already defined on subsequent loads. I'll remove that piece of the docstring in `clearDeclaredLifecycleHooks` in the follow-up PR.

PR #1917, opened 2026-06-30
fix zizmor findings

morganchen12:


Berlioz (approved):
LGTM, discussed over chat

PR #1923, opened 2026-07-14
fix/v2-auth-event-data-unpacking

shettyvarun268:
While testing and migrating extensions to v2 Cloud Functions, we noticed that properties inside event.data for onUserCreated and onUserDeleted triggers were missing vital information like email, displayName, and metadata.creationTime. This disparity from v1 happens because Eventarc delivers v2 Authentication payloads wrapped inside a google.events.firebase.auth.v2.AuthEventData envelope. Specifically, the target User object is nested under .value for creation events and .oldValue (or .old_value) for deletion events. Because getAuthEvent passed the outer envelope directly to userRecordConstructor, the constructor couldn't find the user attributes at the root level and fell back to null values. Additionally, protobuf JSON outputs createTime and photoUrl, whereas our constructor expects createdAt/creationTime and photoURL.

To fix this specifically for v2 without touching common or v1 code paths, we extracted an unpackAndNormalizeUserRecord helper in src/v2/providers/identity.ts. This helper unboxes value, oldValue, or old_value (falling back to raw.data for direct unit test payloads) and normalizes protobuf field names (createTime -> creationTime, photoUrl -> photoURL) with explicit runtime object guards before constructing the user record. This ensures v2 events receive full attribute parity while leaving v1 triggers completely untouched.

To verify this, we expanded the unit test suite in identity.spec.ts with new cases covering .value and .oldValue/.old_value envelope unpacking, photoUrl normalization, and null payload resilience. Ran Manual tests by deploying the functions and checking Cloud logs to ensure that all the attributes were cleanly unpacked.

### DIFF
--- a/src/v2/providers/identity.ts
+++ b/src/v2/providers/identity.ts
@@ -427,7 +427,16 @@ function unpackAndNormalizeUserRecord(rawData: unknown): User | undefined {
     return undefined;
   }
   const dataObj = rawData as Record<string, unknown>;
-  const rawUser = dataObj.value ?? dataObj.oldValue ?? dataObj.old_value ?? dataObj;
+  let rawUser: unknown;
+  if ("value" in dataObj) {
+    rawUser = dataObj.value;
+  } else if ("oldValue" in dataObj) {
+    rawUser = dataObj.oldValue;
+  } else if ("old_value" in dataObj) {
+    rawUser = dataObj.old_value;
+  } else {
+    rawUser = dataObj;
+  }
   if (!rawUser || typeof rawUser !== "object") {
     return undefined;
   }
--- a/src/v2/providers/identity.ts
+++ b/src/v2/providers/identity.ts
@@ -421,8 +421,12 @@ function getOptsAndHandler(
 // Matches both absolute paths (/projects/...) and relative paths (projects/...)
 const PROJECT_ID_REGEX = /(?:^|\/)projects\/([^\/]+)/;
 
-/** @hidden */
-function unpackAndNormalizeUserRecord(rawData: unknown): User | undefined {
+/**
+ * Converts the v2 Eventarc (`AuthEventData`) protobuf wire protocol (`value`/`oldValue`, `createTime`, `photoUrl`)
+ * into the v1 `LegacyEvent` wire protocol expected by `userRecordConstructor`.
+ * @hidden
+ */
+function convertV2EventToV1Event(rawData: unknown): User | undefined {
   if (!rawData || typeof rawData !== "object") {
     return undefined;
   }
@@ -465,7 +469,7 @@ function unpackAndNormalizeUserRecord(rawData: unknown): User | undefined {
 function getAuthEvent(raw: CloudEvent<unknown>): AuthEvent<User> {
   const event: AuthEvent<User> = { ...raw } as any;
   if (raw.data !== undefined && raw.data !== null) {
-    event.data = unpackAndNormalizeUserRecord(raw.data);
+    event.data = convertV2EventToV1Event(raw.data);
   }
   const rawAny = raw as any;
   // Support both lowercase (CloudEvents standard) and camelCase (local testing)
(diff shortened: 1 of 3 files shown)


src/v2/providers/identity.ts
```diff
@@ -422,10 +422,41 @@ function getOptsAndHandler(
 const PROJECT_ID_REGEX = /(?:^|\/)projects\/([^\/]+)/;
 
 /** @hidden */
+function unpackAndNormalizeUserRecord(rawData: unknown): User | undefined {
+  if (!rawData || typeof rawData !== "object") {
+    return undefined;
+  }
+  const dataObj = rawData as Record<string, unknown>;
+  const rawUser = dataObj.value ?? dataObj.oldValue ?? dataObj.old_value ?? dataObj;
```
wandamora: Jetski nit:
If dataObj is { value: null } (an explicit null payload inside an envelope), the nullish coalescing operator (??) skips null and moves to dataObj.oldValue ($\rightarrow$ undefined), dataObj.old_value ($\rightarrow$ undefined), and finally evaluates to dataObj ({ value: null }). This causes userRecordConstructor({ value: null }) to run instead of returning undefined.

Recommended refactor:
```ts
let rawUser: unknown;
if (Array.isArray(rawData)) {
  return undefined;
}
if ("value" in dataObj) {
  rawUser = dataObj.value;
} else if ("oldValue" in dataObj) {
  rawUser = dataObj.oldValue;
} else if ("old_value" in dataObj) {
  rawUser = dataObj.old_value;
} else {
  rawUser = dataObj;
}
```

shettyvarun268: Thanks Wanda! Refactored the code as pointed out.

wandamora (approved):
LGTM, thanks Varun!

src/v2/providers/identity.ts
```diff
@@ -422,10 +422,50 @@ function getOptsAndHandler(
 const PROJECT_ID_REGEX = /(?:^|\/)projects\/([^\/]+)/;
 
 /** @hidden */
+function unpackAndNormalizeUserRecord(rawData: unknown): User | undefined {
```
ajperel: This feels a little wrong in the sense that `userRecordConstructor` itself is documented as:

> * Helper function that creates a `UserRecord` class from data sent over the wire.
> * @param wireData data sent over the wire

Clearly if we're unpacking and normalizing the data we're _not_ calling that class with the raw Wire data. On the other side... this isn't that different from what getAuthEvent is doing either (which has some other normalization).

So I'm conflicted between saying:

- Why do we have this helper method rather than just inlining everything in getAuthEvent
OR
- Why is this not in `common/providers/identity` because it's methods to construct from raw wire events should handle... all wire events.

Thoughts?

shettyvarun268: So First, userRecordConstructor in the src/commom folder is shared by both v1 and v2 triggers. Because the .value / .oldValue envelope only exists in v2 CloudEvents (v1 just sends the flat user object directly), we kept the  unboxing/unpacking here in v2/ so we only pass the clean inner user object down to the shared constructor without risking any side effects for v1. Also I was treating any code that affects V1 as sort of frozen so as to not introduce any unexpected bugs in V1 behavior since its feature complete.

Secondly, I actually had everything inlined inside getAuthEvent at first! I pulled it out into this helper during jetski-review just to keep getAuthEvent clean and readable.

That said, if you'd rather move the field normalizations (createTime and photoUrl) into userRecordConstructor in common/ so it natively handles those across both versions, and leave only the .value unboxing here in v2/, I'm totally happy to adjust that!

ajperel: I think what I'd propose is:

update `userRecordConstructor` to be clear it is for the wire protocol of the v1 LegacyEvent

Then document this method to be clear it's converting from the v2 EventArc auth wire protocol to v1 LegacyEvent wire protocol. That could even be a new name like "convertV2EventToV1Event()"

Then I think everything is very followable and there's some obvious cleanup we could eventually do when we drop v1.

And we leave `AuthUserEvent` alone.


As I thought about it more I do agree with your method separation here. The UserRecord is separate from the other EventFields. I don't feel strongly if you wanted to combine... but with the right naming I think it'll all make sense and be nicely distinct.

ajperel (approved):

src/v2/providers/identity.ts
shettyvarun268: Thanks for the review Andy! Made the changes you proposed.

PR #1925, opened 2026-07-14
Revamp deploy pipeline to be stateless

inlined:
# Overview 

Rewrites the internal deploy pipeline to be stateless so that we don't need to commit to master in releases which improves security.

As of this change, you must include a relnote line even if it is "none". Relnotes can include explicit PR numbers or will use the current PR for tracking. Multiple relnotes can be in a single PR (for example, this one will name all the relnotes previously in CHANGELOG to prime the pump).

After this, you may also run the publisher script directly instead of remembering a cloud build enqueue command. The script will detect it is on your local machine, validate arguments, and then enqueue itself in the cloud build job to run in a trusted environment.

# Testing

Will test with a prerelease and then real release of a RC minor. Would like to not be submitting to cloud build et. al. until review though.

# Release Notes

(Copied from the now removed CHANGELOG to work with commit-based relnotes)

relnote: fix: Remove false warning when using Expression in cors option (#1802)
relnote: feat: Add requiresRole developer API for declarative security support and automatic Manifest extraction (#1908)
relnote: Validate literal `timeoutSeconds` values per v2 trigger type (0-540s for events, 0-3600s for HTTPS/callable, 0-1800s for task queues, 0-7s for identity functions) so misconfigured values fail at function-definition or manifest-extraction time instead of at deploy time. (#1877)
relnote: feat: Add requiresAPI function to allow declaring Google Cloud API dependencies in code. (#1900)
relnote: fix(v1): Call onInit for schedule.onRun functions (#1801)
relnote: feat: Add support to declare lifecycle hooks in functions. (#1915)
relnote: fix(cors): Fix issue using Params to set CORS allowed hosts (#1903)
relnote: fix(v2): Fix event data unpacking for auth event triggers (#1923)
relnote: feat: Add "v2/lifecycle" and "lifecycle" import paths for lifecycle hooks (#1926)


inlined:
Note that this PR's timing was quite interesting and revealed that the script doesn't handle removing reverted changes' relnotes. I can add that now or in the future if it comes up.

Edit: Why not just add it now. Gemini can do this much easier than I can

scripts/publish/cloudbuild.yaml
```diff
@@ -76,27 +81,40 @@ steps:
...
     dir: "${_REPOSITORY_NAME}"
-    args: ["bash", "./scripts/publish.sh", "${_VERSION}"]
+    entrypoint: bash
+    args:
+      - '-c'
+      - |
+        # Map our Cloud Build environment variables for the shell script execution
+        export GITHUB_TOKEN=$$GITHUB_TOKEN
+        
+        ./scripts/publish_worker.sh ${_BUMP_TYPE} \
```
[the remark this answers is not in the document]
inlined: Fixed. Changed the script name in `cloudbuild.yaml` to `./scripts/publish.sh`.

scripts/publish.sh
```diff
@@ -1,162 +1,293 @@
...
+  echo "⚠️ Warning: No stable SemVer tags found. Analyzing full history..."
+  CHANGELOG_NOTES="Initial release v$NEXT_VERSION."
 else
-  npm version $VERSION
+  echo "Identified stable base: $PREVIOUS_TAG. Harvesting notes up to HEAD..."
+  CHANGELOG_NOTES=""
+  
+  while read -r sha; do
+    COMMIT_SUBJECT=$(git log -1 --format="%s" "$sha")
+    PR_SUFFIX=$(echo "$COMMIT_SUBJECT" | grep -oE '\(#[0-9]+\)$')
```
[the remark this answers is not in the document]
inlined: Good catch! Added `|| true` to prevent the script from exiting on commits without a PR suffix.

scripts/publish.sh
```diff
@@ -1,162 +1,293 @@
...
+
+    if [ -n "$RELNOTE_BODY" ]; then
+      if echo "$RELNOTE_BODY" | grep -qE '\(#[0-9]+\)$'; then
+        CHANGELOG_NOTES+="- $RELNOTE_BODY"$'\n'
+      elif [ -n "$PR_SUFFIX" ]; then
+        CHANGELOG_NOTES+="- $RELNOTE_BODY $PR_SUFFIX"$'\n'
+      else
+        CHANGELOG_NOTES+="- $RELNOTE_BODY"$'\n'
+      fi
+    fi
```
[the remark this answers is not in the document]
inlined: Fixed. Implemented a `while read` loop to process each release note line individually.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -28,3 +28,26 @@ before sending PRs. We cannot accept code without this.
...
+### Release notes
+
+<!-- 
+  Your PR description must contain a "relnote:" line. 
+  Our CI release notes linter will verify this before allowing a merge.
+
+  Please keep EXACTLY ONE of the options below:
+-->
+
+relnotes: none
```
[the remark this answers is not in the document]
inlined: Fixed. Updated the default template value to the singular `relnote: none`.

.github/workflows/lint-relnotes.yaml
```diff
@@ -0,0 +1,34 @@
...
+          # This grabs the line, strips the prefix, and trims trailing/leading spaces
+          NOTE_CONTENT=$(echo "$PR_BODY" | grep -iE '^[Rr]elnotes?:' | head -n 1 | sed -E 's/^[Rr]elnotes?:[[:space:]]*//g' | xargs)
+
+          if [ -z "$NOTE_CONTENT" ]; then
+            echo "❌ Error: The 'relnote:' field is empty."
+            echo "Please provide a brief release note or write 'relnote: none' if this change doesn't require one."
+            exit 1
+          fi
+
+          echo "✅ Success! Found valid release note: '$NOTE_CONTENT'"
```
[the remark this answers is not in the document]
inlined: Fixed. Added explicit `permissions` block to limit `GITHUB_TOKEN` to `contents: read`.

scripts/publish.sh
```diff
@@ -1,162 +1,346 @@
...
+  echo "❌ Error: Invalid bump type '$BUMP_TYPE'. Must be 'major', 'minor', or 'patch'."
   exit 1
 fi
-if [[ $REPOSITORY_NAME == "" ]]; then
-  printusage
-  exit 1
+
+# Set default values
+IS_PRERELEASE=false
+DRY_RUN=false
```
joehan: I know this is how we had it before, but should we default to dry_run=true (go/tott/769 and all that?)

scripts/publish.sh
```diff
@@ -1,162 +1,346 @@
...
+# 4. NPM Publish Execution
+if [ "$DRY_RUN" = true ]; then
+  echo "🔍 [Dry Run] Skipping npm publish --tag $DIST_TAG"
+else
+  echo "Publishing package to npm under tag: $DIST_TAG..."
+  npm publish --tag "$DIST_TAG"
+fi
+
+# 5. Stateless Changelog Generation (Strict SemVer Tracking)
+echo "Calculating history interval to extract release notes..."
```
joehan, line 199: Nit: Consider splitting the changelog generation logic into its own script and referencing it here. I imagine this might be useful for folks to be able to run to check the current changelog - they could alwys do a dryrun, but thats a bit of overkill.

scripts/publish.sh
```diff
@@ -1,162 +1,346 @@
...
-cat CHANGELOG.md >> "${RELEASE_NOTES_FILE}"
-echo "Made the release notes."
-
-echo "Publishing to npm..."
-PUBLISH_ARGS=()
-if [[ -n "$DRY_RUN" ]]; then
-  echo "DRY RUN: running publish with --dry-run"
-  PUBLISH_ARGS+=(--dry-run)
+
+if [ -z "$CHANGELOG_NOTES" ]; then
```
joehan, line 212: Nit: Do we want a short circuit if there are no commits since that last release? Right now, we'd just cut another release with this fallback changelog.

scripts/publish.sh
```diff
@@ -1,162 +1,346 @@
...
-#echo "Making the tweet..."
-#npm install --no-save twitter@1.7.1
-#cp -v "${WDIR}/scripts/twitter.json" "${TEMPDIR}/${REPOSITORY_NAME}/scripts/"
-#node ./scripts/tweet.js ${NEW_VERSION}
-#echo "Made the tweet."
+if [ "$DRY_RUN" = true ]; then
+  echo "🏁 Dry run completed successfully! No modifications were made to remote systems."
+else
+  echo "🚀 Release of $PACKAGE_NAME@$NEXT_VERSION successfully completed!"
+fi
```
joehan, line 260: Newline?

scripts/publish.sh
```diff
@@ -1,162 +1,346 @@
...
 
-echo "Pushing to GitHub..."
-git push origin master --tags
-echo "Pushed to GitHub."
+  GH_RELEASE_FLAGS=""
+  if [ "$IS_PRERELEASE" = true ]; then
+    GH_RELEASE_FLAGS="--prerelease"
+  fi
+
+  gh release create "v$NEXT_VERSION" \
```
joehan, line 247: Optional: For firebase-tools, we've enabled immutable releases for improved security. It's less important for functions since you don't host any artifacts there, but still a nice thing to do (especially since it adds an attestation to the release). 

If you do that, I'd recommend having this create a draft release first, and then having the release runner manually publish the draft after they verify that it looks as expected.

scripts/publish.sh
```diff
@@ -1,162 +1,346 @@
...
-  echo "Published a pre-release version. Skipping post-release actions."
-  exit
+  rm release_notes.md
 fi
 
-if [[ $DRY_RUN != "" ]]; then
-  echo "All other commands are mutations, and we are doing a dry run."
-  echo "Terminating."
-  exit
+# 8. Twitter Broadcast
```
joehan: Shall we finally remove this since we disabled twitter a long time ago (and they banned our account?)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "firebase-functions",
-  "version": "7.2.6-rc.0",
+  "version": "0.0.0-development",
```
joehan, line 3: Intentional?

.github/workflows/lint-relnotes.yaml
```diff
@@ -0,0 +1,37 @@
...
+            exit 1
+          fi
+
+          # 2. Extract the actual note content to ensure it isn't blank
+          # This grabs the line, strips the prefix, and trims trailing/leading spaces
+          NOTE_CONTENT=$(echo "$PR_BODY" | grep -iE '^[Rr]elnotes?:' | head -n 1 | sed -E 's/^[Rr]elnotes?:[[:space:]]*//g' | xargs)
+
+          if [ -z "$NOTE_CONTENT" ]; then
+            echo "❌ Error: The 'relnote:' field is empty."
+            echo "Please provide a brief release note or write 'relnote: none' if this change doesn't require one."
```
ajperel, line 33: I _think_ this will also end up just accepting `relnote: ` with no content right? Do we want to allow that?

ajperel (commented):
Was it intentional to remove the executable bit from `publish.sh`?

Did a quick review but I know so little about Github workflows it's not a substantial one. I hope Joe's giving you great feedback.

scripts/publish.sh
inlined: Done. We have defaulted the script to dry-run mode (`DRY_RUN=true`), and added a `--no-dry-run` flag to explicitly perform a real release.

scripts/publish.sh
inlined: Done. Extracted the changelog generation and revert-filtering logic into a standalone `scripts/changelog.sh` script, which is now invoked by `publish.sh`.

scripts/publish.sh
inlined: Done. Added a check to abort the release with an error if no new commits have been made since the last stable tag. I also added a `--force` flag to override this check. Do we want to keep the `--force` override option in case we truly need to force a release when there are no new commits?

scripts/publish.sh
inlined: Done. Formatted the file using Prettier to ensure a trailing newline is present.

scripts/publish.sh
inlined: Thanks for the suggestion! Since this is optional and functions does not host build artifacts, we will leave it as-is for now, but will keep it in mind for future security hardening.

scripts/publish.sh
inlined: Agreed. I have removed the Twitter broadcast step from the publish script and the Cloud Build steps, and deleted the unused `scripts/tweet.js` and `scripts/publish/twitter.json.enc` files.

package.json
inlined: Yes, this is intentional! In the revamped stateless publish flow, we no longer store or increment the release version in `package.json` inside git to avoid committing back to the repo after a release. Instead, the version is set to a placeholder (`0.0.0-development`), and the Cloud Build runner dynamically queries npm for the latest version and bumps it in memory before publishing.

.github/workflows/lint-relnotes.yaml
inlined: Actually, the workflow already rejects this case! If `relnote: ` has no content, `NOTE_CONTENT` will be empty after the `sed` and `xargs` cleaning steps, triggering the `[ -z "$NOTE_CONTENT" ]` check and exiting with `exit 1` (lines 31-35).

inlined:
Done with a pass. Here's what `changelog.sh` prints on this branch:

```
- fix: Remove false warning when using Expression in cors option (#1802)
- feat: Add requiresRole developer API for declarative security support and automatic Manifest extraction (#1908)
- Validate literal `timeoutSeconds` values per v2 trigger type (0-540s for events, 0-3600s for HTTPS/callable, 0-1800s for task queues, 0-7s for identity functions) so misconfigured values fail at function-definition or manifest-extraction time instead of at deploy time. (#1877)
- feat: Add requiresAPI function to allow declaring Google Cloud API dependencies in code. (#1900)
- fix(v1): Call onInit for schedule.onRun functions (#1801)
- feat: Add support to declare lifecycle hooks in functions. (#1915)
- fix(cors): Fix issue using Params to set CORS allowed hosts (#1903)
```

joehan (approved):

inlined:
Andy thought Joe's approval was enough, but he doesn't have write access. Bypassed rules and merged given that Andy had intended to defer to Joe and that it's well after hours.

ajperel:
Confirming that was all true... sorry for not explicitly approving.

PR #1931, opened 2026-07-16
Adding lifecycle events in docgen

shettyvarun268:
In PR #1926, afterFirstDeploy and afterRedeploy were moved from top-level v2/index.ts exports to deep import paths (firebase-functions/v2/lifecycle), causing api-extractor to stop discovering them during documentation extraction. This PR explicitly exports lifecycle in the documentation-only entrypoint (src/v2/index.doc.ts) so afterFirstDeploy(), afterRedeploy(), and their action types (LifecycleAction, TaskAction, CallAction, HttpAction) show up cleanly on the DevSite 2nd Gen API reference.

relnote: Bringing in lifecycle triggers into docgen.


wandamora (approved):
Thank you!

PR #1932, opened 2026-07-16
fix: dynamically increment prerelease versions in publish.sh

inlined:
# Description
Fixes an issue in `scripts/publish.sh` where prerelease publishing (e.g. running with `--prerelease`) would fail if `rc.0` was already published on npm for that target version series, since it always bumped from the stable version. It now dynamically checks npm for any existing prerelease versions of the target version series, and performs a `prerelease` increment (e.g., to `rc.1`) if one exists.

Newly assumes the jq command exists to avoid silly node scripts (will need to test in prod or build a new container). Now handles gh not existing since it didn't in the past.

# Release Notes
relnote: chore: improve release script

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -142,6 +142,17 @@ fi
 # ==============================================================================
 echo "☁️  [Worker Mode] Cloud Build environment detected (Build ID: $BUILD_ID)."
 
+# Pre-flight runner tool checks
+if ! command -v jq &> /dev/null; then
+  echo "❌ Error: 'jq' utility is not installed. Please install jq to run the worker."
+  exit 1
+fi
+
+if ! command -v gh &> /dev/null; then
+  echo "❌ Error: 'gh' CLI tool is not installed. Please install GitHub CLI (gh) to run the worker."
+  exit 1
+fi
+
 # 1. Optional Branch Checkout
 if [ -n "$TARGET_BRANCH" ]; then
   echo "Checking out target branch: $TARGET_BRANCH..."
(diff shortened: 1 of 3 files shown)


scripts/publish.sh
```diff
@@ -174,10 +174,23 @@ echo "--------------------------------------------------------"
...
 if [ "$IS_PRERELEASE" = true ]; then
-  PRE_BUMP_TYPE="pre${BUMP_TYPE}"
-  echo "Preparing prerelease version change using '$PRE_BUMP_TYPE' (preid: rc)..."
-  npm version "$STABLE_VERSION" --no-git-tag-version
-  npm version "$PRE_BUMP_TYPE" --preid="rc" --no-git-tag-version
+  TARGET_PRE_BASE=$(./node_modules/.bin/semver "$STABLE_VERSION" -i "pre${BUMP_TYPE}" --preid "rc")
+  TARGET_STABLE_LIMIT=$(./node_modules/.bin/semver "$STABLE_VERSION" -i "$BUMP_TYPE")
+  
+  echo "Checking npm for existing prerelease versions in the range >=$TARGET_PRE_BASE <$TARGET_STABLE_LIMIT..."
+  LATEST_PRE_VERSION=$(npm view "$PACKAGE_NAME@>=$TARGET_PRE_BASE <$TARGET_STABLE_LIMIT" version 2>/dev/null | tail -n 1 | grep -oE "[0-9]+\.[0-9]+\.[0-9]+-rc\.[0-9]+" | head -n 1 || true)
```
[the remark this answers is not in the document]
inlined: Implemented robust JSON parsing and semver-based sorting of matching versions, while correctly handling error/404 JSON objects returned from npm view.

scripts/publish.sh
```diff
@@ -159,62 +159,52 @@ echo "Executing test suite..."
 npm test
 
 # 3. Stateless Version Discovery
-PACKAGE_NAME=$(node -p "require('./package.json').name")
+PACKAGE_NAME=$(jq -r '.name' package.json)
 STABLE_VERSION=$(npm view "$PACKAGE_NAME" version)
 
+# Find the absolute latest version overall on npm (including prereleases)
+LATEST_VERSION=$(npm view "$PACKAGE_NAME" versions --json | grep -oE '[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+)?' | tail -n 1)
```
ajperel: I was doing some diving to fully understand this. It totally works and I'm fine merging it. Gemini thinks if we're already using jq it'd be slightly more robust to use it:

```
LATEST_VERSION=$(npm view "$PACKAGE_NAME" versions --json | jq -r '[.[] | select(test("^[0-9]+\\.[0-9]+\\.[0-9]+(-rc\\.[0-9]+)?$"))] | .[-1]')
```
It does all the stripping so you just reduce to testing the version numbers to be fair... but I kinda doubt npm view versions --json is going to change much.

scripts/publish.sh
```diff
@@ -159,62 +159,52 @@ echo "Executing test suite..."
 npm test
 
 # 3. Stateless Version Discovery
-PACKAGE_NAME=$(node -p "require('./package.json').name")
+PACKAGE_NAME=$(jq -r '.name' package.json)
```
ajperel, line 173: Why is it safe to assuming jq will be installed? It wasn't on my cloudtop for example (though I've now installed it).

ajperel (approved):

scripts/publish.sh
inlined: Good catch! Since `jq` is a hard dependency of the script now, using it directly for array filtering is much more robust than grep/tail text parsing. Updated to use the suggested `jq` command.

scripts/publish.sh
inlined: Very valid concern. To ensure safety across any local or custom execution environments, I added explicit pre-flight check assertions at the beginning of the worker mode block. This will fail-fast with a clear error message if `jq` or `gh` is missing.

inlined:
Running the script as if it were the builder locally using BUILD_ID=test ./scripts/publish.sh (and not using --no-dry-run) now shows the correct prerelease incrementing

PR #1934, opened 2026-07-17
fix: rename task timeout kind and also apply it to scheduler

### LINKED ISSUE #1933: v7.3.0 rejects valid 1800s timeout for v2 onSchedule functions
### Related issues

- #1737 is related to timeout validation for event-triggered functions, but does not cover scheduled functions.
- #1877 introduced the v2 `timeoutSeconds` validation released in `firebase-functions` 7.3.0.

### [REQUIRED] Version info

**node:** v24.18.0

**firebase-functions:** 7.3.0

**firebase-tools:** 15.24.0

**firebase-admin:** 13.10.0

### [REQUIRED] Test case

```js
import { onSchedule } from "firebase-functions/v2/scheduler";

for (const timeoutSeconds of [540, 541, 1800]) {
  try {
    onSchedule(
      { schedule: "0 0 * * *", timeoutSeconds },
      async () => {},
    );
    console.log(`${timeoutSeconds}: accepted`);
  } catch (error) {
    console.log(`${timeoutSeconds}: rejected: ${error.message}`);
  }
}
```

1. Create an empty Node project.
2. Install `firebase-functions@7.3.0`.
3. Save the test case as an ES module and run it with Node.

For example:

```sh
npm init -y
npm install firebase-functions@7.3.0
node test.mjs
```

`onSchedule()` should accept `timeoutSeconds` values up to 1800 seconds.

The official Firebase documentation lists these maximum durations:

- HTTP/callable: 3600 seconds
- Scheduled/task queue: 1800 seconds
- Other event-driven functions: 540 seconds

https://firebase.google.com/docs/functions/manage-functions#set_timeout_and_memory_allocation

The Firebase CLI also has a distinct 1800-second validation limit for scheduled functions.

The SDK accepts 540 seconds, but rejects both 541 and 1800 seconds at function-definition time:

```text
540: accepted
541: rejected: timeoutSeconds must be between 0 and 540 for event-handling functions. Got 541.
1800: rejected: timeoutSeconds must be between 0 and 540 for event-handling functions. Got 1800.
```

In v7.3.0, the v2 scheduler provider passes `"event"` to `optionsToEndpoint`, so scheduled functions receive the generic 540-second event limit rather than the documented 1800-second scheduled-function limit:

https://github.com/firebase/firebase-functions/blob/ce42106760149bf1ef656ced487187ce10cdcc52/src/v2/providers/scheduler.ts#L226-L239

No. Function definition/source discovery fails before deployment when an existing v2 scheduled function uses a valid timeout above 540 seconds. The same function configuration works with `firebase-functions` 7.2.5.

IzaakGough:
Fixes #1933 

This PR renames the "task" timeout kind to "scheduledOrTask" and ensures that both task and scheduler functions use this. The documentation states that task and scheduler functions should have the same timeout settings, but scheduler functions were falling back to a generic "event" timeout kind.

## Testing
- Verified that the changes fix the example as described in the issue.
- Automated tests pass

relnote: none

### DIFF
--- a/src/v2/options.ts
+++ b/src/v2/options.ts
@@ -44,7 +44,7 @@ export { RESET_VALUE } from "../common/options";
 
 /**
  * Maximum timeout in seconds for event-handling functions (e.g. Firestore,
- * Realtime Database, Pub/Sub, Storage, Eventarc, alerts, scheduler).
+ * Realtime Database, Pub/Sub, Storage, Eventarc, alerts).
  * See https://firebase.google.com/docs/functions/manage-functions
  * @internal
  */
@@ -57,10 +57,10 @@ export const MAX_EVENT_TIMEOUT_SECONDS = 540;
 export const MAX_HTTPS_TIMEOUT_SECONDS = 3600;
 
 /**
- * Maximum timeout in seconds for Task Queue functions.
+ * Maximum timeout in seconds for scheduled and Task Queue functions.
  * @internal
  */
-export const MAX_TASK_TIMEOUT_SECONDS = 1800;
+export const MAX_SCHEDULE_OR_TASK_TIMEOUT_SECONDS = 1800;
 
 /**
  * Maximum timeout in seconds for Identity blocking functions.
@@ -72,7 +72,7 @@ export const MAX_IDENTITY_TIMEOUT_SECONDS = 7;
  * Function category used to pick a `timeoutSeconds` upper bound.
  * @internal
  */
-export type TimeoutKind = "event" | "https" | "task" | "identity";
+export type TimeoutKind = "event" | "https" | "scheduledOrTask" | "identity";
 
 /**
  * List of all regions supported by Cloud Functions (2nd gen).
@@ -412,9 +412,9 @@ export function assertTimeoutSecondsValid(
       max = MAX_HTTPS_TIMEOUT_SECONDS;
       label = "HTTPS and callable";
       break;
-    case "task":
-      max = MAX_TASK_TIMEOUT_SECONDS;
-      label = "task queue";
+    case "scheduledOrTask":
+      max = MAX_SCHEDULE_OR_TASK_TIMEOUT_SECONDS;
+      label = "scheduled or task queue";
       break;
     case "identity":
       max = MAX_IDENTITY_TIMEOUT_SECONDS;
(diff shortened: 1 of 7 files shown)


src/v2/options.ts
```diff
@@ -44,7 +44,7 @@ export { RESET_VALUE } from "../common/options";
 
 /**
  * Maximum timeout in seconds for event-handling functions (e.g. Firestore,
- * Realtime Database, Pub/Sub, Storage, Eventarc, alerts, scheduler).
+ * Realtime Database, Pub/Sub, Storage, Eventarc, alerts).
```
cabljac, line 47: Nice catch applying the kind to the scheduler too, and the boundary tests are exactly right.

One doc gap: the `GlobalOptions.timeoutSeconds` `@remarks` (around line 202) still only lists task queue functions at 1,800s, so the generated reference docs won't tell anyone that scheduled functions now share that cap. Please update that sentence to something like "Task queue and scheduled functions have a maximum timeout of 1,800s (30 minutes)." The same stale copy is on `TaskQueueOptions.timeoutSeconds` in tasks.ts:87, worth fixing both in this PR since it's the same sentence.

(That sentence is copy-pasted across most provider option types too, e.g. pubsub.ts, https.ts, storage.ts. Fine to leave those for a follow-up, but flagging so we don't lose track.)

src/v2/options.ts
```diff
@@ -57,10 +57,10 @@ export const MAX_EVENT_TIMEOUT_SECONDS = 540;
 export const MAX_HTTPS_TIMEOUT_SECONDS = 3600;
 
 /**
- * Maximum timeout in seconds for Task Queue functions.
+ * Maximum timeout in seconds for scheduled and Task Queue functions.
  * @internal
  */
-export const MAX_TASK_TIMEOUT_SECONDS = 1800;
+export const MAX_SCHEDULE_OR_TASK_TIMEOUT_SECONDS = 1800;
```
cabljac: nit: the constant says `SCHEDULE_OR_TASK` but the kind string is `"scheduledOrTask"`. Renaming to `MAX_SCHEDULED_OR_TASK_TIMEOUT_SECONDS` keeps the pair consistent and greppable together, and both are `@internal` so the rename is free.

cabljac (approved):
Approving. Two inline comments: the doc gap is worth fixing in this PR, the naming one is take it or leave it.

wandamora (approved):
LGTM as google review

ottob:
It would be very nice if this could be included in 7.3.2

PR #1941, opened 2026-07-28
fix: avoid CORS declaration interop requirement

cexbrayat:
### Description

`@types/cors` declares the package with `export =`. PR #1903 added exported CORS types that caused TypeScript to preserve the source default import in `lib/common/providers/https.d.ts`.

Default-importing an `export =` module in a published declaration forces consumers to enable `esModuleInterop`. Use a namespace import instead so that compiler-option requirement does not leak into consumer projects. The explicit `.default` invocation preserves the existing runtime behavior with the repository's current TypeScript and tsdown configuration.

This fixes the declaration regression visible in `firebase-functions@7.3.0`.

The packaging test now compiles a real packed consumer without `esModuleInterop` to prevent regressions.

### DIFF
--- a/scripts/packaging-test/index.ts
+++ b/scripts/packaging-test/index.ts
@@ -1,0 +1,3 @@
+import { onCall } from "firebase-functions/v2/https";
+
+onCall({ cors: true }, () => null);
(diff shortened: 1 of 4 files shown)


cexbrayat:
Let me know if the gemini feedback should be applied (and if the packaging test is useful).

PR #1944, opened 2026-07-30
feat: accept RegExp values in Expression/thenElse for CORS options

### LINKED ISSUE #1943: Params CompareExpression `.thenElse` does not support `RegExp` as value, therefore disallows the use of RegExp values for CORS options
### Related issues

#1787 
#1406 
#1903 

Actually I am trying to do this https://github.com/firebase/firebase-functions/pull/1903#issuecomment-5105925317 - however it turned out that #1903 unfortunately does not support setting regex CORS options directly from the .env files.

### [REQUIRED] Version info

**node:** v24.18.0

**firebase-functions:** v7.3.0

**firebase-tools:** v15.23.0

**firebase-admin:** v13.10.0

### [REQUIRED] Test case

```typescript
const environmentVariable_Some = defineBoolean('SOME')
const environmentVariable_Staging = defineBoolean('IS_STAGING')

const corsOptionsLocal = [/^http:\/\/localhost:808[0|1]$/]
const corsOptionsStaging = [/^https:\/\/(staging|otherstaging)\.example\.com$/]
const corsOptionsProduction = [/^https:\/\/(some|other)\.example\.com$/]

const corsOptions = environmentVariable_Some
  .equals(true)
  .thenElse(corsOptionsLocal, environmentVariable_Staging.equals(true).thenElse(corsOptionsStaging, corsOptionsProduction))

export const myFn = onCall(
  {
    cors: corsOptions
  },
  async () => {
    ...
  }
```

1. Run `npx tsc`

I can set dynamic (RegExp-based) allowed CORS hosts dynamically depending on the environment.
The `onCall` `cors` option allows the following types:
```
(property) HttpsOptions.cors?: string | boolean | RegExp | Expression<string> | Expression<string[]> | (string | RegExp)[] | undefined
```
...so it should also be possible to create a `Expression<RegExp>` using the `.equals(xx).thenElse(...)` `CompareExpression` helper.

```
Argument of type 'RegExp[]' is not assignable to parameter of type 'string | number | boolean | string[] | Expression<string | number | boolean | string[]>'.
  Type 'RegExp[]' is not assignable to type 'string[]'.
    Type 'RegExp' is not assignable to type 'string'.
```

No

IzaakGough:
Fixes #1943

`Expression<T>` was constrained to `string | number | boolean | string[]`, so a param ternary could not select between `RegExp` values even though `cors` accepts them. Users hit a type error when writing something like `params.defineBoolean("X").thenElse(/a\.com$/, /b\.com$/)` for a v2 HTTPS function's `cors` option.

This widens the `Expression` type parameter to a new exported `ExpressionValue` union that also covers `RegExp` and `Array<string | RegExp>`, and adds those expression forms to `CorsOption`.

`refOf` also gains a `RegExp` case. That one is not a deploy fix: `cors` is resolved per request by `resolveCorsOrigin` and has no entry in `optionsToEndpoint`, `optionsToTrigger` or `optionsToTriggerAnnotations`, so a cors expression never reaches the wire manifest. What the case fixes is the diagnostic string, reached via `.toCEL()`, `.value()` and `toJSON()`, where a `RegExp` previously stringified to `{}`. A test in both v1 and v2 pins that cors stays out of the manifest, since that is the assumption the change rests on.

relnote: none


cabljac (commented):
Nice work @IzaakGough, solid PR. Checked out the branch, ran the specs and compiled the #1943 repro against it, all good. The `refOf` fix and the two-branch E2E preflight tests are exactly the right shape. This all makes sense to me, but since it widens the public API (`Expression`, `CorsOption`, new `ExpressionValue` export) we'll run it by the Firebase team before it lands.

Small asks inline. One more on the issue itself: this fully fixes #1943 as filed, but the author also mentions wanting regex CORS patterns read directly from .env files (via #1903), which this doesn't cover (an env var string still won't become a RegExp). Since merge auto-closes the issue, please reply on the thread first clarifying that. Whenever a fix partially addresses what a reporter wants, say so before the close, otherwise they reopen confused.

src/v2/providers/https.ts
```diff
@@ -76,13 +76,7 @@ export interface HttpsOptions extends Omit<GlobalOptions, "region" | "enforceApp
...
    * Defaults to true for {@link https.CallableFunction} and false otherwise.
    */
-  cors?:
-    | string
-    | Expression<string>
-    | Expression<string[]>
-    | boolean
-    | RegExp
-    | Array<string | RegExp>;
+  cors?: CorsOption;
```
cabljac, line 82: `HttpsOptions.cors` now references `CorsOption` in the public API, but it isn't in this module's export list (line 54). I ran api-extractor against the branch and it flags `(ae-forgotten-export) The symbol "CorsOption" needs to be exported by the entry point`, so the docs render the type unlinked. It's a warning rather than a failure, and v1 has the same gap on master, but since we're touching this surface please add `CorsOption` to the `export type {...}` here and in v1's provider.

src/params/types.ts
```diff
@@ -24,12 +24,26 @@ import * as logger from "../logger";
 
 const EXPRESSION_TAG = Symbol.for("firebase-functions:Expression:Tag");
 
+/**
+ * The types an `Expression` can resolve to. `string`, `number`, `boolean` and
+ * `string[]` cover the values a param itself can hold; `RegExp` and
+ * `Array<string | RegExp>` are additionally allowed so that expressions can
+ * select between literals for options that accept them, such as `cors`.
+ */
+export type ExpressionValue =
```
cabljac: Tiny one: `string[]` is already covered by `Array<string | RegExp>`, so the union member is redundant for the type checker. Your comment explains it documents the param-holdable types, which is a fair reason to keep it, so your call. If you keep it, no change needed.

src/params/types.ts
```diff
@@ -171,13 +183,17 @@ export function transform(
  * - Arrays are represented as []-delimited, parsable JSON
  * - Numbers and booleans are not quoted explicitly
  */
-function refOf<T extends string | number | boolean | string[]>(arg: T | Expression<T>): string {
+function refOf<T extends ExpressionValue>(arg: T | Expression<T>): string {
```
cabljac, line 186: The doc comment above `refOf` still only describes strings, arrays, numbers and booleans. Please add a line for the new RegExp behavior (emitted as its quoted string form) so it matches what the function does.

src/params/types.ts
```diff
@@ -169,15 +181,20 @@ export function transform(
...
+ * - RegExps are emitted as their quoted `toString()` form, e.g. `"/foo$/"`
  * - Numbers and booleans are not quoted explicitly
  */
-function refOf<T extends string | number | boolean | string[]>(arg: T | Expression<T>): string {
+function refOf<T extends ExpressionValue>(arg: T | Expression<T>): string {
   if (arg instanceof Expression) {
     return arg.toString();
   } else if (typeof arg === "string") {
     return `"${arg}"`;
+  } else if (arg instanceof RegExp) {
```
cabljac, line 191: These branches can't reach the wire manifest. `cors` is never serialized, it has no entry in `optionsToEndpoint`, `optionsToTrigger` or `optionsToTriggerAnnotations`, and is resolved per-request via `resolveCorsOrigin` instead (v2 https.ts:330, v1 https.ts:99). What this actually fixes is the diagnostic string, reached through `.toCEL()`, `.value()`, and `toJSON()` when the expression sits inside a `JSON.stringify` or a `logger.*` call. Worth keeping, but could you reword the PR body? "fixes CEL serialization so a RegExp is emitted as its string form" reads like a deploy bug got fixed.

Then please add a test pinning that a cors expression is resolved at runtime and never serialized. `"/^foo$/"` isn't a CEL regex literal, so if `cors` ever does get added to the manifest, the CLI resolves it to the literal string `/^foo$/` and the origin silently matches nothing. Whenever a change is safe only because a value never reaches some consumer, pin the "never reaches" part, since that's the assumption holding the change up.

Throwing here instead would break `Expression.value()`, which interpolates `this.toString()` into its deploy-time warning (types.ts:62), so leaving it non-throwing is right.

src/common/providers/https.ts
```diff
@@ -715,6 +715,8 @@ export type CorsOption =
   | string
   | Expression<string>
   | Expression<string[]>
+  | Expression<RegExp>
```
cabljac, line 729: `CorsOption` becomes a public name in this PR (the new `export type` in v1/providers/https.ts and v2/providers/https.ts put it on the `v1.https` / `v2.https` namespaces), so it needs a JSDoc block. Exported symbols get JSDoc, and this one will show up in the reference docs with nothing next to it.

While you're there: the `cors` JSDoc on both option types (v2 https.ts:75-79, v1 https.ts:48-52) still only describes `string` / `RegExp` / `Array`. The new capability, a param ternary picking between them, isn't mentioned anywhere a user would look. That's the actual feature, so it belongs in the doc for the option, not only in the PR body.

src/params/types.ts
cabljac: The JSDoc here draws a real distinction: what a param can hold, versus what an expression can resolve to. Only the second half got a name though. The first is still written out literally five times: `CompareExpression` (line 232), `ParamSpec` (425), `WireParamSpec` (448), `ParamOptions` (459), `Param` (477). Could you extract a `ParamValue` alias and use it in those five places? Once two constraints are meant to differ, both need names, otherwise the next person widening one has no way to see the other was deliberate.

src/common/providers/https.ts
```diff
@@ -715,6 +715,8 @@ export type CorsOption =
   | string
   | Expression<string>
   | Expression<string[]>
```
cabljac, line 728: Nit: `Expression` is covariant in `T` (it only appears in return positions), so `Expression<string[]>` is now assignable to `Expression<Array<string | RegExp>>` and this union member is redundant. Same for `string[]` inside `ExpressionValue`, which `Array<string | RegExp>` already covers. I checked both with tsc, and inference at `thenElse` call sites doesn't change either way. Keeping them spelled out is arguably clearer in the published `.d.ts`, so is that intended?

cabljac (changes requested):
This looks like the right general design, requesting changes for some things below.

Widening at the `Expression` level looks forced by the hierarchy (`thenElse` returns a `TernaryExpression`, which extends `Expression<T>`), and I checked the widening can't leak a `RegExp` into anything serialized: tsc rejects `Expression<RegExp>` on `region`, `serviceAccount`, `memory`, `omit` and `networkInterface.tags`.

src/params/types.ts
IzaakGough: Good catch, reworded the PR body to say this is the diagnostic string rather than a deploy fix.

Added the test in both v1 and v2 in ea7f1a50. It asserts the serialized `__endpoint` and `__trigger` contain no `cors` key and none of the origin patterns, with a positive control matching the trigger key so it cannot pass vacuously. I checked it bites by temporarily writing `celOf(opts.cors)` onto the v2 endpoint, which fails it on `expected ... to not include 'cors''.

src/common/providers/https.ts
IzaakGough: Done in 25765415. `CorsOption` has a JSDoc block covering `true`/`false`, string, RegExp, array, and the expression form, and the `cors` option JSDoc in both v1 and v2 now mentions the param ternary with an example.

src/params/types.ts
IzaakGough: Agreed, extracted as `ParamValue` in 0681fd38 and used in all five places, with `ExpressionValue` now written as `ParamValue | RegExp | Array<string | RegExp>` so the widening reads directly. Exported from `params/index.ts` alongside `ExpressionValue`, since it lands in the public `.d.ts` via `ParamSpec`, `ParamOptions` and `Param`.

src/common/providers/https.ts
IzaakGough: Intentional, keeping them spelled out. `string[]` is the type a `defineList` param holds, so naming it separately keeps that capability visible in the published `.d.ts` instead of collapsing into a single array type. The `ParamValue` extraction makes the same point structurally now.

cabljac (approved):

src/common/providers/https.ts
```diff
@@ -711,10 +711,23 @@ type v2CallableHandler<Req, Res, Stream> = (
...
+ */
 export type CorsOption =
   | string
   | Expression<string>
+  // Assignable to Expression<Array<string | RegExp>> already, since Expression is
+  // covariant in T. Kept so the published .d.ts names the `defineList` case.
   | Expression<string[]>
+  | Expression<RegExp>
+  | Expression<Array<string | RegExp>>
   | boolean
```
ajperel, line 731: Should we also support Expression<boolean>?

ajperel (commented):
I need to spend some more time wrapping my head around all this. It's a part of functions I haven't looked at too much. In the meantime I've asked @Berlioz to review as well as he knows more of it. But I want to understand and be confident in this PR.

Berlioz:
Sorry to jump and and say things right before running off for vacation, but my first instinct here is wondering whether doing this kind of invasive surgery on the params type system is worth the risk when there's a perfectly reasonable hack available (we control how options are set by the SDK, so we can just special-case the cors option for anything beginning and ending with `\`, right?). Yes that's a hack, but I'm not sure it's any more of a hack than creating a new Expression type that can be reasonably used in only a single place.

If we _do_ decide that type surgery is worth doing in this case, this PR pretty much LGTM though.

ajperel:
Yeah. I agree this PR works but I'm wondering if it'd better to go in a different direction and I think this unfortunately needs some more deep strategic consideration we need to think through. Some other options include:

Specific to enabling CORS with minimal changes:

- Update it so we convert regex looking strings into regex's as @Berlioz recommends. It looks like Python might do this (though not from Functions SDK but lower level libraries)
- Update it so that `cors` accepts a function that can calculate all this safely at runtime without the `thenElse` expression at all. The underlying infra already accepts a method

More generally like this PR:

- The Dart SDK explicitly separates deploy time options (that must support CEL) and runtime options that don't. If we really want to start making these expressions support a wider array of types that might be the better way to do it.

PR #1947, opened 2026-08-08
fix(v2): re-export express Response from the https provider

### LINKED ISSUE #1798: Re-export express.Response
### [REQUIRED] Version info

**node:**
v20.19.0

**firebase-functions:**
6.1.0

**firebase-tools:**
14.2.0

**firebase-admin:**
12.7.0

### [REQUIRED] Test case
Would it be possible to re-export `express.response` so it can be used in user code for the function definition of the `onRequest` handler?

```typescript
// handler.ts
export async function myHandler(
  request: functions.https.Request,
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  response: any
)
// index.ts
export const foobar = onRequest(
    secrets: [],
    timeoutSeconds: 540,
    minInstances: minInstancesConfig,
  },
  async (request, response) => await myHandler(request, response)
);
```

I'd prefer to have a type available for response instead of having to use `any`, ie `functions.https.Response`.

`Response` is taken from Express 4.x. My project already includes Express 5.x, so it's awkward to include it separately, and it would be cleaner if firebase-functions just provided the type interface.

No problem here.

kyungseopk1m:
### Description

`onRequest` handlers take an `express.Response`, but `src/v2/providers/https.ts` only re-exports `Request`. To annotate a handler parameter you have to add `express` as a direct dependency just to name the type, even though `firebase-functions` already depends on it.

v1 does not have this gap: `src/v1/cloud-functions.ts` re-exports both `Request` and `Response`, so `functions.Response` resolves today. This brings v2 in line.

`Response` is re-exported from `express` rather than aliased as `export type Response = express.Response`, because the express type is generic (`Response<ResBody, Locals, StatusCode>`) and an alias drops the type parameters, which makes `https.Response<MyBody>` fail with TS2315.

Fixes #1798.

### Code sample

Before, this needs express installed and imported directly:

```ts
import { onRequest } from "firebase-functions/v2/https";
import type { Response } from "express";

export const fn = onRequest((req, res: Response<{ ok: boolean }>) => {
  res.json({ ok: true });
});
```

After:

```ts
import { onRequest, type Request, type Response } from "firebase-functions/v2/https";

export const fn = onRequest((req: Request, res: Response<{ ok: boolean }>) => {
  res.json({ ok: true });
});
```

### Scenarios Tested

`npm run build` and `npm run lint` pass.

Checked the emitted surface rather than only the source: after building, a probe importing `https` from `lib/v2/index` and using `https.Response<{ ok: boolean }>`, bare `https.Response`, `https.Request`, and a full handler signature type checks with `tsc --noEmit`. The generic form is what fails on an alias, so it is the case that matters here.

### Release notes

relnote: fix(v2): re-export the express `Response` type from `firebase-functions/v2/https` (#1798)


IzaakGough (approved):
lgtm!

shettyvarun268 (approved):
Nit: Not a blocker at all since its a small change, but we can have a small type assertion test to ensure that Response export does not regress in future refactors.

LGTM otherwise!

kyungseopk1m:
Done in 271c72b0.

PR #1161, opened 2022-06-30
Functions SDK v4

taeold:
### Breaking Changes

- Deprecated `allowInvalidAppCheckToken` option. Instead use`enforceAppCheck`.
- App Check enforcement on callable functions is disabled by default in v4.
- Requests containing invalid App Check tokens won't be denied unless you
- explicitly enable App Check enforcement using the new `enforceAppCheck` option.
- Furthermore, when enforcement is enabled, callable functions will deny
- all requests without App Check tokens.

- Dropped support for Node.js versions 8, 10, and 12.
- Dropped support for Admin SDK versions 8 and 9.
- Removed the `functions.handler` namespace.
- `DataSnapshot` passed to the Firebase Realtime Database trigger now
  matches the `DataSnapshot` returned by the Admin SDK, with null values
  removed.
- Removed `__trigger` object on function handlers.
- Reorganized source code location. This affects only apps that directly import files instead of using the recommend entry points specified in the
- Reworked the `apps` library and removed `lodash` as a runtime dependency.
- Unspecified function configuration value will be reset to platform default. Use `preserveExternalChanges` to prevent this behavior.

### Enhancements

- Logs created with the `functions.logger` package in v2 functions
  are now annotated with each request's trace ID, making it easy to correlate
  log entries with the incoming request. Trace IDs are especially useful for
  cases where 2nd gen's concurrency feature permits a function
  to handle multiple requests at any given time. See
  [Correlate log entries](https://cloud.google.com/logging/docs/view/correlate-logs) to learn more.
- `functions.logger.error` now always outputs an error object and is included in Google Cloud Error Reporting.
- The logging severity of Auth/App Check token validation has changed from `info` to `debug` level.
- Event parameters for 2nd generation functions are now strongly typed, permitting stronger TypeScript types for matched parameters.
- Add new params package to support parameterized environment configuration. See https://firebase.google.com/docs/functions/config-env for more information.
- Add new `functions.RESET_VALUE` and `functions.v2.options.RESET_VALUE` sentinel value for explicitly resetting function configuration to platform default.
- Add new `preserveExternalChanges` option to prevent Firebase CLI from resetting unspecified configuration option to platform default

PR #701, opened 2020-06-09
Makes "logger/compat" more focused on compatibility. Fixes #697

mbleigh:
Refactors the implementation of `logger/compat` to be a simple straightforward "frontport" of Node 8 logging to newer runtimes. This should make it easier for folks migrating to Node 10 to use as a drop-in fix for logging while they migrate to the `logger` package over the long-term.

Driven by discussion in #697

PR #1948, opened 2026-08-12
fix(logger): do not synthesize error stack on console.error in logger…

### LINKED ISSUE #1945: Firebase schedule function error
> I have the same issue. My schedule function logs the error after I use `firebase-functions/logger/compat`
> 
> Image
> 
> It's a compat package error or something? 

 _Originally posted by @samuel-pp in [#721](https://github.com/firebase/firebase-functions/issues/721#issuecomment-5140828672)_

shettyvarun268:
### Description

Fixes #1945

#### Background:
When users import `firebase-functions/logger/compat`, it monkeypatches global `console.*` methods to format logs into structured Cloud Logging JSON lines.

Previously, `patchedConsole("ERROR")` unconditionally created a synthetic `Error` stack for all `console.error` calls via `message = new Error(message).stack || message`.

This caused two problems:
1. **Benign Node.js runtime warnings escalated into false application crashes:** Node.js internal warning handlers (such as `MaxListenersExceededWarning` triggered when connection pooling with libraries like `axios`) print warnings via `console.error`. `logger/compat` intercepted these string warnings and attached a fake stack trace pointing to `patchedConsole` inside `compat.js`. Google Cloud Error Reporting then interpreted these harmless warnings as active application crashes in the function runtime.
2. **Error instance stack corruption:** When actual `Error` objects were passed to `console.error(err)`, `util.format(err)` already rendered the callsite stack trace. Wrapping it again in `new Error(message).stack` resulted in a distorted `Error: Error: ...` double-wrapped stack trace.

#### Changes:
1. **`src/logger/compat.ts`**: Removed `new Error(message).stack` wrapper from `patchedConsole`. String logs and Node runtime warnings to `console.error` are now output directly with `ERROR` severity without synthetic stack traces, while actual `Error` objects retain their authentic callsite stack.
2. **`spec/logger.spec.ts`**: Added a `describe("compat", ...)` test block covering `console.log`, `console.info`, `console.debug`, `console.warn`, and `console.error` (handling strings, formatted templates, and `Error` objects).

### Code sample

```typescript
require("firebase-functions/logger/compat");

// Output before:
// {"severity":"ERROR","message":"Error: (node:...) MaxListenersExceededWarning...\n    at patchedConsole (/compat.js:31:17)..."}

// Output after:
// {"severity":"ERROR","message":"(node:...) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 testEvent listeners added to [EventEmitter]..."}
```

### Release notes

relnote: Fixed an issue where `firebase-functions/logger/compat` added synthetic Error stack traces to string `console.error` calls and Node.js runtime warnings.

### DIFF
--- a/spec/logger.spec.ts
+++ b/spec/logger.spec.ts
@@ -263,14 +263,6 @@ describe("logger", () => {
       });
     });
 
-    it("should patch console.info with INFO severity", () => {
-      console.info("test info log");
-      expectStdout({
-        severity: "INFO",
-        message: "test info log",
-      });
-    });
-
     it("should patch console.debug with DEBUG severity", () => {
       console.debug("test debug log");
       expectStdout({
--- a/spec/logger.spec.ts
+++ b/spec/logger.spec.ts
@@ -218,4 +218,100 @@ describe("logger", () => {
       });
     });
   });
+
+  describe("compat", () => {
+    const originalConsole = {
+      debug: console.debug,
+      info: console.info,
+      log: console.log,
+      warn: console.warn,
+      error: console.error,
+    };
+
+    before(async () => {
+      // Patch global console methods
+      await import("../src/logger/compat");
+    });
+
+    beforeEach(() => {
+      lastOut = "";
+      lastErr = "";
+    });
+
+    after(() => {
+      // Restore original console methods so other tests remain unaffected
+      console.debug = originalConsole.debug;
+      console.info = originalConsole.info;
+      console.log = originalConsole.log;
+      console.warn = originalConsole.warn;
+      console.error = originalConsole.error;
+    });
+
+    it("should patch console.log with INFO severity", () => {
+      console.log("test info log");
+      expectStdout({
+        severity: "INFO",
+        message: "test info log",
+      });
+    });
+
+    it("should patch console.log with no arguments", () => {
+      console.log();
+      expectStdout({
+        severity: "INFO",
+        message: "",
+      });
+    });
+
+    it("should patch console.info with INFO severity", () => {
+      console.info("test info log");
+      expectStdout({
+        severity: "INFO",
+        message: "test info log",
+      });
+    });
+
+    it("should patch console.debug with DEBUG severity", () => {
+      console.debug("test debug log");
+      expectStdout({
+        severity: "DEBUG",
+        message: "test debug log",
+      });
+    });
+
+    it("should patch console.warn with WARNING severity", () => {
+      console.warn("test warning log");
+      expectStderr({
+        severity: "WARNING",
+        message: "test warning log",
+      });
+    });
+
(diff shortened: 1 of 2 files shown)


IzaakGough (approved):
lgtm!

spec/logger.spec.ts
```diff
@@ -218,4 +218,100 @@ describe("logger", () => {
...
+
+    it("should patch console.log with no arguments", () => {
+      console.log();
+      expectStdout({
+        severity: "INFO",
+        message: "",
+      });
+    });
+
+    it("should patch console.info with INFO severity", () => {
```
wandamora: Remove this duplicate test.

src/logger/compat.ts
```diff
@@ -24,12 +24,14 @@ import { format } from "util";
 import { CONSOLE_SEVERITY, UNPATCHED_CONSOLE } from "./common";
 
 /** @hidden */
-function patchedConsole(severity: string): (data: any, ...args: any[]) => void {
-  return function (data: any, ...args: any[]): void {
-    let message = format(data, ...args);
-    if (severity === "ERROR") {
```
ajperel, line 30: Are there any cases where we still want this behavior if the message was from an actual stack trace or something? I'm not sure if in addition to fixing a real problem we're regressing in some cases where we actually wanted this behavior. Someone wrote it this way for a reason the first time but I don't know why.  I'd be curious if you did any research into this and what you found.

wandamora (approved):
LGTM, just one nit.

src/logger/compat.ts
shettyvarun268: I looked into the git history to see why it was written this way. logger/compat was originally added in v3 (PR #701) purely as a drop-in shim to output structured JSON without any stack synthesis. In v4 (PR #1161), when the main SDK logger (functions.logger.error) was updated to synthesize stacks for Cloud Error Reporting, that line was copied over to compat.ts as well (and without the instanceof Error check).

While synthesizing a stack makes sense when someone explicitly calls functions.logger.error("..."), doing it in console.error causes Node runtime warnings (like MaxListenersExceededWarning from connection pools) to get decorated with a fake stack trace and misclassified as application crashes in Cloud Error Reporting. It also double-wrapped actual Error objects passed to console.error(err).

Taking this out doesn't regress intended behavior: console.error(err) with real Error objects still outputs the genuine stack trace through util.format, string logs will still be categorized as severity: "ERROR", and the main SDK logger (src/logger/index.ts) remains untouched. This just restores logger/compat to its original intended behavior.

spec/logger.spec.ts
shettyvarun268: Made the change. Thanks Wanda!

src/logger/compat.ts
ajperel: OK. Thanks for digging a bit more. I'm convinced.

ajperel (approved):

PR #1955, opened 2026-09-01
feat: unify global manifest on globalThis for forward-compatible manifest extraction

inlined:
### Description
Backs all manifest-level declarations (params, requiredRoles, lifecycleHooks, requiredAPIs, extensions, and any arbitrary future manifest fields) by a shared globalManifest object on globalThis (Symbol.for("firebase-functions:manifest")).

This ensures that:
1. Duplicate or nested node_modules installations of firebase-functions (e.g. within function kits or dependencies) write directly to the shared global manifest.
2. loadStack() in src/runtime/loader.ts is completely agnostic as to what features and options exist in globalManifest, simply spreading ...globalManifest alongside endpoints and specVersion.
3. setGlobalOptions and getGlobalOptions share state across module boundaries using Symbol.for("firebase-functions:v2:globalOptions").

### Scenarios Tested
- Ran unit tests across full test suite: npm test (all 976 tests passing).
- Added test in spec/runtime/loader.spec.ts verifying arbitrary future properties on globalManifest are preserved in ManifestStack and properly cleared.
- Ran linter: npm run lint:ts -- --quiet (0 errors).
- Built project: npm run build.

### Sample Commands
npm test

relnote: Unify global manifest on globalThis for forward-compatible manifest extraction across duplicate module installations

### DIFF
--- a/src/lifecycle/index.ts
+++ b/src/lifecycle/index.ts
@@ -43,14 +43,25 @@ export type LifecycleAction =
 
 import { globalManifest } from "../runtime/manifest";
 
-function getDeclaredHooks(): Record<string, LifecycleAction> {
+/**
+ * Gets or creates the mutable lifecycle hooks record directly on `globalManifest`.
+ * Used exclusively for write operations (e.g., registering a hook) so that
+ * `globalManifest.lifecycleHooks` is created lazily on demand.
+ */
+function getOrCreateDeclaredLifecycleHooks(): Record<string, LifecycleAction> {
   if (!globalManifest.lifecycleHooks || typeof globalManifest.lifecycleHooks !== "object") {
     globalManifest.lifecycleHooks = {};
   }
   return globalManifest.lifecycleHooks as Record<string, LifecycleAction>;
 }
 
-function getExistingDeclaredHooks(): Record<string, LifecycleAction> {
+/**
+ * Gets the existing lifecycle hooks record on `globalManifest` if defined,
+ * or returns a fallback empty object without mutating `globalManifest`.
+ * Used for read-only operations (e.g., property lookup, keys, descriptors)
+ * to avoid emitting an unwanted empty `lifecycleHooks: {}` into the manifest.
+ */
+function getExistingDeclaredLifecycleHooks(): Record<string, LifecycleAction> {
   if (!globalManifest.lifecycleHooks || typeof globalManifest.lifecycleHooks !== "object") {
     return {};
   }
@@ -73,32 +84,36 @@ export const declaredLifecycleHooks: Record<string, LifecycleAction> = new Proxy
   {
     get(_, prop) {
       if (typeof prop === "string") {
-        return getExistingDeclaredHooks()[prop];
+        return getExistingDeclaredLifecycleHooks()[prop];
       }
       return undefined;
     },
     set(_, prop, value) {
       if (typeof prop === "string") {
-        getDeclaredHooks()[prop] = value as LifecycleAction;
+        getOrCreateDeclaredLifecycleHooks()[prop] = value as LifecycleAction;
         return true;
       }
       return false;
     },
     deleteProperty(_, prop) {
-      if (typeof prop === "string") {
-        delete getExistingDeclaredHooks()[prop];
+      if (typeof prop === "string" && globalManifest.lifecycleHooks) {
+        const hooks = globalManifest.lifecycleHooks as Record<string, LifecycleAction>;
+        delete hooks[prop];
+        if (Object.keys(hooks).length === 0) {
+          delete globalManifest.lifecycleHooks;
+        }
         return true;
       }
       return false;
     },
     has(_, prop) {
-      return Reflect.has(getExistingDeclaredHooks(), prop);
+      return Reflect.has(getExistingDeclaredLifecycleHooks(), prop);
     },
     ownKeys() {
-      return Reflect.ownKeys(getExistingDeclaredHooks());
+      return Reflect.ownKeys(getExistingDeclaredLifecycleHooks());
     },
     getOwnPropertyDescriptor(_, prop) {
-      const hooks = getExistingDeclaredHooks();
+      const hooks = getExistingDeclaredLifecycleHooks();
       if (typeof prop === "string" && prop in hooks) {
         return {
           enumerable: true,
@@ -119,7 +134,7 @@ export const declaredLifecycleHooks: Record<string, LifecycleAction> = new Proxy
  * @param action The lifecycle action to execute.
  */
 export function afterFirstDeploy(action: LifecycleAction): void {
-  const hooks = getDeclaredHooks();
+  const hooks = getOrCreateDeclaredLifecycleHooks();
   if (hooks.afterFirstDeploy) {
     throw new Error("Only one afterFirstDeploy lifecycle hook is allowed per codebase.");
   }
@@ -133,7 +148,7 @@ export function afterFirstDeploy(action: LifecycleAction): void {
  * @param action The lifecycle action to execute.
  */
 export function afterRedeploy(action: LifecycleAction): void {
-  const hooks = getDeclaredHooks();
+  const hooks = getOrCreateDeclaredLifecycleHooks();
   if (hooks.afterRedeploy) {
     throw new Error("Only one afterRedeploy lifecycle hook is allowed per codebase.");
   }
--- a/src/lifecycle/index.ts
+++ b/src/lifecycle/index.ts
@@ -41,32 +41,16 @@ export type LifecycleAction =
   | { call: CallAction; task?: never; http?: never }
   | { http: HttpAction; task?: never; call?: never };
 
-/**
- * Use a global singleton to manage the list of declared lifecycle hooks.
- *
- * This ensures that lifecycle hooks are shared between CJS and ESM builds,
- * avoiding the "dual-package hazard" where the src/bin/firebase-functions.ts (CJS) sees
- * an empty list while the user's code (ESM) populates a different list.
- */
-const majorVersion =
-  // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
-  typeof __FIREBASE_FUNCTIONS_MAJOR_VERSION__ !== "undefined"
-    ? // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
-      __FIREBASE_FUNCTIONS_MAJOR_VERSION__
-    : "0";
+import { globalManifest } from "../runtime/manifest";
 
-const GLOBAL_SYMBOL = Symbol.for(`firebase-functions:lifecycle:declaredHooks:v${majorVersion}`);
-const globalSymbols = globalThis as unknown as Record<symbol, Record<string, LifecycleAction>>;
-
-if (!globalSymbols[GLOBAL_SYMBOL]) {
-  globalSymbols[GLOBAL_SYMBOL] = {};
+function getDeclaredHooks(): Record<string, LifecycleAction> {
+  if (!globalManifest.lifecycleHooks || typeof globalManifest.lifecycleHooks !== "object") {
+    globalManifest.lifecycleHooks = {};
+  }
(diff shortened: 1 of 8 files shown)


src/lifecycle/index.ts
```diff
@@ -41,32 +41,16 @@ export type LifecycleAction =
...
+  }
+  return globalManifest.lifecycleHooks as Record<string, LifecycleAction>;
 }
 
-/**
- * Singleton dictionary of declared lifecycle hooks.
- * @internal
- */
-export const declaredLifecycleHooks: Record<string, LifecycleAction> = globalSymbols[GLOBAL_SYMBOL];
+export const declaredLifecycleHooks: Record<string, LifecycleAction> = {};
```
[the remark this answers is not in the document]
inlined: Done. Backed `declaredLifecycleHooks` with a Proxy delegating to `globalManifest.lifecycleHooks`.

src/security/roles.ts
```diff
@@ -20,17 +20,53 @@
...
+  },
+  [Symbol.iterator](): IterableIterator<string> {
+    return getDeclaredRolesList()[Symbol.iterator]();
+  },
+  forEach(callback: (value: string, value2: string, set: typeof declaredRoles) => void): void {
+    for (const role of getDeclaredRolesList()) {
+      callback(role, role, declaredRoles);
+    }
+  },
+};
```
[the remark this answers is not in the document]
inlined: Done. Added `delete`, `keys`, `values`, and `entries` to ensure full `Set` interface compatibility.

src/params/index.ts
```diff
@@ -54,34 +54,18 @@ export type {
...
+import { globalManifest } from "../runtime/manifest";
+import type { WireParamSpec } from "./types";
 
-if (!globalSymbols[GLOBAL_SYMBOL]) {
-  globalSymbols[GLOBAL_SYMBOL] = [];
-}
+type SecretOrExpr = Param<any> | SecretParam | JsonSecretParam<any>;
 
-export const declaredParams: SecretOrExpr[] = globalSymbols[GLOBAL_SYMBOL];
+export const declaredParams: SecretOrExpr[] = [];
```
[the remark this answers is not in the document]
inlined: Done. Backed `declaredParams` with a shared symbol on `globalThis`.

src/security/roles.ts
```diff
@@ -20,17 +20,81 @@
...
+    return [];
+  }
+  return globalManifest.requiredRoles as string[];
+}
+
 /**
  * Global set of declared IAM roles required by this codebase.
- * Accumulated in-memory during manifest generation.
+ *
+ * NOTE: This cannot be a native `Set` because `globalManifest.requiredRoles`
```
ajperel, line 37: Thank you for this clear explanation of my first question when looking at it.

src/security/roles.ts
```diff
@@ -20,17 +20,81 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+import { globalManifest } from "../runtime/manifest";
+
+function getDeclaredRolesList(): string[] {
+  if (!Array.isArray(globalManifest.requiredRoles)) {
+    return [];
```
ajperel: Hmm. Technically if a bunch of people call this they'll all get different empty arrays... do we ever populate declared roles _after_ this might have been called?

I'm wondering why this doesn't just do `globalManifest.requiredRoles = [];` similar to `registerRole` below?

src/runtime/manifest.ts
```diff
@@ -166,11 +165,42 @@ export interface ManifestLifecycleAction {
...
+export const globalManifest: Record<string, unknown> = globalSymbols[GLOBAL_MANIFEST_SYMBOL];
+
+/**
+ * Clears all entries from the shared global manifest.
+ * @internal
+ */
+export function clearGlobalManifest(): void {
+  for (const key of Object.keys(globalManifest)) {
+    const val = globalManifest[key];
+    if (val instanceof Set) {
```
ajperel: Did you not get rid of all sets? I saw you got rid of one for roles at least. I wonder if this is dead code.

src/lifecycle/index.ts
```diff
@@ -41,32 +41,76 @@ export type LifecycleAction =
...
+function getDeclaredHooks(): Record<string, LifecycleAction> {
+  if (!globalManifest.lifecycleHooks || typeof globalManifest.lifecycleHooks !== "object") {
+    globalManifest.lifecycleHooks = {};
+  }
+  return globalManifest.lifecycleHooks as Record<string, LifecycleAction>;
+}
 
-if (!globalSymbols[GLOBAL_SYMBOL]) {
-  globalSymbols[GLOBAL_SYMBOL] = {};
+function getExistingDeclaredHooks(): Record<string, LifecycleAction> {
```
ajperel: I had to read like 5 times to see the difference between this and `getDeclaredHooks` can you add some comments to each to explain the difference and why we need them both to avoid emitting empty `{}` into the manifest?

I do worry someone in the future will hold this wrong.

src/lifecycle/index.ts
```diff
@@ -41,32 +41,76 @@ export type LifecycleAction =
...
+      return undefined;
+    },
+    set(_, prop, value) {
+      if (typeof prop === "string") {
+        getDeclaredHooks()[prop] = value as LifecycleAction;
+        return true;
+      }
+      return false;
+    },
+    deleteProperty(_, prop) {
```
ajperel, line 98: technically if we delete _all_ properties we still end up with an empty `{}` that we don't want right?

But in practice I can't imagine this happening outside of a testing edge case.

src/lifecycle/index.ts
```diff
@@ -41,32 +41,76 @@ export type LifecycleAction =
...
-  // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
-  typeof __FIREBASE_FUNCTIONS_MAJOR_VERSION__ !== "undefined"
-    ? // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
-      __FIREBASE_FUNCTIONS_MAJOR_VERSION__
-    : "0";
+import { globalManifest } from "../runtime/manifest";
 
-const GLOBAL_SYMBOL = Symbol.for(`firebase-functions:lifecycle:declaredHooks:v${majorVersion}`);
-const globalSymbols = globalThis as unknown as Record<symbol, Record<string, LifecycleAction>>;
+function getDeclaredHooks(): Record<string, LifecycleAction> {
```
ajperel: Slight bikeshed on if this and below should include `Lifecycle` in the name, e.g., `getDeclaredLifecycleHooks` ... but given the package maybe implied enough?

src/lifecycle/index.ts
```diff
@@ -41,32 +41,76 @@ export type LifecycleAction =
...
+ * NOTE: A Proxy is necessary here because `lifecycleHooks` is only initialized
+ * on `globalManifest` when hooks are registered (to avoid emitting empty `{}`
+ * into the manifest). The Proxy dynamically delegates all property reads, writes,
+ * and key enumerations directly to `globalManifest.lifecycleHooks` on `globalThis`,
+ * ensuring dual-package hazard protection across ESM/CJS or duplicate module contexts.
+ *
  * @internal
  */
-export const declaredLifecycleHooks: Record<string, LifecycleAction> = globalSymbols[GLOBAL_SYMBOL];
+export const declaredLifecycleHooks: Record<string, LifecycleAction> = new Proxy(
```
ajperel, line 82: Should we in a future PR eliminate the need for these proxies and just always use the getters/direct manifest object?

src/params/index.ts
```diff
@@ -54,34 +54,27 @@ export type {
...
 type SecretOrExpr = Param<any> | SecretParam | JsonSecretParam<any>;
 
-/**
- * Use a global singleton to manage the list of declared parameters.
- *
- * This ensures that parameters are shared between CJS and ESM builds,
- * avoiding the "dual-package hazard" where the src/bin/firebase-functions.ts (CJS) sees
- * an empty list while the user's code (ESM) populates a different list.
- */
-const majorVersion =
```
ajperel, line 71: Why did we have this major version bit before and why is it safe to get rid of it now?

src/runtime/loader.ts
```diff
@@ -195,25 +192,29 @@ export async function loadStack(functionsDir: string): Promise<ManifestStack> {
...
+  const existingExtensions =
+    globalManifest.extensions && typeof globalManifest.extensions === "object"
+      ? (globalManifest.extensions as Record<string, ManifestExtension>)
+      : {};
+  globalManifest.extensions = {
+    ...extensions,
+    ...existingExtensions,
+  };
+
+  if (!globalManifest.requiredAPIs) {
```
ajperel, line 212: Worth a comment why we are putting an empty object in for this one but carefully avoiding it for other cases.

ajperel (approved):
This hurts my brain and I only like 80% understand it. But Gemini thinks it's the bee's knees. I'm going to trust you and approve and we'll test it all with our kits.

src/security/roles.ts
inlined: Updated `getDeclaredRolesList()` to return a shared frozen empty array `EMPTY_ROLES` when `globalManifest.requiredRoles` is not set, ensuring consistent references without inadvertently creating empty `requiredRoles: []` on `globalManifest`.

src/runtime/manifest.ts
inlined: Done. Removed the dead `instanceof Set` check in `clearGlobalManifest()`.

src/lifecycle/index.ts
inlined: Done. Renamed and added detailed doc comments explaining that `getOrCreateDeclaredLifecycleHooks` initializes `globalManifest.lifecycleHooks = {}` for writes, while `getExistingDeclaredLifecycleHooks` returns a fallback `{}` for reads without mutating `globalManifest`.

src/lifecycle/index.ts
inlined: Done. `deleteProperty` now deletes `globalManifest.lifecycleHooks` if no properties remain.

src/lifecycle/index.ts
inlined: Done. Renamed to `getOrCreateDeclaredLifecycleHooks` and `getExistingDeclaredLifecycleHooks`.

src/lifecycle/index.ts
inlined: 100% agreed. In a follow-up PR we should update tests and callers to assert directly on `globalManifest` / `loadStack()` and remove the legacy `declaredLifecycleHooks` / `declaredRoles` wrappers completely.

src/params/index.ts
inlined: Previously, symbol names included the SDK major version (e.g. `firebase-functions:params:declaredParams:v${majorVersion}`) to isolate state across different SDK major versions. However, in modern setups (especially with function kits or nested dependencies where different packages may resolve slightly different versions of `firebase-functions`), this version partitioning caused cross-module communication failures where the loader in one package couldn't see params defined in another. Since the underlying wire protocol (`v1alpha1`) and manifest representation are unified across versions, using a single unversioned symbol on `globalThis` ensures seamless interop across module boundaries.

src/runtime/loader.ts
inlined: Done. Added a comment in `loader.ts` explaining that the `v1alpha1` spec expects `extensions: {}` and `requiredAPIs: []` to always be defined in `ManifestStack`, whereas optional features like `params`, `requiredRoles`, and `lifecycleHooks` are omitted when empty.

src/lifecycle/index.ts
ajperel: I'm trying to wrap my head around everything this does a bit still.

(1) Is this change backward compatible such that if a kit users an older firebase-functions and the wrapper has this one we're ok? I suspect not and we just need to be very clear that kits must use at least whatever version this releases in? But that normal deployments will be fine because there's only one version of firebase-functions.

(2) Cleaning this up in the future wouldn't cause new version compatibility problems right?

PR #1957, opened 2026-09-08
fix(params): restore backwards compatibility for declared params with pre-#1955 SDK versions

ajperel:
### Description
PR #1955 unified manifest extraction by storing all declared parameters on a shared `globalManifest` object (`Symbol.for("firebase-functions:manifest")`). While this provides a forward-compatible manifest for future releases, it broke compatibility when mixing newer SDK versions with older ones (v7.3.2 and earlier) that still rely on legacy versioned symbol arrays (`Symbol.for("firebase-functions:params:declaredParams:v${majorVersion}")`).

This change restores two-way backwards compatibility between new and older versions:

1. **New Nested Modules in Older Apps (e.g., nested v7.4+ dependency in a v7.3 root app)**:
   When `registerParam()` runs in the new SDK, it continues writing to `globalManifest.params` (per PR #1955) while also mirroring the parameters to the legacy versioned symbol array so older root loaders can discover them.

2. **Older Nested Modules in Newer Apps (e.g., nested v7.3 dependency in a v7.4+ root app)**:
   When `loader.ts` builds the deployment manifest, it reads from `globalManifest.params` and also inspects legacy versioned symbols to pull in any parameters declared by older nested packages, automatically deduplicating any shared entries.

### Scenarios Tested
- **Unit Tests**: Full test suite passing (`npm test`, 996 / 996 tests passing), including deduplication tests when parameters exist on both `globalManifest` and legacy symbols.
- **Production Deployments on GCP (`ajp-testing`)**:
  - Tested **v7.4+ Root with v7.3.2 Nested Dependency**: Successfully extracted manifest, deployed functions, and resolved runtime parameter values via Cloud Run.
  - Tested **v7.3.2 Root with v7.4+ Nested Dependency**: Successfully extracted manifest, deployed functions, and resolved runtime parameter values via Cloud Run.

### Release notes
relnote: Restore backwards compatibility for declared parameters when mixing SDK versions in nested dependencies

### DIFF
--- a/src/params/index.ts
+++ b/src/params/index.ts
@@ -66,9 +66,29 @@ import type { WireParamSpec } from "./types";
 type SecretOrExpr = Param<any> | SecretParam | JsonSecretParam<any>;
 
 const GLOBAL_PARAMS_SYMBOL = Symbol.for("firebase-functions:params:declaredParams");
+
+/**
+ * Use a global singleton to manage the list of declared parameters.
+ *
+ * This ensures that parameters are shared between CJS and ESM builds,
+ * avoiding the "dual-package hazard" where the src/bin/firebase-functions.ts (CJS) sees
+ * an empty list while the user's code (ESM) populates a different list.
+ */
+const majorVersion =
+  // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
+  typeof __FIREBASE_FUNCTIONS_MAJOR_VERSION__ !== "undefined"
+    ? // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
+      __FIREBASE_FUNCTIONS_MAJOR_VERSION__
+    : "0";
+
+const GLOBAL_SYMBOL = Symbol.for(`firebase-functions:params:declaredParams:v${majorVersion}`);
+
 const globalSymbols = globalThis as unknown as Record<symbol, SecretOrExpr[]>;
 if (!globalSymbols[GLOBAL_PARAMS_SYMBOL]) {
   globalSymbols[GLOBAL_PARAMS_SYMBOL] = [];
+}
+if (!globalSymbols[GLOBAL_SYMBOL]) {
+  globalSymbols[GLOBAL_SYMBOL] = [];
 }
 
 /**
@@ -89,6 +109,14 @@ function registerParam(param: SecretOrExpr) {
   }
   declaredParams.push(param);
 
+  const legacyList = globalSymbols[GLOBAL_SYMBOL];
+  for (let i = 0; i < legacyList.length; i++) {
+    if (legacyList[i].name === param.name) {
+      legacyList.splice(i, 1);
+    }
+  }
+  legacyList.push(param);
+
   if (!Array.isArray(globalManifest.params)) {
     globalManifest.params = [];
   }
@@ -107,6 +135,9 @@ function registerParam(param: SecretOrExpr) {
  */
 export function clearParams() {
   declaredParams.length = 0;
+  if (globalSymbols[GLOBAL_SYMBOL]) {
+    globalSymbols[GLOBAL_SYMBOL].length = 0;
+  }
   delete globalManifest.params;
 }
 
--- a/src/runtime/loader.ts
+++ b/src/runtime/loader.ts
@@ -216,21 +216,22 @@ export async function loadStack(functionsDir: string): Promise<ManifestStack> {
 
   // Ingest parameters declared on legacy versioned symbols (e.g. by older SDK modules in nested packages)
   // that were not already declared via the shared global manifest, preventing duplicates.
-  for (const sym of Object.getOwnPropertySymbols(globalThis)) {
-    const desc = sym.description || sym.toString();
-    if (desc.includes("firebase-functions:params:declaredParams:v")) {
-      const legacyParams = (globalThis as unknown as Record<symbol, unknown>)[sym];
-      if (Array.isArray(legacyParams) && legacyParams.length > 0) {
-        for (const p of legacyParams) {
-          if (typeof p?.toSpec === "function") {
-            const spec = p.toSpec();
-            if (!Array.isArray(globalManifest.params)) {
-              globalManifest.params = [];
-            }
-            const manifestParams = globalManifest.params as WireParamSpec<any>[];
-            if (!manifestParams.some((m) => m.name === spec.name)) {
-              manifestParams.push(spec);
-            }
+  const legacySymbols = Object.getOwnPropertySymbols(globalThis).filter((sym) =>
+    (sym.description || sym.toString()).includes("firebase-functions:params:declaredParams:v")
+  );
+
+  for (const sym of legacySymbols) {
+    const legacyParams = (globalThis as unknown as Record<symbol, unknown>)[sym];
+    if (Array.isArray(legacyParams)) {
+      for (const p of legacyParams) {
+        if (typeof p?.toSpec === "function") {
+          const spec = p.toSpec();
+          if (!Array.isArray(globalManifest.params)) {
+            globalManifest.params = [];
+          }
+          const manifestParams = globalManifest.params as WireParamSpec<any>[];
+          if (!manifestParams.some((m) => m.name === spec.name)) {
+            manifestParams.push(spec);
           }
         }
       }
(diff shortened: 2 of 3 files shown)


src/params/index.ts
```diff
@@ -89,6 +109,14 @@ function registerParam(param: SecretOrExpr) {
   }
   declaredParams.push(param);
 
+  const legacyList = globalSymbols[GLOBAL_SYMBOL];
+  for (let i = 0; i < legacyList.length; i++) {
+    if (legacyList[i].name === param.name) {
+      legacyList.splice(i, 1);
+    }
+  }
+  legacyList.push(param);
```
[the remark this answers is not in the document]
ajperel: This is exactly the implementation of the original code restored: https://github.com/firebase/firebase-functions/blob/0d78dc2c7e5be3393e4900f238f68fcad8573010/src/params/index.ts

Could it be improved yes, but for this PR I'm focused on doing exactly what we did before.

src/runtime/loader.ts
```diff
@@ -213,6 +214,29 @@ export async function loadStack(functionsDir: string): Promise<ManifestStack> {
...
+          if (typeof p?.toSpec === "function") {
+            const spec = p.toSpec();
+            if (!Array.isArray(globalManifest.params)) {
+              globalManifest.params = [];
+            }
+            const manifestParams = globalManifest.params as WireParamSpec<any>[];
+            if (!manifestParams.some((m) => m.name === spec.name)) {
+              manifestParams.push(spec);
+            }
+          }
```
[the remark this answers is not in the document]
ajperel: The old code we're trying to be backwards compatible with also always pushed the spec without checking it: https://github.com/firebase/firebase-functions/blob/0d78dc2c7e5be3393e4900f238f68fcad8573010/src/runtime/loader.ts#L208

src/params/index.ts
```diff
@@ -66,10 +66,30 @@ import type { WireParamSpec } from "./types";
...
+
+/**
+ * Use a global singleton to manage the list of declared parameters.
+ *
+ * This ensures that parameters are shared between CJS and ESM builds,
+ * avoiding the "dual-package hazard" where the src/bin/firebase-functions.ts (CJS) sees
+ * an empty list while the user's code (ESM) populates a different list.
+ */
+const majorVersion =
+  // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
```
taeold, line 78: i think this is broken now.

https://github.com/firebase/firebase-functions/blob/dd11b14ed3e7adacbbffc69a30131ac9ac69a450/package.json#L3

i'm not sure why package.json no longer sets the version to the latest released version.

src/params/index.ts
```diff
@@ -66,10 +66,30 @@ import type { WireParamSpec } from "./types";
...
+    ? // @ts-expect-error __FIREBASE_FUNCTIONS_MAJOR_VERSION__ is injected at build time
+      __FIREBASE_FUNCTIONS_MAJOR_VERSION__
+    : "0";
+
+const GLOBAL_SYMBOL = Symbol.for(`firebase-functions:params:declaredParams:v${majorVersion}`);
+
 const globalSymbols = globalThis as unknown as Record<symbol, SecretOrExpr[]>;
 if (!globalSymbols[GLOBAL_PARAMS_SYMBOL]) {
   globalSymbols[GLOBAL_PARAMS_SYMBOL] = [];
 }
```
taeold, line 89: why do we stil need this and also GLOBAL_SYMBOL?

ajperel: This is what gives us backwards compatibility

GLOBAL_PARAMS_SYMBOL is going to be "firebase-functions:params:declaredParams:v7" and is what was used prior to Thomas's PR.

GLOBAL_SYMBOL is functions:params:declaredParams and it is an array in the new global manifest that is already wire protocol ready.

By having both this version of the SDK can talk params with versions 7.3 and below (because of GLOBAL_PARAMS_SYMBOL) and also use the newer system for all versions later than this. In the 8.0 release we should remove GLOBAL_PARAMS_SYMBOL if we want for cleanup.

src/params/index.ts
ajperel: I think this is working. Thomas made this change a while ago so that the version "in development" on head is always 0.0.0-development but it is correctly populated as part of our release process. If you look at the package.json from NPM you'll see it is correctly set:

https://www.npmjs.com/package/firebase-functions?activeTab=code

taeold: Good to know. Then I will just make sure that this works during release process because otherwise majorVersion would always be `0`!

src/params/index.ts
taeold: When we refer to Thomas's PR, are we talking about https://github.com/firebase/firebase-functions/pull/1955?

If yes - we don't seem to have cut a release that contain that PR. Why do we need backward compatibility with an unreleased version?

ajperel: Sorry if it's confusing. I want to add to what Thomas implemented to #1955. He completely changed the global state to be forward compatible, but at the same time broke compatibility with old versions of firebase-functions. So _before_ we release I want to fix that. The next version of firebase-functions will be 7.4

Without this change if you have one firebase-functions on 7.4 and a nested dependency on 7.3 they will use different global objects to store params and things will be broken. After this change they would find _a_ channel where this would work. Just like it would work today between 7.3 and 7.1.

src/params/index.ts
```diff
@@ -89,6 +109,14 @@ function registerParam(param: SecretOrExpr) {
   }
   declaredParams.push(param);
 
+  const legacyList = globalSymbols[GLOBAL_SYMBOL];
```
taeold, line 112: is this really required?

If you are using new SDK versions, params will be correctly registered on the NEW param list, which will be preserved anyway.

src/runtime/loader.ts
```diff
@@ -213,6 +214,29 @@ export async function loadStack(functionsDir: string): Promise<ManifestStack> {
...
+    const desc = sym.description || sym.toString();
+    if (desc.includes("firebase-functions:params:declaredParams:v")) {
+      const legacyParams = (globalThis as unknown as Record<symbol, unknown>)[sym];
+      if (Array.isArray(legacyParams) && legacyParams.length > 0) {
+        for (const p of legacyParams) {
+          if (typeof p?.toSpec === "function") {
+            const spec = p.toSpec();
+            if (!Array.isArray(globalManifest.params)) {
+              globalManifest.params = [];
+            }
```
taeold: kinda weird place to do this initilzation. i might move this outside of the loop?

src/runtime/loader.ts
```diff
@@ -213,6 +214,29 @@ export async function loadStack(functionsDir: string): Promise<ManifestStack> {
     globalManifest.requiredAPIs = [];
   }
 
+  // Ingest parameters declared on legacy versioned symbols (e.g. by older SDK modules in nested packages)
+  // that were not already declared via the shared global manifest, preventing duplicates.
+  for (const sym of Object.getOwnPropertySymbols(globalThis)) {
```
taeold: kinda weird to put this in a loop. you could probabl findAll for `desc.includes("firebase-functions:params:declaredParams:v")`. It will remove 1 cyclomatic complexity something something:

```js
const legacyParamDeclarations = Object.getOwnPropertySymbols(globalThis).filter(...)
for (const legacyParams  of legacyParamDeclarations) {
...
}

```

taeold (approved):

src/params/index.ts
ajperel: It is so we can bidirectionally work. Otherwise 7.2 could talk to 7.4 but 7.4 couldn't talk to 7.2

src/runtime/loader.ts
ajperel: We can't because the array must not be present _unless_ there's a param to place in it according to existing tests, and old behavior. So we must only initialize it only as we need it. Though I agree it's not the best here, it needs to be here.

src/params/index.ts
taeold: That's good point - though I wonder if other direction really works at all (e.g. things like requiredAPIs won't really work?)

src/runtime/loader.ts
ajperel: Done.

src/params/index.ts
ajperel: yeah... those never worked across versions so I didn't both trying to fix them. The only other thing that used shared global state is lifecycle events and those are super new and not really announced (we added them for function kits) so I'm guessing they are not used enough to need backwards compatibility like this. If we get complaints I can easily add it with a similar PR.

PR #1965, opened 2026-09-15
Wire format for Optional secret params

Berlioz:


spec/params/params.spec.ts
```diff
@@ -16,24 +16,28 @@ describe("Params spec extraction", () => {
     expect(foo.toSpec().input).to.deep.equal({ text: { validationRegex: "\\d{5}" } });
   });
 
-  it("includes label and description for secret params", () => {
+  it("includes label, description, and optional for secret params", () => {
```
ajperel: Should we also have a test that shows we can leave optional off (like have it not specified at all) and everything is fine.

spec/runtime/loader.spec.ts
ajperel: can you explain why you made these non-secret realted changes?

PR #1966, opened 2026-09-15
Adds 'optional' config field to secret params

Berlioz:
relnote: none


src/params/types.ts
```diff
@@ -438,12 +440,13 @@ export type WireParamSpec<T extends string | number | boolean | string[]> = {
...
   format?: string;
+  optional?: boolean;
 };
 
-/** Configuration options which can be used to customize the prompting behavior of a parameter. */
+/** Configuration options which can be used to customize the prompting behavior of a non-secret parameter. */
 export type ParamOptions<T extends string | number | boolean | string[]> = Omit<
   ParamSpec<T>,
-  "name" | "type"
+  "name" | "type" | "optional"
```
ajperel, line 449: For my own learning what is this doing?

src/params/types.ts
```diff
@@ -452,6 +455,8 @@ export interface SecretParamOptions {
   label?: string;
   /** An optional long-form description of the parameter to be displayed while prompting. */
   description?: string;
+  /** If true, allows the user to decline to create a backing Cloud Secret Manager resource, resulting in undefined runtime value. */
+  optional?: boolean;
```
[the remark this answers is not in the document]
Berlioz: Okay, what I did instead is clunky, but also there appear to be literally zero users of non-required JSON secrets in actual 1p or 3p extensions right now so...eh....

spec/params/params.spec.ts
```diff
@@ -25,15 +25,29 @@ describe("Params spec extraction", () => {
...
     const jsonSecret = params.defineJsonSecret("TEST_JSON_SECRET", {
       label: "My JSON Secret",
       description: "A very JSON secret value",
+      optional: false,
     });
     const jsonSpec = jsonSecret.toSpec();
     expect(jsonSpec.name).to.equal("TEST_JSON_SECRET");
     expect(jsonSpec.label).to.equal("My JSON Secret");
     expect(jsonSpec.description).to.equal("A very JSON secret value");
+    expect(jsonSpec.optional).to.be.false;
```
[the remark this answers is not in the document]
Berlioz: Should we? fine, can't hurt

src/params/types.ts
```diff
@@ -606,6 +613,9 @@ export class JsonSecretParam<T = any> {
   runtimeValue(): T {
     const val = process.env[this.name];
     if (val === undefined) {
+      if (this.options.optional) {
+        return {} as T;
```
ajperel: Should we have a test for this return value?

ajperel (commented):
I think this looks good but will wait on API review

src/params/types.ts
Berlioz: ParamOptions<T> is the type being used for a `Param<string>, Param<number>` etc, so this is expressing that `optional` is not a valid option for those.

